PR #1, opened 2020-09-15
feat: add test for create schema

iand:
Adds makefile to support running postgres in docker and a test that checks the names
and column types of all expected tables.


iand:
Will reopen against master

PR #2, opened 2020-09-15
MVP sentinel-visor implementation

frrist:
In the interested of parallelizing work on this we are moving forward with the minimal implementation to be improved upon iteratively.

### What is this?
This is a rewrite of [lotus-chainwatch](https://github.com/filecoin-project/lotus/tree/master/cmd/lotus-chainwatch). It defines the following patterns/packages:

#### Model
Defines relations of blockchain types (actors, blocks, stateroots, etc.) using ORM provided by github.com/go-pg/pg/v10. Currently models Blocks, Drand entries, and the Miner Actor with subsequent models to follow.

#### Storage
A wrapper around a Postgres database providing instantiation, read, and write access.

#### Indexer
Service that listens on the lotus API ChainNotify channel and persists new blocks to Storage. To be run as a standalone singleton process.

#### Processor
Selects a batches of blocks from Storage for processing, computes a list of actors that changed for the batch, and dispatches tasks for processing the actors changes. Multiple processors can run simultaneously to horizontally scale chain state processing.

#### Scheduler
Consumes tasks from the processor and inserts them into their corresponding worker queue. The worker queue is implemented by github.com/gocraft/work backed by redis.

#### Tasks
Represents a piece of chain state to processes. Tasks are scheduled by the scheduler and upon successful completion persisted using the Publisher. Currently there is a single task implemented for processing miner actor changes with more to follow such that sentienl-visor matches processing done by lotus-chainwatch.  

#### Publisher
The Publisher consumes the result of a successful task and persists it to Storage. It serves as separation between processing tasks and persisting their results.

#### Separate ingest and query database instances.
The ingest instance will define a relatively flat scheme built to ingest data (task results) quickly. Keeping this flat will allow greater parallelization in the processing of tasks since there are little to no table constraints to satisfy.
The query instance will be built for query performance and will have data loaded into it from the ingest instance. It will serve as the backing for dashboards and developer/user queries.

#### More Tasks and Models!
Sentienl-Visor needs to match the data collected by lotus-chainwatch, this will require more tasks to be added for processing different actor and chain types as well as models for persistence.

#### Publisher Improvements
Task results are published synchronously with their processing. The Publisher needs to be modified such that task results are published asynchronously, writing to storage should not block a task from completing.

#### Testing
Neither lotus-chainwatch nor sentinel-visor have strong testing patterns defined. Each of the above packages will need their own set of tests implemented. Initially, I think the focus here should favor integration testing as I suspect this code will experience rapid change over the next few weeks.

#### Lotus API Performance Improvements
Crossing the lotus API is expensive to hit, likely (but not proven) due to the json marshaling. Improvements need to be made such that the API is hit as little as possible.

#### Misc. TODO's and Panics
Code has many TODO's and often panics instead of erroring. These all need to be addressed and errors handled more thoughtfully.


iand (approved):
LGTM with known issues

olizilla:
For the public record, and those less able to attend meetings, could we list the known issues here?

frrist:
@olizilla yes writing up a description now that will cover this.

PR #3, opened 2020-09-15
feat: add test for create schema

iand:
Adds makefile to support running postgres in docker and a test that checks the names
and column types of all expected tables.

### DIFF
--- a/storage/sql.go
+++ b/storage/sql.go
@@ -31,6 +31,8 @@ func NewDatabase(ctx context.Context, url string) (*Database, error) {
 	}
 
 	db := pg.Connect(opt)
+	db = db.WithContext(ctx)
+
 	// Check if connection credentials are valid and PostgreSQL is up and running.
 	if err := db.Ping(ctx); err != nil {
 		return nil, xerrors.Errorf("ping database: %w", err)
--- a/storage/sql_test.go
+++ b/storage/sql_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/go-pg/pg/v10"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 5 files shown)


storage/sql.go
```diff
@@ -31,6 +31,8 @@ func NewDatabase(ctx context.Context, url string) (*Database, error) {
 	}
 
 	db := pg.Connect(opt)
+	db = db.WithContext(ctx)
```
frrist, line 34: :+1:

storage/sql_test.go
```diff
@@ -0,0 +1,423 @@
...
+
+	err = resetSchema(db)
+	if !assert.NoError(t, err, "resetting schema") {
+		return
+	}
+
+	err = db.CreateSchema()
+	if !assert.NoError(t, err, "creating schema") {
+		return
+	}
```
frrist, line 397: Is there a reason to use this `if assert.NoError(t,err) { return }` pattern instead of `require.NoError(t,err)`?

storage/sql_test.go
```diff
@@ -0,0 +1,423 @@
...
+	return nil
+}
+
+type column struct {
+	name     string
+	datatype string
+	nullable bool
+}
+
+var expectedTables = []struct {
```
frrist, line 36: passing comment: as the number of tables grow I think this pattern may become tedious to maintain.

frrist (approved):
couple questions, LGTM

storage/sql_test.go
iand: I can see that, but I think it's valuable as a guard against regressions and as a prompt to write a migration.

storage/sql_test.go
iand: No, I just forgot all about the require package :smile: I'll change before I commit

PR #6, opened 2020-09-15
polish(publisher): receive publish operations on channel

frrist:
prevents tasks from blocking on publish

### DIFF
--- a/services/processor/publisher.go
+++ b/services/processor/publisher.go
@@ -9,20 +9,37 @@ import (
 	"github.com/filecoin-project/sentinel-visor/storage"
 )
 
-func NewPublisher(s *storage.Database) *Publisher {
+func NewPublisher(s *storage.Database, pubCh <-chan model.Persistable) *Publisher {
 	return &Publisher{
 		storage: s,
+		pubCh:   pubCh,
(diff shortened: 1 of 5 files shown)


placer14 (approved):
LGTM

services/processor/publisher.go
```diff
@@ -9,20 +9,37 @@ import (
...
+func (p *Publisher) Start(ctx context.Context) {
+	p.log.Info("starting publisher")
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				p.log.Info("stopping publisher")
+				return
+			case persistable := <-p.pubCh:
+				go func() {
```
iand, line 35: I have a slight worry about the unbounded concurrency of starting a goroutine for each persistence job, mainly due to the use of database connections from the pool. Something to look at as part of tuning later.

iand (approved):
LGTM just a couple of comments for future

services/processor/publisher.go
frrist: Agreed. I initially considered adding throttling based on a connection pool size, but doing so this early felt premature.

PR #7, opened 2020-09-16
fix(model): replace BeginContext with RunInTransaction

frrist:
Correctly handles commit and rollback cases for errors that we not handled before. Thank you for the prompt to use this pattern @iand!

### DIFF
--- a/services/indexer/types.go
+++ b/services/indexer/types.go
@@ -109,7 +109,7 @@ func (u *UnindexedBlockData) Persist(ctx context.Context, db *pg.DB) error {
 
 		if err := grp.Wait(); err != nil {
 			log.Info("Rollback unindexed block data", "error", err)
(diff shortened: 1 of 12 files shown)


services/indexer/types.go
```diff
@@ -68,61 +68,53 @@ func (u *UnindexedBlockData) Has(bh *types.BlockHeader) bool {
...
+		grp.Go(func() error {
+			if err := u.drandBlockEntries.PersistWithTx(ctx, tx); err != nil {
+				return err
+			}
+			return nil
+		})
+
+		if err := grp.Wait(); err != nil {
+			log.Info("Rollback unindexed block data", "error", err)
+			return tx.RollbackContext(ctx)
```
iand, line 112: I don't think we should rollback here. Return the error and let `db.RunInTransaction` roll it back

PR #8, opened 2020-09-16
feat(genesis): Add task for processing genesis state

frrist:
only handles miner state, as more models are added the genesis task
processing logic will need to be extended to include new models in the
genesis state.

special handling is required here since *most* task processing works by
comparing the state of actors across two tipsets, since there isn't a
tipset before the genesis block we need special handling.

to follow #7

### DIFF
--- a/services/processor/tasks/genesis/genesis.go
+++ b/services/processor/tasks/genesis/genesis.go
@@ -1,0 +1,209 @@
+package genesis
+
+import (
+	"bytes"
+	"context"
+	"github.com/gocraft/work"
+	"github.com/gomodule/redigo/redis"
+	"github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log/v2"
+
+	"github.com/filecoin-project/go-address"
+	lapi "github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/specs-actors/actors/builtin"
+	"github.com/filecoin-project/specs-actors/actors/builtin/miner"
+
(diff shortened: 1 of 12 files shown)


services/processor/tasks/genesis/genesis.go
```diff
@@ -0,0 +1,211 @@
...
+		case builtin.PaymentChannelActorCodeID:
+			// TODO
+		case builtin.MultisigActorCodeID:
+			// TODO
+		case builtin.RewardActorCodeID:
+			// TODO
+		case builtin.VerifiedRegistryActorCodeID:
+			// TODO
+		default:
+			return fmt.Errorf("unknown actor in genesis state. address: %s code: %s head: %s", addr, genesisAct.Code, genesisAct.Head)
```
iand, line 133: Seems drastic to abort just because we have an unknown actor. Maybe log and continue.

services/processor/tasks/genesis/genesis.go
```diff
@@ -0,0 +1,211 @@
...
+		for _, dealID := range sector.Info.DealIDs {
+			dealsModel = append(dealsModel, &minermodel.MinerDealSector{
+				MinerID:  addr.String(),
+				SectorID: uint64(sector.ID),
+				DealID:   uint64(dealID),
+			})
+		}
+	}
+	return &genesismodel.GenesisMinerTaskResult{
+		stateModel,
```
iand, line 206: Use field names here otherwise this will break if we add a field

frrist:
@iand I have addressed the feedback and rebased onto master, this is ready for another look when you are available.

iand (approved):
LGTM

PR #9, opened 2020-09-16
feat(task): add market actor task

frrist:
- process the market actors deal states and propols
- add handling for market actor processing in genesis task

to follow #8


iand (approved):
LGTM

PR #11, opened 2020-09-17
feat(task): add power actor processing task

frrist:
Need to wait for other inflight items to land before finishing this since we will want to pull out genesis state and that's done in #8


iand (approved):
LGTM

PR #13, opened 2020-09-17
feat(task): implement message processing task

frrist:
- process messages
- process message block inclusions
- process receipts


services/processor/scheduler.go
```diff
@@ -62,7 +70,12 @@ func (s *Scheduler) Stop() {
 }
 
 func (s *Scheduler) Dispatch(tips indexer.ActorTips) error {
-	for _, actors := range tips {
+	for ts, actors := range tips {
+		// TODO this is pretty ugly, need some local cache of tsKey to ts in here somewhere.
```
frrist, line 74: Want to call attention to this specifically as a better pattern is needed -- I think this is fine for now but should be addressed once we reach feature parity with chain-watch, after which we can implement some sort of tipset caching.

iand (approved):
LGTM

PR #15, opened 2020-09-18
feat: instrument with tracing

iand:
Adds tracing to primary indexer and processor functions, all lotus api calls and database persistence operations.


placer14 (approved):
One nit. Looks good otherwise!

docker-compose.yml
```diff
@@ -24,3 +30,31 @@ services:
...
+    container_name: jaeger
+    image: jaegertracing/all-in-one:1.19
+    ports:
+      - "6831:6831/udp"
+      - "5778:5778"
+      - "16686:16686"
+
+volumes:
+  timescaledb:
+  grafana-data:
```
placer14, line 59: Did you intend to include the grafana volume here?

lens/lotus/api.go
```diff
@@ -23,3 +30,87 @@ type APIWrapper struct {
 func (aw *APIWrapper) Store() adt.Store {
 	return aw.store
 }
+
```
frrist, line 33: A couple thoughts on the change: this is definitely an improvement, but I have some concerns --- some of which you already mentioned:
- Are we going to grow this as the api consumed by Visor grows? Or is there some alternative?
- Since we are wrapping the API would it be better to have it as a field instead of embedding?
- Would it be more appropriate to add API tracing to lotus instead? (as a follow on later)
- Should we include the arguments to these calls as tags on the Span?

README.md
```diff
@@ -6,6 +6,33 @@ A component of [**Sentinel**](https://github.com/filecoin-project/sentinel), a c
 A **Visor** process collects _permanent_ Filecoin chain meterics from a [**Lotus**](https://github.com/filecoin-project/lotus/) daemon, and writes them to a [**TimescaleDB**](https://github.com/timescale/timescaledb) time-series and relational datastore.
 
 
+## Getting Started
```
frrist, line 9: :heart: :pray:

services/indexer/indexer.go
```diff
@@ -8,7 +8,7 @@ import (
 	pg "github.com/go-pg/pg/v10"
 	cid "github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
-	trace "go.opentelemetry.io/otel/api/trace"
+	"github.com/opentracing/opentracing-go"
```
frrist, line 11: The ORM we are using supports tracing out-of-the-box: https://pg.uptrace.dev/tracing/. Are there any advantages to using opencensus over opentelemetry?

frrist (approved):
:+1: Good stuff. We are in urgent need of telemetry so I'd like to have this in asap since we can starting gathering metrics with it. 

I'd also like to have some more discussion around the comments I've left. If the result of that discussion results in things here needing to change I am happy to take that as a follow on.

lanzafame:
Could we make use of opencensus tracing instead of opentracing? We make use of opencensus for both metrics and tracing in lotus and it would be ideal use just the one tool.

iand:
> Could we make use of opencensus tracing instead of opentracing? We make use of opencensus for both metrics and tracing in lotus and it would be ideal use just the one tool.

Yep, no problem. I'll rework

iand:
OpenTracing and OpenCensus merged to start OpenTelemetry which is what the pg package supports, so I'll rework for that. There are software bridges to OpenCensus if needed.

lens/lotus/api.go
iand: We'll need to grow this if we want complete tracing coverage (we may not) or we should attempt to get it into lotus instead. I'm not convinced we want every method covered, especially since we are concerned about performance.

Wrapping the api, making it unexported would help enforce some type safety.

We could include the arguments as tags. I don't know useful it would be in general so I avoided paying the marshalling cost of things like cids in this pass.

docker-compose.yml
iand: Nope. Fixed

iand:
PTAL

Switched to using opentelemetry. It has a more awkward API (3 imports needed to add one tag to a span) but it operates very closely to the opentracing code so it was mostly mechanical changes.

Bonus screenshot of a trace showing go-pg traces as well as visor ones 

Screenshot from 2020-09-19 13-36-36

PR #18, opened 2020-09-21
Add standard build targets

iand:

PR #19, opened 2020-09-21
chore: add docker build and make targets

olizilla:
- add Dockerfile to build `filecoin/sentinel-visor` docker image
- ~~add make targets to build filecoin-ffi and sentinel-visor~~
- add .dockerignore to reduce context sent to docker build

Updated to use standard make targets from @iand in #18

WIP on https://github.com/filecoin-project/sentinel/issues/141

License: MIT


.dockerignore
```diff
@@ -0,0 +1,7 @@
+/.circleci
+/.github
+/extern/filecoin-ffi
+/visor
+/Dockerfile*
+/docker-compose*
+/sentinel-visor
```
iand, line 7: Line ending?

iand (approved):
LGTM with a minor nit

Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Builder
+FROM golang:1.14.9 as builder
```
iand, line 2: Why not 1.15?

olizilla: This was a guess based on the current go.mod file

https://github.com/filecoin-project/sentinel-visor/blob/47b18855ece919e6c8d337b2a7483764f928090c/go.mod#L3

iand: go.mod just defines the minimum version, so it's ok to use a later one

PR #22, opened 2020-09-21
chore: fix docker hub submodule error

olizilla:
Our autobuilds on dockerhub are failing when it tries to fetch the submodule.

The internet suggests the fix is to use dockerhub hooks to init the submodule post checkout.

see: https://stackoverflow.com/questions/58690455/how-to-correctly-initialize-git-submodules-in-dockerfile-for-docker-cloud

License: MIT


placer14 (approved):
Not thrilled about the non-hidden path, but LGTM.

olizilla:
Agreed. Dockerhub integration is weird. I can switch it to pushing images from CI.

PR #23, opened 2020-09-22
chore: tighten up error propagation

iand:
Move goroutine management out of processor start into the caller so we can return an
error instead of panicing. Also make NewCacheCtxStore return an error instead of
panicing.

Add informative context to errors throughout core logic of indexer, processor and storage.


frrist (approved):
Good stuff!

PR #26, opened 2020-09-22
chore: push docker tags from ci

olizilla:
Adds script and ci config to push to docker hub when.
- the master or main branch changes
- semver like tags are pushed to the repo

Pushing a semver like tag pushes a docker image with the same tag. The docker latest tag points to the last semver version tag

Pushing changes to main or master gets you a docker tag of the form <branch name>-<date>-<truncated commit sha>
e.g main-2020-f435621. a <branch>-latest docker tag is provided as a convenince if you really want the bleeding edge. The branch name is included in the docker tag name to make that clearer.

This is a tweak on what I set up for go-ipfs in https://github.com/ipfs/go-ipfs/pull/6949

This PR is motivated by the fact that dockerhub has repeatedly been a time sink, and it's latest issue is that our docker builds fail there when we try to init the git submodules, which is not a problem locally or on circleci. The suggested docker hub hooks fix in #22 did not solve it, so here we take back control and push the docker image we were building anyway to the hub.

License: MIT

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,11 @@ executors:
   golang:
     docker:
       - image: cimg/go:1.15.2
+  dockerizer:
+    docker:
+      - image: cimg/go:1.15.2
+    environment:
+      IMAGE_NAME: filecoin/sentinel-visor
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


placer14 (approved):
Looks wonderful. Ty!

PR #28, opened 2020-09-22
pg: configurable pool size

ognots:
provide a CLI flag and environment variable to set postgres pool size
set default as 75

### DIFF
--- a/run.go
+++ b/run.go
@@ -170,7 +170,7 @@ func setupStorageAndAPI(cctx *cli.Context) (context.Context, *RunContext, error)
 		return nil, nil, err
 	}
 
-	db, err := storage.NewDatabase(ctx, cctx.String("db"))
+	db, err := storage.NewDatabase(ctx, cctx.String("db"), cctx.Int("db-pool-size"))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


frrist (approved):

frrist:
CI failure is due to cargo not being install: `'[build_from_source] Error: cargo is not installed.`- No need to block this change due to it.

frrist (approved):

run.go
```diff
@@ -170,7 +170,7 @@ func setupStorageAndAPI(cctx *cli.Context) (context.Context, *RunContext, error)
 		return nil, nil, err
 	}
 
-	db, err := storage.NewDatabase(ctx, cctx.String("db"))
+	db, err := storage.NewDatabase(ctx, cctx.String("db"), cctx.Int("db-pool-size"))
```
placer14, line 173: Should we enforce a lower-bound?
```
poolSize := cctx.Int("db-pool-size")
if poolSize < 10 { poolSize = 10 }
```

placer14 (commented):
One thought, but GTM otherwise.

PR #30, opened 2020-09-22
feat: read blocks directly from chainstore

iand:
DO NOT MERGE

This is an untested experiment to build an API facade that bypasses the lotus API to read block data directly from disk


iand:
Superceded by #75

PR #31, opened 2020-09-22
fix(indexer): don't error on empty blocks_synced table

frrist:


services/indexer/indexer.go
```diff
@@ -195,9 +195,9 @@ func (i *Indexer) mostRecentlySyncedBlockHeight(ctx context.Context) (cid.Cid, i
 	task, err := i.storage.MostRecentSyncedBlock(ctx)
 	if err != nil {
 		if err == pg.ErrNoRows {
-			return i.genesis.Cids()[0], 0, xerrors.Errorf("query recent synced: %w", err)
+			return i.genesis.Cids()[0], 0, nil
 		}
-		return cid.Undef, 0, err
+		return cid.Undef, 0, xerrors.Errorf("query recent synced: %w", err)
```
placer14, line 200: I think the first return value is wrong.

placer14 (changes requested):
Check the first arg.

services/indexer/indexer.go
frrist: Can you explain more?

placer14: Ah, I misunderstood what was happening here in my haste to review. GTM

placer14 (approved):
False alarm. LGTM

PR #40, opened 2020-09-28
feat: add schema migration capability

iand:
Introduce schema versioning and migrations between versions. The database schema is versioned using
a simple monotonic counter.

Migrations are placed in storage/migrations and define the steps needed to migrate to and from
different database schema versions.

A new `migrate` subcommand is added. When run without arguments it checks the version of the database schema
in the connected database and verifies that it is compatible with the current version required by visor.

The `--latest` flag instructs the migrate command to migrate the database schema to the latest version.

The `--to` flag instructs the migrate command to migrate to a specific version, if possible.

By default the `index` and `process` subcommands verify the schema version but do not automatically migrate to
the latest. This behaviour can be enabled by supplying the `--allow-schema-migrations` flag.

Every running instance of visor now takes a shared advisory lock in the database to indicate their presence.
A schema migration requires an exclusive advisory lock which will fail if there are any existing locks already
taken. This ensures that migrations are only performed by a single instance of visor.


README.md
```diff
@@ -62,6 +62,37 @@ or by specifying the following flags:
...
+
+Visor will only migrate a schema if it determines that it has exclusive access to the database. 
+
+Visor can also be configured to automatically migrate the database when indexing or processing by passing the `--allow-schema-migration` flag.
+
+### Reverting a schema migration
+
+To revert to an earlier version, run:
+
+    sentinel-visor migrate --to <version>
```
frrist, line 92: nit: I think the binary name was changed back to just `visor`.

main.go
```diff
@@ -83,10 +84,16 @@ func main() {
 				EnvVars: []string{"JAEGER_SAMPLER_PARAM"},
 				Value:   0.0001,
 			},
+			&cli.BoolFlag{
+				Name:    "allow-schema-migration",
+				EnvVars: []string{"LOTUS_ALLOW_SCHEMA_MIGRATION"},
```
frrist, line 89: nit: `VISOR_ALLOW_SCHEMA_MIGRATION` seems more appropriate, but this isn't the first time we have been inconsistent with env var names

model/actors/common/actors.go
```diff
@@ -30,7 +30,7 @@ func (a *Actor) PersistWithTx(ctx context.Context, tx *pg.Tx) error {
 type ActorState struct {
 	Head  string `pg:",pk,notnull"`
 	Code  string `pg:",pk,notnull"`
-	State string `pg:",notnull"`
+	State string `pg:",type:jsonb,notnull"`
```
frrist, line 33: I couldn't find this in the go-pg docs, glad it exists!

storage/migrate.go
```diff
@@ -0,0 +1,155 @@
...
+
+	_ "github.com/filecoin-project/sentinel-visor/storage/migrations"
+	"github.com/go-pg/migrations/v8"
+	"github.com/go-pg/pg/v10"
+	"golang.org/x/xerrors"
+)
+
+// GetSchemaVersions returns the schema version in the database and the latest schema version defined by the available
+// migrations.
+func (d *Database) GetSchemaVersions(ctx context.Context) (int64, int64, error) {
```
frrist, line 15: nit: could use an `int` and avoid the casting below

storage/migrate.go
```diff
@@ -0,0 +1,155 @@
...
+		if m.Version > desiredVersion {
+			desiredVersion = m.Version
+		}
+	}
+
+	return dbVersion, desiredVersion, nil
+}
+
+// MigrateSchema migrates the database schema to the latest version based on the list of migrations available
+func (d *Database) MigrateSchema(ctx context.Context) error {
```
frrist, line 57: commentary: would be great if the below logic could be reused by `MigrateSchemaTo`

main.go
iand: Yeah, I keep swapping the lotus and visor names accidentally

storage/migrate.go
iand: It's not easy to because they have different logic between getting the schema versions and taking the lock

storage/migrate.go
iand: done

PR #41, opened 2020-09-28
feat: refactor task scheduler implementation

iand:
Fixes https://github.com/filecoin-project/sentinel/issues/153

This is a major refactoring of the scheduling system for Visor tasks. This is the additive change, a subsequent
commit will remove the old scheduler code including redis.

A new Scheduler type is added that runs a set of Tasks, restarting them when they fail. Concurrency levels can
be managed by adding more or less tasks of a specific type. Each task runs in a single goroutine, usually in
a loop to process work in batches. 

New tasks added (in tasks package):

 - ChainHeadIndexer - follows the chain head recording new tipsets
 - ChainHistoryIndexer - walks the chain history recording historic tipsets
 - ActorStateChangeProcessor - processes tipsets to extract actor state changes
 - ActorStateProcessor - processes actor state changes found by the ActorStateChangeProcessor
 - MessageProcessor - processes tipsets to extract messages

The postgres database is used as a global coordination point. ChainHeadIndexer and ChainHistoryIndexer write
newly discovered tipsets to the visor_processing_statechanges and visor_processing_messages tables. The
ActorStateChangeProcessor and MessageProcessor tasks lease tipsets from these tables in a batch to process.
The ActorStateChangeProcessor writes actor information to the visor_processing_actors table which 
the ActorStateProcessor task uses to determine which actors require processing.

Tasks may be marked as global singletons by using a GlobalLock when registering with the Scheduler. The
GlobalLock takes an exclusive advisory lock in postgres for the duration of the session which ensures that
only one task of that type may run anywhere. The ChainHeadIndexer and ChainHistoryIndexer tasks are expected
to be global singletons.

A new cli subcommand called run is added which instantiates the new scheduler with a sample number of each type
of task. This is suitable for testing but further configuration of task arrangement and concurrency needs to
be added in the future. The index and process subcommands will be changed to use the new scheduler in a later
change.

### DIFF
--- a/commands/run.go
+++ b/commands/run.go
@@ -86,7 +86,7 @@ var Run = &cli.Command{
 		// TODO: get these from CLI flags
 		defaultStateProcessors := 15
 
-		// By default we proces all supported actor types but we could limit here
+		// By default we process all supported actor types but we could limit here
 		defaultActorCodesToProcess := actorstate.SupportedActorCodes()
 		defaultActorCodesMaxHeight := int64(math.MaxInt64)
 
--- a/commands/run.go
+++ b/commands/run.go
@@ -40,6 +40,12 @@ var Run = &cli.Command{
 			Value:   true,
 			EnvVars: []string{"VISOR_INDEXHEAD"},
 		},
+		&cli.IntFlag{
+			Name:    "indexhead-confidence",
+			Usage:   "Sets the size of the cache used to hold tipsets for possible reversion before being committed to the database",
+			Value:   25,
+			EnvVars: []string{"VISOR_INDEXHEAD_CONFIDENCE"},
+		},
 		&cli.BoolFlag{
 			Name:    "indexhistory",
 			Value:   true,
@@ -161,7 +167,7 @@ var Run = &cli.Command{
 		if cctx.Bool("indexhead") {
 			scheduler.Add(schedule.TaskConfig{
 				Name:                "ChainHeadIndexer",
-				Task:                indexer.NewChainHeadIndexer(rctx.db, rctx.api),
+				Task:                indexer.NewChainHeadIndexer(rctx.db, rctx.api, cctx.Int("indexhead-confidence")),
 				Locker:              NewGlobalSingleton(ChainHeadIndexerLockID, rctx.db), // only want one forward indexer anywhere to be running
 				RestartOnFailure:    true,
 				RestartOnCompletion: true, // we always want the indexer to be running
--- a/commands/run.go
+++ b/commands/run.go
@@ -5,6 +5,8 @@ import (
 	"math"
 	"time"
 
+	"github.com/filecoin-project/specs-actors/actors/builtin"
+	"github.com/ipfs/go-cid"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/xerrors"
 
@@ -18,12 +20,118 @@ var Run = &cli.Command{
 	Name:  "run",
 	Usage: "Index and process blocks from the filecoin blockchain",
 	Flags: []cli.Flag{
+		&cli.Int64Flag{
+			Name:    "from",
+			Usage:   "Limit actor and message processing to tipsets at or above `HEIGHT`",
+			EnvVars: []string{"VISOR_HEIGHT_FROM"},
+		},
+		&cli.Int64Flag{
+			Name:        "to",
+			Usage:       "Limit actor and message processing to tipsets at or below `HEIGHT`",
+			Value:       math.MaxInt64,
+			DefaultText: "MaxInt64",
+			EnvVars:     []string{"VISOR_HEIGHT_TO"},
+		},
+		&cli.BoolFlag{
+			Name:    "indexhead",
+			Usage:   "Start indexing tipsets by following the chain head",
+			Value:   true,
+			EnvVars: []string{"VISOR_INDEXHEAD"},
+		},
+		&cli.BoolFlag{
+			Name:    "indexhistory",
+			Value:   true,
+			Usage:   "Start indexing tipsets by walking the chain history",
+			EnvVars: []string{"VISOR_INDEXHISTORY"},
+		},
+
+		&cli.DurationFlag{
+			Name:    "statechange-lease",
+			Aliases: []string{"scl"},
+			Value:   time.Minute * 15,
+			Usage:   "Lease time for the actor state change processor",
+			EnvVars: []string{"VISOR_STATECHANGE_LEASE"},
+		},
 		&cli.IntFlag{
-			Name:  "max-batch",
-			Value: 10,
+			Name:    "statechange-batch",
+			Aliases: []string{"scb"},
+			Value:   10,
+			Usage:   "Batch size for the actor state change processor",
+			EnvVars: []string{"VISOR_STATECHANGE_BATCH"},
+		},
+		&cli.IntFlag{
+			Name:    "statechange-workers",
+			Aliases: []string{"scw"},
+			Value:   5,
+			Usage:   "Number of actor state change processors to start",
+			EnvVars: []string{"VISOR_STATECHANGE_WORKERS"},
+		},
+
+		&cli.DurationFlag{
+			Name:    "actorstate-lease",
+			Aliases: []string{"asl"},
+			Value:   time.Minute * 15,
+			Usage:   "Lease time for the actor state processor",
+			EnvVars: []string{"VISOR_ACTORSTATE_LEASE"},
+		},
+		&cli.IntFlag{
+			Name:    "actorstate-batch",
+			Aliases: []string{"asb"},
+			Value:   10,
+			Usage:   "Batch size for the actor state processor",
+			EnvVars: []string{"VISOR_ACTORSTATE_BATCH"},
+		},
+		&cli.IntFlag{
+			Name:    "actorstate-workers",
+			Aliases: []string{"asw"},
+			Value:   5,
+			Usage:   "Number of actor state processors to start",
+			EnvVars: []string{"VISOR_ACTORSTATE_WORKERS"},
+		},
+		&cli.StringSliceFlag{
+			Name:        "actorstate-include",
+			Usage:       "List of actor codes that should be procesed by actor state processors",
+			DefaultText: "all supported",
+			EnvVars:     []string{"VISOR_ACTORSTATE_INCLUDE"},
+		},
+		&cli.StringSliceFlag{
+			Name:        "actorstate-exclude",
+			Usage:       "List of actor codes that should be not be procesed by actor state processors",
+			DefaultText: "none",
+			EnvVars:     []string{"VISOR_ACTORSTATE_EXCLUDE"},
+		},
+
+		&cli.DurationFlag{
+			Name:    "message-lease",
+			Aliases: []string{"ml"},
+			Value:   time.Minute * 15,
+			Usage:   "Lease time for the message processor",
+			EnvVars: []string{"VISOR_MESSAGE_LEASE"},
+		},
+		&cli.IntFlag{
+			Name:    "message-batch",
+			Aliases: []string{"mb"},
+			Value:   10,
+			Usage:   "Batch size for the message processor",
+			EnvVars: []string{"VISOR_MESSAGE_BATCH"},
+		},
+		&cli.IntFlag{
+			Name:    "message-workers",
+			Aliases: []string{"mw"},
+			Value:   5,
+			Usage:   "Number of message to start",
+			EnvVars: []string{"VISOR_MESSAGE_WORKERS"},
 		},
 	},
 	Action: func(cctx *cli.Context) error {
+		// Validate flags
+		heightFrom := cctx.Int64("from")
+		heightTo := cctx.Int64("to")
+		actorCodes, err := getActorCodes(cctx)
+		if err != nil {
+			return err
+		}
+
 		if err := setupLogging(cctx); err != nil {
 			return xerrors.Errorf("setup logging: %w", err)
 		}
@@ -48,51 +156,40 @@ var Run = &cli.Command{
 		scheduler := schedule.NewScheduler()
 
 		// Add one indexing task to follow the chain head
-		// TODO: enable/disable this with CLI flag
-		scheduler.Add(schedule.TaskConfig{
-			Name:                "ChainHeadIndexer",
-			Task:                indexer.NewChainHeadIndexer(rctx.db, rctx.api),
-			Locker:              NewGlobalSingleton(ChainHeadIndexerLockID, rctx.db), // only want one forward indexer anywhere to be running
-			RestartOnFailure:    true,
-			RestartOnCompletion: true, // we always want the indexer to be running
-		})
+		if cctx.Bool("indexhead") {
+			scheduler.Add(schedule.TaskConfig{
+				Name:                "ChainHeadIndexer",
+				Task:                indexer.NewChainHeadIndexer(rctx.db, rctx.api),
+				Locker:              NewGlobalSingleton(ChainHeadIndexerLockID, rctx.db), // only want one forward indexer anywhere to be running
+				RestartOnFailure:    true,
+				RestartOnCompletion: true, // we always want the indexer to be running
+			})
+		}
 
 		// Add one indexing task to walk the chain history
-		// TODO: enable/disable this with CLI flag
-		scheduler.Add(schedule.TaskConfig{
-			Name:                "ChainHistoryIndexer",
-			Task:                indexer.NewChainHistoryIndexer(rctx.db, rctx.api),
-			Locker:              NewGlobalSingleton(ChainHistoryIndexerLockID, rctx.db), // only want one history indexer anywhere to be running
-			RestartOnFailure:    true,
-			RestartOnCompletion: false, // run once only
-		})
-
-		// TODO: get these from CLI flags
-		defaultLeaseTime := time.Minute * 15
-		defaultBatchSize := 10
-		defaultStateChangeProcessors := 5
-		defaultStateChangeMaxHeight := int64(math.MaxInt64)
+		if cctx.Bool("indexhistory") {
+			scheduler.Add(schedule.TaskConfig{
+				Name:                "ChainHistoryIndexer",
+				Task:                indexer.NewChainHistoryIndexer(rctx.db, rctx.api),
+				Locker:              NewGlobalSingleton(ChainHistoryIndexerLockID, rctx.db), // only want one history indexer anywhere to be running
+				RestartOnFailure:    true,
+				RestartOnCompletion: false, // run once only
+			})
+		}
 
 		// Add several state change tasks to read which actors changed state in each indexed tipset
-		for i := 0; i < defaultStateChangeProcessors; i++ {
+		for i := 0; i < cctx.Int("statechange-workers"); i++ {
 			scheduler.Add(schedule.TaskConfig{
 				Name:                fmt.Sprintf("ActorStateChangeProcessor%03d", i),
-				Task:                actorstate.NewActorStateChangeProcessor(rctx.db, rctx.api, defaultLeaseTime, defaultBatchSize, defaultStateChangeMaxHeight),
+				Task:                actorstate.NewActorStateChangeProcessor(rctx.db, rctx.api, cctx.Duration("statechange-lease"), cctx.Int("statechange-batch"), heightFrom, heightTo),
 				RestartOnFailure:    true,
 				RestartOnCompletion: true,
 			})
 		}
 
-		// TODO: get these from CLI flags
-		defaultStateProcessors := 15
-
-		// By default we process all supported actor types but we could limit here
-		defaultActorCodesToProcess := actorstate.SupportedActorCodes()
-		defaultActorCodesMaxHeight := int64(math.MaxInt64)
-
 		// Add several state tasks to read actor state from each indexed block
-		for i := 0; i < defaultStateProcessors; i++ {
-			p, err := actorstate.NewActorStateProcessor(rctx.db, rctx.api, defaultLeaseTime, defaultBatchSize, defaultActorCodesMaxHeight, defaultActorCodesToProcess)
+		for i := 0; i < cctx.Int("actorstate-workers"); i++ {
+			p, err := actorstate.NewActorStateProcessor(rctx.db, rctx.api, cctx.Duration("actorstate-lease"), cctx.Int("actorstate-batch"), heightFrom, heightTo, actorCodes)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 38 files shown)


storage/sql.go
```diff
@@ -289,3 +296,165 @@ func verifyModel(ctx context.Context, db *pg.DB, m *orm.Table) error {
...
+    UPDATE visor_processing_blocks
+    SET claimed_until = ?
+    FROM (
+	    SELECT *
+	    FROM visor_processing_blocks
+	    WHERE completed_at IS null AND
+	          (claimed_until IS null OR claimed_until < ?) AND
+	          height > 0
+	    ORDER BY height DESC
+	    LIMIT ?
```
placer14, line 315: FYI the prior selection algorithm used `RANK()` to ensure we were making batch selections which fall on tipset boundaries (such that some blocks from a tipset are not divided across two processing batches). This would cause problems with the state resolution if we separated blocks.

iand: Noted. I'll adjust.

iand: I think this method should return a list of tipsets

iand: Reworked actor stage change detection and message processing to use tipsets instead of blocks. Renamed the `visor_processing_blocks` table to `visor_processing_statechanges` since it's a record of which tipsets have been analysed for actor state changes. It now has a tipset column instead of cid.

iand:
I believe this is feature complete. I want to continue testing against a synced node to uncover problems before merging. Testing will take a couple of days I think.

This PR doesn't break the existing indexer/processor so they can exist side by side for a while. I'll remove them in a later PR.

tasks/indexer/chainhistoryindexer.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/filecoin-project/sentinel-visor/storage"
+)
+
+func NewChainHistoryIndexer(d *storage.Database, node lens.API) *ChainHistoryIndexer {
+	return &ChainHistoryIndexer{
+		node:    node,
+		storage: d,
+
+		// TODO base finality value on the spec: https://github.com/filecoin-project/specs-actors/pull/702
+		finality: 1400,
```
placer14, line 24: Latest specs-actors has finality at `900` and doesn't appear to be changing soon.

placer14 (changes requested):
A bunch of questions and thoughts. Overall very happy with the interfaces and abstractions you arrived at. Let's consider including the simple fixes and separate the larger work for another PR.

storage/sql.go
```diff
@@ -289,3 +319,168 @@ func verifyModel(ctx context.Context, db *pg.DB, m *orm.Table) error {
...
+		_, err := tx.QueryContext(ctx, &blocks, `
+WITH leased AS (
+    UPDATE visor_processing_statechanges
+    SET claimed_until = ?
+    FROM (
+	    SELECT *
+	    FROM visor_processing_statechanges
+	    WHERE completed_at IS null AND
+	          (claimed_until IS null OR claimed_until < ?) AND
+	          height >= 0 AND height <= ?
```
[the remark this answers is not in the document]
placer14: Noticed this was included. Ignore.

storage/sql_test.go
```diff
@@ -41,3 +44,473 @@ func TestSchemaIsCurrent(t *testing.T) {
...
+	assert.Equal(t, "cid5", claimed[0].TipSet, "first claimed tipset")
+	assert.Equal(t, "cid3a", claimed[1].TipSet, "second claimed tipset")
+	assert.Equal(t, "cid2a,cid2b,cid2c", claimed[2].TipSet, "third claimed tipset")
+
+	// Check the database contains the leases
+	var count int
+	_, err = db.QueryOne(pg.Scan(&count), `SELECT COUNT(*) FROM visor_processing_statechanges WHERE claimed_until=?`, claimUntil)
+	require.NoError(t, err)
+	assert.Equal(t, batchSize, count)
+}
```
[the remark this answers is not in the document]
placer14: We discussed this in chat. This will create extra work and errors for visor. If we can be sure that errors don't impact other processing results from persisting to the DB, then it's probably okay.

tasks/actorstate/actorstatechange.go
```diff
@@ -0,0 +1,203 @@
...
+	if item.Height > 0 {
+		if err := p.processTipSet(ctx, ts); err != nil {
+			return xerrors.Errorf("process tipset: %w", err)
+		}
+	} else {
+		gp := NewGenesisProcessor(p.storage, p.node)
+		if err := gp.ProcessGenesis(ctx, ts); err != nil {
+			return xerrors.Errorf("process genesis: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
placer14: Re: https://github.com/filecoin-project/sentinel-visor/pull/41#discussion_r498377223

tasks/message/message.go
```diff
@@ -0,0 +1,246 @@
...
+	"github.com/filecoin-project/sentinel-visor/storage"
+	"github.com/filecoin-project/sentinel-visor/wait"
+)
+
+const idleSleepInterval = 60 * time.Second   // time to wait if the processor runs out of blocks to process
+const batchInterval = 100 * time.Millisecond // time to wait between batches
+
+var log = logging.Logger("message")
+
+var timeNow = time.Now
```
lanzafame, line 25: Why?

model/visor/processing.go
```diff
@@ -0,0 +1,201 @@
...
+	AddedAt time.Time `pg:",notnull"`
+
+	// ClaimedUntil marks the block as claimed for processing until the set time
+	ClaimedUntil time.Time
+
+	// CompletedAt is the time the block was read from the chain and analysed for actor state changes
+	CompletedAt time.Time
+
+	// ErrorsDetected contains any error encountered when analysing the block for actor state changes
+	ErrorsDetected string
```
lanzafame, line 44: Can this be made nullable? It makes checking for whether an error has occurred simpler

schedule/scheduler.go
```diff
@@ -0,0 +1,152 @@
...
+						break
+					}
+					log.Errorw("task exited with failure, restarting", "task", tc.Name, "error", err.Error())
+				} else {
+					if !tc.RestartOnCompletion {
+						// Exit the task
+						log.Infow("task exited cleanly, not restarting", "task", tc.Name)
+						break
+					}
+					log.Infow("task exited cleanly, restarting", "task", tc.Name)
```
lanzafame, line 122: This line gets triggered when ^Cing the run process:
```
^C2020-10-02T17:16:19.409+1000	INFO	indexer	indexer/chainheadindexer.go:44	stopping Indexer
2020-10-02T17:16:19.409+1000	INFO	schedule	schedule/scheduler.go:122	task exited cleanly, restarting	{"task": "ChainHeadIndexer"}
2020-10-02T17:16:19.409+1000	FATAL	visor	sentinel-visor/main.go:102	context canceled
```

model/visor/processing.go
lanzafame: hmm it seems it is but we are setting it to an empty string at some point

tasks/message/message.go
```diff
@@ -0,0 +1,246 @@
...
+	leaseLength time.Duration // length of time to lease work for
+	batchSize   int           // number of tipsets to lease in a batch
+	maxHeight   int64         // limit processing to tipsets equal to or below this height
+}
+
+// Run starts processing batches of tipsets and blocks until the context is done or
+// an error occurs.
+func (p *MessageProcessor) Run(ctx context.Context) error {
+	// Loop until context is done or processing encounters a fatal error
+	return wait.RepeatUntil(ctx, batchInterval, p.processBatch)
```
lanzafame, line 51: The for loop inside `wait.RepeatUntil` interfers with the schedulers `Run` loop. With `wait.RepeatUntil` the 4 message processors will only run once and then they block on something. Still digging...

lanzafame: Removing `wait.RepeatUntil` allows for the processors to keep restarting as intended but without the sleep interval between each run of the task.

commands/setup.go
```diff
@@ -154,3 +154,29 @@ func setupLogging(cctx *cli.Context) error {
 
 	return nil
 }
+
+const (
+	ChainHeadIndexerLockID    = 9898981111
+	ChainHistoryIndexerLockID = 9898981112
```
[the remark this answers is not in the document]
iand: Yes

model/visor/processing.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	return nil
+}
+
+func TipSetKeyFromString(s string) (types.TipSetKey, error) {
+	if len(s) < 2 {
+		return types.EmptyTSK, xerrors.Errorf("invalid tipset")
+	}
+
+	s = s[1 : len(s)-1]
```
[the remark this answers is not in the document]
iand: Because we're using `.String()` to serialize into the database, which wraps with `{}`:
```
// String() returns a human-readable representation of the key.
func (k TipSetKey) String() string {
	b := strings.Builder{}
	b.WriteString("{")
	cids := k.Cids()
	for i, c := range cids {
		b.WriteString(c.String())
		if i < len(cids)-1 {
			b.WriteString(",")
		}
	}
	b.WriteString("}")
	return b.String()
}
```

I considered using JSON but it's possible that the Go JSON encoder can change over time. Raw bytes are also a possibility but less friendly to work with when manually querying the database.

storage/sql_test.go
```diff
@@ -41,3 +44,473 @@ func TestSchemaIsCurrent(t *testing.T) {
 		})
 	}
 }
+
+func TestLeaseStateChanges(t *testing.T) {
+	if testing.Short() || !testutil.DatabaseAvailable() {
+		t.Skip("short testing requested or VISOR_TEST_DB not set")
+	}
```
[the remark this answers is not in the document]
iand: Yeah, there's quite a lot of boilerplate in the tests still as I'm still looking for patterns. Will clean that up in a separate pr.

tasks/actorstate/actor.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/filecoin-project/sentinel-visor/lens"
+	"github.com/filecoin-project/sentinel-visor/model"
+	commonmodel "github.com/filecoin-project/sentinel-visor/model/actors/common"
+)
+
+// was services/processor/tasks/common/actor.go
+
+// ActorExtracter extracts common actor state
+type ActorExtracter struct{}
```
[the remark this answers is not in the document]
iand: Yeah. lol. Will change

tasks/actorstate/actorstate.go
```diff
@@ -0,0 +1,249 @@
...
+		case <-ctx.Done():
+			return false, nil // Don't propagate cancelation error so we can resume processing cleanly
+		default:
+		}
+
+		info, err := NewActorInfo(actor)
+		if err != nil {
+			log.Errorw("unmarshal actor", "error", err.Error())
+			if err := p.storage.MarkActorComplete(ctx, actor.Head, actor.Code, timeNow(), err.Error()); err != nil {
+				log.Errorw("failed to mark actor complete", "error", err.Error())
```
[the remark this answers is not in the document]
iand: Yes, I'll do another pass through to add more context to errors.

tasks/actorstate/actorstate.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		return xerrors.Errorf("extract actor state: %w", err)
+	}
+	if err := data.Persist(ctx, p.storage.DB); err != nil {
+		// TODO handle this case with a retry
+		return xerrors.Errorf("persisting raw state: %w", err)
+	}
+
+	// Find a specific extracter for the actor type
+	extracter, exists := p.extracters[info.Actor.Code]
```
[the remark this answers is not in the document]
iand: Yes, we copy the extractors into each processor so we can avoid taking a shared lock

tasks/actorstate/actorstatechange.go
```diff
@@ -0,0 +1,203 @@
...
+		select {
+		case <-ctx.Done():
+			return false, nil // Don't propagate cancelation error so we can resume processing cleanly
+		default:
+		}
+
+		if err := p.processItem(ctx, item); err != nil {
+			log.Errorw("failed to process tipset", "error", err.Error(), "height", item.Height)
+			if err := p.storage.MarkStateChangeComplete(ctx, item.TipSet, item.Height, timeNow(), err.Error()); err != nil {
+				log.Errorw("failed to mark tipset complete", "error", err.Error(), "height", item.Height)
```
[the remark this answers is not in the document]
iand: OK, will do

tasks/actorstate/actorstatechange.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+
+		_, err = p.node.StateGetActor(ctx, addr, pts.Key())
+		if err != nil {
+			if strings.Contains(err.Error(), "actor not found") {
+				ll.Debugw("actor not found", "addr", str)
+				// TODO consider tracking deleted actors
+				continue
+			}
+			return xerrors.Errorf("get actor: %w", err)
```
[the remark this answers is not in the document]
iand: TODO was copied over from previous version, but it's a good point and probably we could allow the error through to be persisted.

tasks/actorstate/genesis.go
```diff
@@ -0,0 +1,233 @@
...
+
+func (p *GenesisProcessor) storageMinerState(ctx context.Context, gen *types.TipSet, addr address.Address, act *types.Actor) (*genesismodel.GenesisMinerTaskResult, error) {
+	// actual miner actor state and miner info
+	mstate, err := miner.Load(p.node.Store(), act)
+	if err != nil {
+		return nil, err
+	}
+
+	// miner raw and qual power
+	// TODO this needs caching so we don't re-fetch the power actors claim table for every tipset.
```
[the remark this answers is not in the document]
iand: Agree

tasks/indexer/chainhistoryindexer.go
iand: Adjusted

schedule/scheduler.go
```diff
@@ -0,0 +1,152 @@
...
+						log.Infow("task not started: lock not acquired", "task", tc.Name)
+						return
+					}
+					log.Errorw("task not started: lock not acquired", "task", tc.Name, "error", err.Error())
+					return
+				}
+				defer func() {
+					if err := tc.Locker.Unlock(ctx); err != nil {
+						if !errors.Is(err, context.Canceled) {
+							log.Errorw("failed to unlock task", "task", tc.Name, "error", err.Error())
```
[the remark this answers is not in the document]
iand: No they are session based and postgres removes them when we exit. They are editable in `pg_locks` if necessary.

tasks/message/message.go
iand: To assist testing so we can use a known time in comparisons. Yes, I know there are no tests in that package... yet :smile:

lanzafame: I recommend looking at https://github.com/raulk/clock which is what lotus uses for this

model/visor/processing.go
iand: Done, will write null when marking complete with no error message

tasks/message/message.go
iand: This is the deadlock I mentioned. Should be fixed in current branch

iand:
Thanks for the detailed reviews @placer14 and @lanzafame,  a nearly 4k loc change takes a lot of time and energy to read through. Much appreciated.

tasks/message/message.go
lanzafame: it is 👍

tasks/message/message.go
iand: Will do although we don't need anything sophisticated yet. Mainly just want to be able to control the time written to completed_at columns.

tasks/actorstate/actor.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (ActorExtracter) Extract(ctx context.Context, a ActorInfo, node lens.API) (model.Persistable, error) {
```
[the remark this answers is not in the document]
iand: Renamed

tasks/actorstate/actorstate.go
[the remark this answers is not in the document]
iand: Done

iand:
@placer14 I added some cli flags and env variables to the run subcommand. Interested in your feedback on the ux.

schedule/scheduler.go
iand: I think in this case the task had just finished and was restarting just as you killed the process. I've tidied up the logging in the area though.

tasks/actorstate/actorstate.go
```diff
@@ -0,0 +1,249 @@
...
+type ActorInfo struct {
+	Actor           types.Actor
+	Address         address.Address
+	ParentStateRoot cid.Cid
+	TipSet          types.TipSetKey
+	ParentTipSet    types.TipSetKey
+}
+
+// An ActorStateExtracter extracts actor state into a persistable format
+type ActorStateExtracter interface {
```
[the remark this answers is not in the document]
iand: Done

tasks/actorstate/actorstatechange.go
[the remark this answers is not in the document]
iand: The problem is that this is in a loop over the list of state changes for the tipset. We continue here so a missing actor does not abort the collection of other actor state changes. We don't have an opportunity to write the error to the database for this tipset.

storage/migrations/5_scheduler.go
```diff
@@ -0,0 +1,57 @@
...
+
+	up := batch(`
+CREATE TABLE IF NOT EXISTS "visor_processing_statechanges" (
+	"tip_set" text NOT NULL,
+	"height" bigint,
+	"added_at" timestamptz NOT NULL,
+	"claimed_until" timestamptz,
+	"completed_at" timestamptz,
+	"errors_detected" text,
+	PRIMARY KEY ("tip_set")
```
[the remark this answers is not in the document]
iand: In the ChainHeadIndexer we have a switch on the type of head change. There is currently a TODO in the HCRevert case. Perhaps that is where we would remove a tipset. 

I wonder if we should be maintaining a buffer of head changes so we can resolve apply/revert to recent changes and only each tipset after a certain time window.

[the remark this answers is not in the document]
iand: In fact I see this is exactly how it's done in the lotus codebase (chain/events). Pity the tipSetCache is not exported

iand:
@placer14 PTAL particularly at the new CLI flags for the run subcommand and the tipset cache I implemented for the head indexer

tasks/message/message.go
lanzafame: clock provides what you are looking for and I would prefer if we don't duplicate pkgs

tasks/indexer/chainheadindexer.go
```diff
@@ -0,0 +1,107 @@
...
+		case store.HCCurrent:
+			fallthrough
+		case store.HCApply:
+			log.Debugw("add tipset", "event", ch.Type, "height", ch.Val.Height(), "tipset", ch.Val.Key().String())
+			if int64(ch.Val.Height()) > height {
+				height = int64(ch.Val.Height())
+			}
+			tail, err := c.cache.Add(ch.Val)
+			if err != nil {
+				log.Errorw("tipset cache", "error", err.Error())
```
lanzafame, line 81: So this keeps logging errors when the new value height is the same as the current head height of the cache. I modified the `Add` function to return the current head with this error so I could see what the diff between the heights is and this is the logs that kept repeating:
```
2020-10-03T12:25:30.672+1000	ERROR	indexer	indexer/chainheadindexer.go:93	tipset cache	{"error": "reverted tipset does not match current head"}
2020-10-03T12:25:30.672+1000	ERROR	indexer	indexer/chainheadindexer.go:81	tipset cache	{"error": "added tipset height lower than current head", "oldheight": "112851", "newheight": "112851"}
2020-10-03T12:25:31.126+1000	ERROR	indexer	indexer/chainheadindexer.go:93	tipset cache	{"error": "reverted tipset does not match current head"}
2020-10-03T12:25:31.126+1000	ERROR	indexer	indexer/chainheadindexer.go:81	tipset cache	{"error": "added tipset height lower than current head", "oldheight": "112851", "newheight": "112851"}
2020-10-03T12:25:31.128+1000	ERROR	indexer	indexer/chainheadindexer.go:93	tipset cache	{"error": "reverted tipset does not match current head"}
2020-10-03T12:25:31.128+1000	ERROR	indexer	indexer/chainheadindexer.go:81	tipset cache	{"error": "added tipset height lower than current head", "oldheight": "112851", "newheight": "112851"}
2020-10-03T12:25:31.130+1000	ERROR	indexer	indexer/chainheadindexer.go:93	tipset cache	{"error": "reverted tipset does not match current head"}
2020-10-03T12:25:31.130+1000	ERROR	indexer	indexer/chainheadindexer.go:81	tipset cache	{"error": "added tipset height lower than current head", "oldheight": "112851", "newheight": "112851"}
2020-10-03T12:25:37.094+1000	ERROR	indexer	indexer/chainheadindexer.go:93	tipset cache	{"error": "reverted tipset does not match current head"}
2020-10-03T12:25:37.094+1000	ERROR	indexer	indexer/chainheadindexer.go:81	tipset cache	{"error": "added tipset height lower than current head", "oldheight": "112851", "newheight": "112851"}
```

If we attempted to add an actual old height, I would want to know about it but at the same height is going to happen.

iand: Possibly because of the fallthrough for HCCurrent events. I don't have a fully synced node to try this against yet.

iand: Added explicit handling for current head

schedule/scheduler.go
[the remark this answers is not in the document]
placer14: As opposed to mucking around inside DB? https://github.com/filecoin-project/sentinel-visor/pull/59

commands/run.go
```diff
@@ -0,0 +1,308 @@
...
+			return nil, fmt.Errorf("invalid cid: %w", err)
+		}
+		codes = append(codes, c)
+	}
+
+	return codes, nil
+}
+
+var actorNamesToCodes = map[string]cid.Cid{
+	"fil/2/system":           builtin.SystemActorCodeID,
```
willscott, line 297: can you use [`ActorNameByCode`](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/codes.go#L77) to get these, rather than duplicating them?

frrist (approved):
Awesome stuff :+1: 

I'm not a huge fan of long lived feature branches and would be happy to take this as it exists right now. I think we should be able to address issues we find with follow on PR's. My approval is sticky, great work @iand.

tasks/actorstate/actorstatechange.go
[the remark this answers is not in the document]
placer14: ACK. Will track this concern for later.

tasks/indexer/tipset_test.go
```diff
@@ -0,0 +1,502 @@
...
+
+		ts16 := mustMakeTs(nil, 16, dummyCid)
+		_, err = c.Add(ts16)
+		require.NoError(t, err)
+		assert.Equal(t, 3, c.Len())
+
+		ts15alt := mustMakeTs(nil, 15, dummyCid)
+		err = c.SetCurrent(ts15alt)
+		assert.NoError(t, err)
+		assert.Equal(t, 2, c.Len()) // ts16 has been reverted, ts16 replaced by ts16alt
```
placer14, line 391:
```suggestion
		assert.Equal(t, 2, c.Len()) // ts16 has been reverted, ts16 replaced by ts15alt
```

placer14 (approved):
LGTM 🤝

tasks/indexer/chainheadindexer.go
placer14: Seems to be resolved. :)

commands/run.go
placer14: I think we're going to remove these literals in exchange for tokens which are easy to pass into the CLI... (ex: `storageminer` -> all versions of `/fil/X/storageminer`)

PR #45, opened 2020-09-29
add metrics

lanzafame:


metrics/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+)
+
+var defaultMillisecondsDistribution = view.Distribution(0.01, 0.05, 0.1, 0.3, 0.6, 0.8, 1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1000, 2000, 5000, 10000, 20000, 50000, 100000)
+
+var (
+	TaskNS, _ = tag.NewKey("namespace")
```
placer14, line 14: Why generic name? We might consider naming it for the app instead?
```suggestion
	TaskNS, _ = tag.NewKey("visor")
```

placer14 (approved):
GTM. Thanks for putting this together.

iand:
I'm happy to rework this for the new scheduler changes if you like @lanzafame

tasks/tasks.go
```diff
@@ -0,0 +1,27 @@
+package tasks
```
frrist, line 1: I _think_ we can delete these, these are no longer used after the refactor.

model/actors/init/idaddress.go
```diff
@@ -38,6 +44,14 @@ func (ias IdAddressList) PersistWithTx(ctx context.Context, tx *pg.Tx) error {
 }
 
 func (ias IdAddressList) Persist(ctx context.Context, db *pg.DB) error {
+	ctx, _ = tag.New(ctx, tag.Upsert(metrics.TaskNS, tasks.InitActorPoolName))
+	stats.Record(ctx, metrics.TaskQueueLen.M(-1))
+
+	start := time.Now()
+	defer func() {
+		stats.Record(ctx, metrics.PersistDuration.M(metrics.SinceInMilliseconds(start)))
+	}()
```
frrist, line 53: It would be great it this was a timer thingy in the `metrics/` package so we could do things like:
```go
mt := NewMetricTimer()
defer mt.Stop(ctx)
```

model/actors/miner/task.go
```diff
@@ -41,6 +47,15 @@ type MinerTaskResult struct {
 func (mtr *MinerTaskResult) Persist(ctx context.Context, db *pg.DB) error {
 	ctx, span := global.Tracer("").Start(ctx, "MinerTaskResult.Persist")
 	defer span.End()
+
+	ctx, _ = tag.New(ctx, tag.Upsert(metrics.TaskNS, tasks.MinerPoolName))
+	stats.Record(ctx, metrics.TaskQueueLen.M(-1))
```
frrist, line 52: :question: If this is tagged before creating the trace above, will the trace have the tags on it?

frrist (commented):
Good stuff, excited to have this!

model/actors/miner/task.go
lanzafame: Yes it will, I can move it above if you would like?

tasks/tasks.go
lanzafame: Yeah, rebasing currently.

commands/setup.go
```diff
@@ -180,3 +188,46 @@ func (g *GlobalSingleton) Lock(ctx context.Context) error {
...
+		mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+		mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+		mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+		mux.Handle("/debug/pprof/block", pprof.Handler("block"))
+		mux.Handle("/debug/pprof/goroutine", pprof.Handler("goroutine"))
+		mux.Handle("/debug/pprof/heap", pprof.Handler("heap"))
+		mux.Handle("/debug/pprof/mutex", pprof.Handler("mutex"))
+		mux.Handle("/debug/pprof/threadcreate", pprof.Handler("threadcreate"))
+		if err := http.ListenAndServe(cctx.String("prometheus-port"), mux); err != nil {
+			log.Fatalf("Failed to run Prometheus /metrics endpoint: %v", err)
```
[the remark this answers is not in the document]
lanzafame: Most common reason this fails is the port that has been passed in is already in use and it is very easy to miss the log saying that the port is already in use.

model/actors/init/idaddress.go
```diff
@@ -38,6 +44,14 @@ func (ias IdAddressList) PersistWithTx(ctx context.Context, tx *pg.Tx) error {
 }
 
 func (ias IdAddressList) Persist(ctx context.Context, db *pg.DB) error {
+	ctx, _ = tag.New(ctx, tag.Upsert(metrics.TaskNS, tasks.InitActorPoolName))
+	stats.Record(ctx, metrics.TaskQueueLen.M(-1))
```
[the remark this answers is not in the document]
lanzafame: The queuelen is going to get removed as this is not handled by redis anymore. But regarding creating a wrapper function, personally, I don't want to have to create a wrapper func for every metric that is created as it makes adding further tags more difficult.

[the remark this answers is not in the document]
lanzafame: Also, the tagging and recording of the metric is not always as straightforward as this instance.

iand (approved):
LGTM

metrics/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// Timer is a function stopwatch, calling it starts the timer,
+// calling the returned function will record the duration.
+func Timer(ctx context.Context, m *stats.Float64Measure) func() {
+	start := time.Now()
+	return func() {
+		stats.Record(ctx, m.M(SinceInMilliseconds(start)))
+	}
+}
```
frrist, line 62: :heart:

tasks/actorstate/actorstate.go
```diff
@@ -190,6 +198,8 @@ func (p *ActorStateProcessor) processActor(ctx context.Context, info ActorInfo)
 		return xerrors.Errorf("no extractor defined for actor code %q", info.Actor.Code.String())
 	}
 
+	ctx, _ = tag.New(ctx, tag.Upsert(metrics.TaskType, builtin.ActorNameByCode(info.Actor.Code)))
```
frrist, line 201: would it be worth including this tag on the trace, or does that happen automagically?

frrist (approved):
:shipit:

tasks/actorstate/actorstate.go
lanzafame: I can include it in the trace

PR #48, opened 2020-09-30
fix: add migration to remove old chainwatch schema constraints

iand:


.gitignore
```diff
@@ -15,5 +15,6 @@
 # vendor/
 .idea
 visor
+sentinel-visor
```
placer14, line 18: Is a build still using this name somewhere?

placer14 (approved):
GTM

.gitignore
iand: I use `go build` a lot when testing

PR #49, opened 2020-09-30
chore: Push docker images on all *-master branch updates

placer14:


ognots:
you also have to adjust `scripts/push-docker-tags.sh `
see https://github.com/filecoin-project/sentinel-drone/blob/master/scripts/push-docker-tags.sh

ognots:
circleci workflows also needs a tag filter
```            
tags:
  only: /.*/
```

.circleci/config.yml
```diff
@@ -77,5 +77,6 @@ workflows:
               only:
                 - master
                 - main
+                - /.*-master$/
             tags:
               only: /^v[0-9].*/
```
ognots, line 82: change to /.*/

ognots (approved):

placer14:
Unfortunate that `chore/docker-push-on-all-master` matches the new filter and attempted to push. It failed due to the slash in the branch name. We could further restrict the pattern to match only a single word before `-master`.

PR #55, opened 2020-10-05
fix(migration): Remove set ownership

placer14:
These statements were causing
```
2020-10-04T03:00:06.280Z	FATAL	visor	sentinel-visor/main.go:102	migrate schema: run migration: ERROR #42501 must be member of role "postgres"
```


ognots:
doesn't fully resolve issue, still getting error in timescaledb

iand:
This should be resolved now, there was an additional grant of ownership to postgres.

PR #57, opened 2020-10-05
fix(headindexer): fix tipsetcache comparison of reverted tipsets

iand:


placer14 (approved):
Seems GTM. Unsure what the return (displaced tipset) on that Add does and what the impact of returning `nil` is here. If it's ignored and only there for accounting, I have no further concerns.

placer14:
Found that the returned (displaced) tipset is what is ultimately queued up for processing. The cache size is managed by the `visor run --indexhead-confidence`. All good with this.

PR #59, opened 2020-10-05
feat: Add flag to remove lock on startup

placer14:
Something temporary to fix `2020-10-05T18:18:04.584Z        INFO    schedule        schedule/scheduler.go:79        task not started: lock not acquired     {"task": "ChainHeadIndexer"}`

### DIFF
--- a/commands/run.go
+++ b/commands/run.go
@@ -53,6 +53,13 @@ var Run = &cli.Command{
 			EnvVars: []string{"VISOR_INDEXHISTORY"},
 		},
 
+		&cli.BoolFlag{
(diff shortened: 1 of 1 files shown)


iand:
This is the message I would expect if another indexer is running.

placer14:
There's definitely not another indexer. This also doesn't appear to be working as I expected... :/

iand:
Can you rebase to make it easier to see the change?

placer14:
Alright, closing this and will revisit if this becomes a bigger problem later.

PR #67, opened 2020-10-06
feat: compute gas outputs

iand:
This change adds a gas outputs processing task. The results of the task are written to
a new table derived_gas_outputs. The task fetches batches of messages that have receipts
and calls the lotus vm ComputeGasOutputs method, persisting the results.

This change also performs a significant refactor of the visor processing tables. Previously
the chain indexers would write duplicate rows to visor_processing_statechanges and
visor_processing_messages. Each of those tables contained claimed_until, completed_at and
errors_detected columns. They were used by the ActorStateChange and Message tasks. However
if we were to add another task that processed tipsets we would need to add another processing
table and then reindex the entire chain to populate it.

To avoid this, the visor_processing_statechanges and visor_processing_messages tables have
been combined into a single visor_processing_tipsets table. Each task now has a set of three
columns in this table to track its work: statechange_claimed_until/statechange_completed_at/
statechange_errors_detected and message_claimed_until/message_completed_at/message_errors_detected.

Adding a new tipset-oriented task is then a matter of adding three columns to this table.

The visor_processing_messages table has been repurposed to track work that needs to be performed
on each message seen. The GasOutputs task is the first task to use this table. It uses the
gas_outputs_claimed_until, gas_outputs_completed_at and gas_outputs_errors_detected columns.

Additional message-oriented tasks can be supported by adding new columns to this table, avoiding
the need to read every message in the chain again.

fixes https://github.com/filecoin-project/sentinel-visor/issues/63

### DIFF
--- a/model/derived/gasoutputs.go
+++ b/model/derived/gasoutputs.go
@@ -1,0 +1,64 @@
+package derived
+
+import (
+	"context"
+
+	"github.com/go-pg/pg/v10"
+	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/trace"
+	"go.opentelemetry.io/otel/label"
+	"golang.org/x/xerrors"
+)
+
+type GasOutputs struct {
+	tableName          struct{} `pg:"derived_gas_outputs"`
+	Cid                string   `pg:",pk,notnull"`
+	From               string   `pg:",notnull"`
+	To                 string   `pg:",notnull"`
+	Value              string   `pg:",notnull"`
+	GasFeeCap          string   `pg:",notnull"`
+	GasPremium         string   `pg:",notnull"`
+	GasLimit           int64    `pg:",use_zero,notnull"`
+	SizeBytes          int      `pg:",use_zero,notnull"`
+	Nonce              uint64   `pg:",use_zero,notnull"`
+	Method             uint64   `pg:",use_zero,notnull"`
+	StateRoot          string   `pg:",notnull"`
+	ExitCode           int64    `pg:",use_zero,notnull"`
+	GasUsed            int64    `pg:",use_zero,notnull"`
+	ParentBaseFee      string   `pg:",notnull"`
+	BaseFeeBurn        string   `pg:",notnull"`
+	OverEstimationBurn string   `pg:",notnull"`
+	MinerPenalty       string   `pg:",notnull"`
+	MinerTip           string   `pg:",notnull"`
+	Refund             string   `pg:",notnull"`
+	GasRefund          int64    `pg:",use_zero,notnull"`
+	GasBurned          int64    `pg:",use_zero,notnull"`
+}
+
+func (g *GasOutputs) PersistWithTx(ctx context.Context, tx *pg.Tx) error {
+	if _, err := tx.ModelContext(ctx, g).
+		OnConflict("do nothing").
+		Insert(); err != nil {
+		return xerrors.Errorf("persisting derived gas outputs: %w", err)
+	}
+	return nil
+}
+
+type GasOutputsList []*GasOutputs
+
+func (l GasOutputsList) Persist(ctx context.Context, db *pg.DB) error {
+	return db.RunInTransaction(ctx, func(tx *pg.Tx) error {
+		return l.PersistWithTx(ctx, tx)
+	})
+}
+
+func (l GasOutputsList) PersistWithTx(ctx context.Context, tx *pg.Tx) error {
+	ctx, span := global.Tracer("").Start(ctx, "GasOutputsList.PersistWithTx", trace.WithAttributes(label.Int("count", len(l))))
+	defer span.End()
+	for _, p := range l {
+		if err := p.PersistWithTx(ctx, tx); err != nil {
+			return nil
+		}
+	}
+	return nil
+}
--- a/storage/migrations/6_tipset_processing_tables.go
+++ b/storage/migrations/6_tipset_processing_tables.go
@@ -1,0 +1,117 @@
+package migrations
+
+import (
+	"github.com/go-pg/migrations/v8"
+)
+
+// Schema version 6 merges tipset oriented processing tables
+
+func init() {
+
+	up := batch(`
+-- Merge visor_processing_statechanges and visor_processing_messages tables into a new tipset table
+ALTER TABLE public.visor_processing_statechanges RENAME TO visor_processing_tipsets;
+ALTER TABLE public.visor_processing_tipsets RENAME COLUMN claimed_until TO statechange_claimed_until;
+ALTER TABLE public.visor_processing_tipsets RENAME COLUMN completed_at TO statechange_completed_at;
+ALTER TABLE public.visor_processing_tipsets RENAME COLUMN errors_detected TO statechange_errors_detected;
+ALTER TABLE public.visor_processing_tipsets ADD COLUMN message_claimed_until timestamptz;
+ALTER TABLE public.visor_processing_tipsets ADD COLUMN message_completed_at timestamptz;
+ALTER TABLE public.visor_processing_tipsets ADD COLUMN message_errors_detected text;
+
+-- Copy data from visor_processing_messages before it is repurposed
+UPDATE public.visor_processing_tipsets t
+SET message_claimed_until = m.claimed_until,
+    message_completed_at = m.completed_at,
+    message_errors_detected = m.errors_detected
+FROM public.visor_processing_messages m
+WHERE m.tip_set = t.tip_set;
+
+
+-- Repurpose visor_processing_messages table
+DROP TABLE IF EXISTS public.visor_processing_messages;
+
+CREATE TABLE public.visor_processing_messages (
+	"cid" text NOT NULL,
+	"height" bigint,
+	"added_at" timestamptz NOT NULL,
+	"gas_outputs_claimed_until" timestamptz,
+	"gas_outputs_completed_at" timestamptz,
+	"gas_outputs_errors_detected" text,
+	PRIMARY KEY ("cid")
+);
+
+CREATE TABLE IF NOT EXISTS public.derived_gas_outputs (
+	"cid" text NOT NULL,
+	"from" text NOT NULL,
+	"to" text NOT NULL,
+	"value" text NOT NULL,
+	"gas_fee_cap" text NOT NULL,
+	"gas_premium" text NOT NULL,
+	"gas_limit" bigint,
+	"size_bytes" bigint,
+	"nonce" bigint,
+	"method" bigint,
+	"state_root" text NOT NULL,
+	"exit_code" bigint NOT NULL,
+	"gas_used" bigint NOT NULL,
+	"parent_base_fee" text NOT NULL,
+	"base_fee_burn" text NOT NULL,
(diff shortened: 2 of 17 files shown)


storage/sql.go
```diff
@@ -499,3 +430,76 @@ func useNullIfEmpty(s string) *string {
...
+func (d *Database) MostRecentSyncedBlock(ctx context.Context) (*blocks.BlockSynced, error) {
+	panic("REMOVED")
+}
+
+func (d *Database) CollectAndMarkBlocksAsProcessing(ctx context.Context, batch int) (blocks.BlocksSynced, error) {
+	panic("REMOVED")
+}
+
+// LeaseGasOutputsMessages leases a set of messages that have receipts for gas output processing. minHeight and maxHeight define an inclusive range of heights to process.
+func (d *Database) LeaseGasOutputsMessages(ctx context.Context, claimUntil time.Time, batchSize int, minHeight, maxHeight int64) (derived.GasOutputsList, error) {
```
iand, line 447: This function contains a query that joins messages to receipts to only return messages that have corresponding receipts in the database. It also joins to block_headers to get the parent_base_fee. Is this the correct value to use?

model/derived/gasoutputs.go
```diff
@@ -0,0 +1,70 @@
...
+	"context"
+
+	"github.com/go-pg/pg/v10"
+	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/trace"
+	"go.opentelemetry.io/otel/label"
+	"golang.org/x/xerrors"
+)
+
+type GasOutputs struct {
```
iand, line 13: This is what is persisted in the new `derived_gas_outputs` table

storage/sql.go
frrist: > This function contains a query that joins messages to receipts to only return messages that have corresponding receipts in the database

It should never be the case that a message without a receipt exists in the tables produced by visor since it operates on valid blocks, I believe this join on block_headers is unnecessary if its only for ensuring messages are returned that have a receipt. ([spec-ref](https://spec.filecoin.io/#systems__filecoin_vm__message__message-semantic-validation)) 

> It also joins to block_headers to get the parent_base_fee. Is this the correct value to use?

"The ParentBaseFee is used when calculating how much a sender burns when executing a message. Burning simply refers to sending attoFIL to a dedicated, unreachable account. A message causes ParentBaseFee * GasUsed attoFIL to be burnt" - ([lotus-ref](https://github.com/filecoin-project/lotus/blob/master/CHANGELOG.md#network-fee)). 

The `ParentBaseFee` value being extracted looks right to me :+1:

tasks/message/gasoutputs.go
```diff
@@ -0,0 +1,132 @@
...
+	feeCap, err := big.FromString(item.GasFeeCap)
+	if err != nil {
+		return xerrors.Errorf("parse fee cap: %w", err)
+	}
+	gasPremium, err := big.FromString(item.GasPremium)
+	if err != nil {
+		return xerrors.Errorf("parse gas premium: %w", err)
+	}
+
+	outputs := vm.ComputeGasOutputs(item.GasUsed, item.GasLimit, baseFee, feeCap, gasPremium)
```
frrist, line 115: :+1:

frrist (approved):

model/derived/gasoutputs.go
frrist: Any reason for duplicating the message and receipt data in this table? Is it to avoid complicating our ingestion strategy?

placer14: This is intended to be query optimized (indexed, normalized) format for supporting analysis and monitoring. This (query-optimized) schema may eventually live in another place so we shouldn't assume the presence of our ingestion schema. It also cleanly isolates our ingestion concerns from our analysis concerns.

storage/sql.go
iand: I think the join between messages and receipts is necessary so as to ensure we only process messages that have received a receipt, which happens at some time in the future. The indexer will add messages and then some time later will find the receipts for those messages so there will be a period of time where we have messages with no receipts in the database.

Plus we need the gas_used value that is only in the receipt. The join to block headers it to retrieve the parent_base_fee. Both of these are needed for the call to ComputeGasOutputs.

model/derived/gasoutputs.go
iand: Columns are as requested in https://github.com/filecoin-project/sentinel-visor/issues/63

storage/migrations/6_tipset_processing_tables.go
```diff
@@ -0,0 +1,109 @@
...
+	"return" bytea,
+	"parent_base_fee" text,
+	"base_fee_burn" text,
+	"over_estimation_burn" text,
+	"miner_penalty" text,
+	"miner_tip" text,
+	"refund" text,
+	"gas_refund" bigint,
+	"gas_burned" bigint,
+	PRIMARY KEY ("cid")
```
placer14, line 68: Indexes to include here:
- `from` (using Hash)
- `to` (using Hash)
- `method`(using BTree)
- `exit_code` (using BTree)

storage/sql.go
```diff
@@ -364,27 +367,27 @@ func (d *Database) MarkActorComplete(ctx context.Context, head string, code stri
 	return nil
 }
 
-// LeaseActors leases a set of tipsets containing messages to process. minHeight and maxHeight define an inclusive range of heights to process.
-func (d *Database) LeaseTipSetMessages(ctx context.Context, claimUntil time.Time, batchSize int, minHeight, maxHeight int64) (visor.ProcessingMessageList, error) {
-	var messages visor.ProcessingMessageList
+// LeaseTipSetMessages leases a set of tipsets containing messages to process. minHeight and maxHeight define an inclusive range of heights to process.
```
placer14, line 370: Nice catch.

storage/sql.go
placer14: @frrist Thanks for the response. The join on `block_headers` is for returning `parent_base_fee`.

model/derived/gasoutputs.go
```diff
@@ -0,0 +1,70 @@
...
+	BaseFeeBurn        string
+	OverEstimationBurn string
+
+	MinerPenalty string
+	MinerTip     string
+	Refund       string
+
+	GasRefund int64
+	GasBurned int64
+}
```
placer14, line 42: I think all of these columns should be `notnull` (even the cases which `use_zero`).

storage/sql.go
```diff
@@ -427,3 +430,64 @@ func useNullIfEmpty(s string) *string {
...
+
+func (d *Database) MarkGasOutputsMessagesComplete(ctx context.Context, cid string, completedAt time.Time, errorsDetected string) error {
+	if err := d.DB.RunInTransaction(ctx, func(tx *pg.Tx) error {
+		_, err := tx.ExecContext(ctx, `
+    UPDATE visor_processing_messages
+    SET gas_outputs_claimed_until = null,
+        gas_outputs_completed_at = ?,
+        gas_outputs_errors_detected = ?
+    WHERE cid = ?
+`, completedAt, useNullIfEmpty(errorsDetected), cid)
```
placer14, line 482: > `useNullIfEmpty`

Love it.

tasks/message/gasoutputs.go
placer14: Can we express this within the Lens abstraction? This will keep our lotus deps within that package and make it easy to diversify our sources later.

placer14 (changes requested):
A few changes and concerns to address. 🤝

storage/sql.go
placer14: BTW, this query looks good to me.

placer14 (approved):
@iand In the interest of allowing this to move along before I'm up, I'll put a ✅ now. I know you'll see the feedback is applied appropriately. 🤝

placer14:
FYI, restarts are very aggressive.
```
2020-10-07T03:56:03.596Z        INFO    schedule        schedule/scheduler.go:124       task exited cleanly     {"task": "ChainHistoryIndexer"}
2020-10-07T03:56:03.596Z        INFO    schedule        schedule/scheduler.go:106       restarting task {"task": "ChainHistoryIndexer"}
2020-10-07T03:56:03.663Z        INFO    schedule        schedule/scheduler.go:124       task exited cleanly     {"task": "ChainHistoryIndexer"}
2020-10-07T03:56:03.663Z        INFO    schedule        schedule/scheduler.go:106       restarting task {"task": "ChainHistoryIndexer"}
2020-10-07T03:56:03.728Z        INFO    schedule        schedule/scheduler.go:124       task exited cleanly     {"task": "ChainHistoryIndexer"}
2020-10-07T03:56:03.728Z        INFO    schedule        schedule/scheduler.go:106       restarting task {"task": "ChainHistoryIndexer"}
2020-10-07T03:56:03.792Z        INFO    schedule        schedule/scheduler.go:124       task exited cleanly     {"task": "ChainHistoryIndexer"}
2020-10-07T03:56:03.792Z        INFO    schedule        schedule/scheduler.go:106       restarting task {"task": "ChainHistoryIndexer"}
2020-10-07T03:56:03.856Z        INFO    schedule        schedule/scheduler.go:124       task exited cleanly     {"task": "ChainHistoryIndexer"}
2020-10-07T03:56:03.856Z        INFO    schedule        schedule/scheduler.go:106       restarting task {"task": "ChainHistoryIndexer"}
2020-10-07T03:56:03.920Z        INFO    schedule        schedule/scheduler.go:124       task exited cleanly     {"task": "ChainHistoryIndexer"}
2020-10-07T03:56:03.920Z        INFO    schedule        schedule/scheduler.go:106       restarting task {"task": "ChainHistoryIndexer"}
```

Additionally, the higher worker counts on a pretty beefy box end up resulting in a lot of this, which is also exacerbated by the aggressive restart.

```2020-10-07T03:54:32.995Z        ERROR   schedule        schedule/scheduler.go:117       task exited with failure        {"task": "GasOutputsProcessor013", "error": "FATAL #53300 sorry, too many clients al
ready"}
2020-10-07T03:54:32.995Z        INFO    schedule        schedule/scheduler.go:106       restarting task {"task": "GasOutputsProcessor013"}```

File: storage/migrations/6_tipset_processing_tables.go (hunk 4 above)
Comment (iand) [reply storage/migrations/6_tipset_processing_tables.go:68]: Done

File: model/derived/gasoutputs.go [hunk 8]
```diff
@@ -0,0 +1,70 @@
...
+	SizeBytes     int      `pg:",use_zero"`
+	Nonce         uint64   `pg:",use_zero"`
+	Method        uint64   `pg:",use_zero"`
+	Params        []byte
+	StateRoot     string `pg:",notnull"`
+	Idx           int    `pg:",use_zero"`
+	ExitCode      int64  `pg:",use_zero"`
+	GasUsed       int64  `pg:",use_zero"`
+	Return        []byte
+	ParentBaseFee string
```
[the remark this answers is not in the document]
iand: Done

model/derived/gasoutputs.go
iand: Done

iand:
Moved ComputeGasOutputs into the node abstraction as suggested by @placer14 but github won't let me reply to that comment

PR #69, opened 2020-10-07
chore(cli): nice supported actor args

frrist:
Usage now reads (irrelevant parts omitted for clarity):
```
frrist@sequoia:~/s/g/f/sentinel-visor|frrist/actor-by-name✓
➤ ./visor run --help
NAME:
   visor run - Index and process blocks from the filecoin blockchain

USAGE:
   visor run [command options] [arguments...]

OPTIONS:
...
   --actorstate-include value                List of actor names that should be processed by actor state processors: [account cron init market miner multisig payment power registry reward system] [$VISOR_ACTORSTATE_INCLUDE]
   --actorstate-exclude value                List of actor names that should be not be processed by actor state processors: [account cron init market miner multisig payment power registry reward system] [$VISOR_ACTORSTATE_EXCLUDE]
...

### DIFF
--- a/commands/run.go
+++ b/commands/run.go
@@ -98,13 +98,13 @@ var Run = &cli.Command{
 		},
 		&cli.StringSliceFlag{
 			Name:        "actorstate-include",
-			Usage:       "List of actor codes that should be procesed by actor state processors",
+			Usage:       fmt.Sprintf("List of actor names that should be processed by actor state processors: %v", supportedActorArgs),
 			DefaultText: "all supported",
 			EnvVars:     []string{"VISOR_ACTORSTATE_INCLUDE"},
 		},
 		&cli.StringSliceFlag{
 			Name:        "actorstate-exclude",
-			Usage:       "List of actor codes that should be not be procesed by actor state processors",
+			Usage:       fmt.Sprintf("List of actor names that should be not be processed by actor state processors: %v", supportedActorArgs),
 			DefaultText: "none",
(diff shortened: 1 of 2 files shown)


placer14 (approved):
🍻 GTM!

iand:
I think it's a little confusing because these are short names for actors, not their codes. Maybe reword to "list of actor codes or names"?

iand:
Another thought: my understanding is that the v1 and v2 actors have different codes. The cid for an actor code is the multihash of its name, either `fil/1/storageminer` or `fil/2/storageminer`.

frrist:
The visor actor state extractors are agnostic to the actor version they operate on. This is achieved by using the [actor-shim interfaces from lotus](https://github.com/filecoin-project/lotus/tree/master/chain/actors/builtin#actors). I think this negates the need to supply which actor version to include/exclude for extraction when passing args to visor, and due to this I've simplified the arg requirements to simple actor names.

I will change the help text to read `List of actor names that should/should-not be processed by actor state processors.` Does that sound reasonable?

iand:
I'm happy for the pr to be merged. But we should investigate the actor code issue a bit more. The state change processor writes the code for each statechange into the visor_processing_actors table. The actor state processor then selects from this table specifying which codes it can work on, plus does a switch on the code. If actors v1 and v2 have different codes then we need to select both from the table and add them to the switch statement.

frrist:
I don't think this PR adds a lot of value, and in light of this conversation it feels premature. I'm going to close it until we have better understanding of how we intend to process different actor version.

Issue #61, opened 2020-10-05
Capture Miner Deadlines


placer14:
In order to understand WindowPoSt commitments and how they are being maintained on-chain, we will extract StorageMinerActor.Deadlines and its related data into Sentinel schema for monitoring and analysis.

Requirements:
1. Produced schema should express ability to query for
  a. PoSts expected (per miner, per sector, per deadline CID (unsure how to uniquely identify deadlines))
  b. Messages (as well as other mechanisms required by Miner to fulfill successful PoSt) which satisfy expected PoSts

Deliverables:
- [ ] Migration which relates the following objects `Miners` --hasmany-> `Deadlines` --hasmany-> `Partitions` --hasmany-> `Sectors`
- [ ] StorageMinerActor extractors which support the schema produced in the previous migration
- [ ] Example query (or pseudoquery) expressing ability to achieve R1 (a and b)

frrist:
@willscott did #74 resolve this? If not, what is left to do here?

willscott:
that's correct:
* the current table is populated when relevant miner state changes across tipsets. this is the correct and sufficient trigger.
* When this happens, we find all relevant messages, and populate message, miner, and affected sectors.
* Miner changes may also be unrelated to messages, and be due to cron-based expiry of post windows. we initially planned to populate these with an empty message ID. It is somewhat unclear to me how valuable that additional signal is.

is the p1 adding this additional set of rows, or having a generated table and seeing if it's good enough for windowPoSt analysis?

PR #74, opened 2020-10-07
Add miner_sector_posts tracking of window posts

willscott:
* [x] Data model
* [x] Task for populating

Address #61


model/actors/miner/sectorposts.go
```diff
@@ -0,0 +1,54 @@
...
+	"go.opentelemetry.io/otel/label"
+	"golang.org/x/xerrors"
+)
+
+type MinerSectorPost struct {
+	MinerID  string `pg:",pk,notnull"`
+	SectorID uint64 `pg:",pk,use_zero"`
+	Epoch    int64  `pg:",pk,notnull"`
+
+	PostMessageID string
```
placer14, line 18: Nit:
```suggestion
	PostMessageCID string
```

storage/migrations/5_windowpost.go
```diff
@@ -0,0 +1,24 @@
...
+// Schema version 5 provides additional tracking of the miner->deadline->partition->sector relation
+
+func init() {
+
+	up := batch(`
+	CREATE TABLE IF NOT EXISTS "miner_sector_posts" (
+		"miner_id" text,
+		"sector_id" number,
+		"epoch" bigint,
+		"post_message_id" ?,
```
placer14, line 16: Same some disk space? And match null constraints in the model.
```suggestion
		"miner_id" text not null,
		"sector_id" int not null,
		"epoch" int not null,
		"post_message_id" text not null,
```

model/actors/miner/sectorposts.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/go-pg/pg/v10"
+	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/trace"
+	"go.opentelemetry.io/otel/label"
+	"golang.org/x/xerrors"
+)
+
+type MinerSectorPost struct {
+	MinerID  string `pg:",pk,notnull"`
+	SectorID uint64 `pg:",pk,use_zero"`
```
placer14, line 15:
```suggestion
	SectorID uint64 `pg:",pk,notnull,use_zero"`
```

placer14 (commented):
Minor suggestions. 🤝

storage/migrations/5_windowpost.go
```diff
@@ -0,0 +1,24 @@
...
+// Schema version 5 provides additional tracking of the miner->deadline->partition->sector relation
+
+func init() {
+
+	up := batch(`
+	CREATE TABLE IF NOT EXISTS "miner_sector_posts" (
+		"miner_id" text not null,
+		"sector_id" int not null,
+		"epoch" int not null,
+		"post_message_id" text not null,
```
willscott, line 16:
```suggestion
		"post_message_cid" text,
```

placer14: Quite right. Sorry.

model/actors/miner/sectorposts.go
```diff
@@ -0,0 +1,54 @@
...
+func (msp *MinerSectorPost) PersistWithTx(ctx context.Context, tx *pg.Tx) error {
+	if _, err := tx.ModelContext(ctx, msp).
+		OnConflict("do nothing").
+		Insert(); err != nil {
+		return xerrors.Errorf("persisting miner sector window post: %w", err)
+	}
+	return nil
+}
+
+// TODO: what makes this.
```
frrist, line 32: I would start by looking here: https://github.com/filecoin-project/sentinel-visor/blob/master/tasks/actorstate/miner.go#L26

model/actors/miner/sectorposts.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+
+	"github.com/go-pg/pg/v10"
+	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/trace"
+	"go.opentelemetry.io/otel/label"
+	"golang.org/x/xerrors"
+)
+
+type MinerSectorPost struct {
```
frrist, line 13: You'll want to add this to the list here: https://github.com/filecoin-project/sentinel-visor/blob/master/storage/sql.go#L29

storage/migrations/5_windowpost.go
```diff
@@ -0,0 +1,24 @@
...
+
+// Schema version 5 provides additional tracking of the miner->deadline->partition->sector relation
+
+func init() {
+
+	up := batch(`
+	CREATE TABLE IF NOT EXISTS "miner_sector_posts" (
+		"miner_id" text not null,
+		"sector_id" int not null,
+		"epoch" int not null,
```
frrist, line 15: `bigint`

tasks/actorstate/miner.go
```diff
@@ -94,9 +110,84 @@ func (m StorageMinerExtractor) Extract(ctx context.Context, a ActorInfo, node le
...
+	loadPartitions := func(state miner.State, epoch abi.ChainEpoch) ([]miner.Partition, error) {
+		info, err := state.DeadlineInfo(epoch)
+		if err != nil {
+			return nil, err
+		}
+		dline, err := state.LoadDeadline(info.Index)
+		if err != nil {
+			return nil, err
+		}
+		partitions := make([]miner.Partition, 0, 8)
```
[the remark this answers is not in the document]
willscott: just a capacity hint. the largest number of partitions I've seen through manual observation of deadlines of miners so far is 8. I don't think it's a hard limit, but should generally help the allocator.

storage/migrations/5_windowpost.go
placer14: I suggested he correct this to `int`, you're right though.

placer14: I suggested he correct this to `int`, you're right though.

storage/migrations/5_windowpost.go
```diff
@@ -0,0 +1,24 @@
+package migrations
```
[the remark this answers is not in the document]
placer14: It'll be 8 now. This is going to continue to be a problem while we have an integer index for migrations. Perhaps it would be better to use timestamps to make the index collisions (at least between migrations touching mutually exclusive schema) less painful to reconcile. But maybe this is a short term problem while dev velocity is high. Just a showerthought.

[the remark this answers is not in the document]
iand: The migrations package we use expects a monotonic counter. It will fail if there are two migrations for the same schema version. I'll add a test for that so we see it at build time instead of migration time.

frrist (approved):
Good stuff, have you been able to manually verify this works against a lotus node and database?

storage/migrations/8_windowpost.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// Schema version 8 provides additional tracking of the miner->deadline->partition->sector relation
+
+func init() {
+
+	up := batch(`
+	CREATE TABLE IF NOT EXISTS "miner_sector_posts" (
+		"miner_id" text not null,
+		"sector_id" bigserial not null,
```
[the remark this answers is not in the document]
frrist: Just a heads up, bigserial is used for sector_id everywhere(?) else: https://github.com/filecoin-project/sentinel-visor/blob/master/storage/migrations/2_visor_initial.go#L83

Im not convinced bigserial isn't the right choice here and brought up this point [here](https://github.com/filecoin-project/sentinel-visor/pull/74#discussion_r501371926)

storage/migrations/5_windowpost.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// Schema version 5 provides additional tracking of the miner->deadline->partition->sector relation
+
+func init() {
+
+	up := batch(`
+	CREATE TABLE IF NOT EXISTS "miner_sector_posts" (
+		"miner_id" text not null,
+		"sector_id" int not null,
```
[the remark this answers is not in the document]
iand: bigserial is autoincrementing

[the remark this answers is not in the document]
iand: I can't explain why we have it in those three tables :cry:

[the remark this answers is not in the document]
iand: It's because go-pg uses bigserial when a bigint is part of a primary key. See https://github.com/go-pg/pg/blob/v10.3.1/orm/table_create.go#L144 and the `pkSQLType` function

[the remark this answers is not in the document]
iand: `bigserial` is basically an alias for `bigint` with a hidden sequence. You can insert into it normally but it will also take the next increment if you specify DEFAULT as its value. Annoying that we have it but not a disaster. Perhaps we should add a migration to bigint in this PR so they are all consistent.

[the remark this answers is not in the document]
frrist: Tracking here: https://github.com/filecoin-project/sentinel-visor/issues/107

Issue #68, opened 2020-10-06
lens for reading from on disk lotus repo


willscott:
The JSON API against running lotus fullnode is good for staying up with a current node.
when doing historic processing, it will likely be faster and probably path of least resistance to have a snapshot of a lotus repo, but not take the overhead associated with a lotus process and going through a json API.

It would be nice to be able to directly run the badger and library code for accessing an on-disk repo to directly provide the lens API rather than indirecting it through a lotus API.

willscott:
closing this and we can file more specific issues to track / fix bugs as we encounter them

PR #75, opened 2020-10-07
in-process lotus "repo" lens

willscott:
Create a full lotus.Node (without bringing it `Online`) and use its chainstore / api for a lens as an alternative to the jsonrpc API.

address #68

### DIFF
--- a/lens/lotusrepo/repo.go
+++ b/lens/lotusrepo/repo.go
@@ -1,0 +1,63 @@
+package lotusrepo
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/sentinel-visor/lens"
+	"github.com/urfave/cli/v2"
+
+	"github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/chain/vm"
+	"github.com/filecoin-project/lotus/node/"
+	"github.com/filecoin-project/lotus/node/impl"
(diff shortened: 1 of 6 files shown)


lens/lotusrepo/repo.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (ra *RepoAPI) Store() adt.Store {
+	i, ok := ra.FullNode.(*impl.FullNodeAPI)
+	if !ok {
+		return nil
+	}
+	store := i.ChainAPI.Chain.Blockstore()
+	cs := cbor.NewCborStore(store)
+	adtStore := adt.WrapStore(ra.Context, cs)
+	return adtStore
```
frrist, line 37: commentary: At some point we will want caching here, this looks like an easy options: https://github.com/filecoin-project/lotus/blob/master/lib/cachebs/cachebs.go#L22

lens/lotusrepo/repo.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func GetAPI(c *cli.Context) (context.Context, lens.API, lens.APICloser, error) {
+	rapi := RepoAPI{}
+
+	r, err := repo.NewFS(c.String("repo"))
+	if err != nil {
+		return c.Context, nil, nil, err
```
frrist, line 45: why return the context instead of nil?

frrist (commented):
looks like you're on the right path.

frrist (approved):

PR #76, opened 2020-10-08
feat(task): add miner sector event processing

frrist:
Status: 
Blocked on https://github.com/filecoin-project/lotus/pull/4403
- [x] Feature parity with chainwatchv1
- [x] Deploy for manual testing, ensure we observe the following events:
	- [x] PreCommit Add
	- [x] Sector Add
	- [x] Commit Capacity Add
	- [ ] Sector Expire
	- [ ] Sector Terminate
	- [ ] Sector Fault
	- [x] Sector Recovering
	- [x] Sector Recovered  

The box I was given by infra tipped over last night while I was testing that the above events are handled correctly by visor. Given time constraints I would prefer to merge this now and address any problems that may come up when/if they occur.


go.mod
```diff
@@ -5,12 +5,12 @@ go 1.14
...
 	github.com/davidlazar/go-crypto v0.0.0-20200604182044-b73af7476f6c // indirect
-	github.com/filecoin-project/go-address v0.0.3
+	github.com/filecoin-project/go-address v0.0.4
 	github.com/filecoin-project/go-amt-ipld/v2 v2.1.1-0.20200731171407-e559a0579161 // indirect
-	github.com/filecoin-project/go-bitfield v0.2.0
+	github.com/filecoin-project/go-bitfield v0.2.1
 	github.com/filecoin-project/go-jsonrpc v0.1.2-0.20200822201400-474f4fdccc52
 	github.com/filecoin-project/go-state-types v0.0.0-20200911004822-964d6c679cfc
-	github.com/filecoin-project/lotus v0.8.0
+	github.com/filecoin-project/lotus v0.8.2-0.20201007170031-099b982ea831
```
frrist, line 13: we need the latest version of lotus to get the deadline diffing code.

storage/migrations/1_chainwatch.go
```diff
@@ -21,15 +21,14 @@ CREATE EXTENSION IF NOT EXISTS timescaledb WITH SCHEMA public;
 
 CREATE TYPE public.miner_sector_event_type AS ENUM (
```
frrist, line 22: So I've cheated a bit here since this is the first migration. I found it to be a bit of a headache to write the migration for this type and since we are ~burning~ not using the previous chainwatch data this feels relatively harmless, but I will write the migration if that's the desired outcome.

storage/sql.go
```diff
@@ -239,6 +240,8 @@ func verifyModel(ctx context.Context, db *pg.DB, m *orm.Table) error {
 			datatype = "timestamptz"
 		} else if datatype == "timestamp without time zone" {
 			datatype = "timestamp"
+		} else if datatype == "USER-DEFINED" && tableName == "miner_sector_events" {
```
frrist, line 243: The query above on line 230 populates `datatype` with the value `USER-DEFINED` which correct but to vague to be of value for validating. I've hard coded this here since its the only user defined type we have, open to alternatives.

storage/migrations/1_chainwatch.go
iand: Any existing sentinel database won't have the new enum values and you can't safely re-run the version 1 migration against them.

Maybe this migration would work:
```
ALTER TYPE public.miner_sector_event_type ADD VALUE 'PRECOMMIT_REMOVED' AFTER 'PRECOMMIT_ADDED';
ALTER TYPE public.miner_sector_event_type ADD VALUE 'SECTOR_REMOVED' AFTER 'SECTOR_ADDED';
```

Maybe a better solution is to create a new enum type and change the table to use it. That's what I'd do to keep backwards compatibility if this was a protobuf.

storage/sql.go
iand: I think think the `udt_name` column contains the name of the user defined type (udt=user defined type)

frrist: That did the trick, thanks.

tasks/actorstate/actorstate.go
```diff
@@ -50,6 +54,7 @@ type ActorStateAPI interface {
 	StateGetActor(ctx context.Context, actor address.Address, tsk types.TipSetKey) (*types.Actor, error)
 	StateMinerPower(ctx context.Context, addr address.Address, tsk types.TipSetKey) (*api.MinerPower, error)
 	StateReadState(ctx context.Context, actor address.Address, tsk types.TipSetKey) (*api.ActorState, error)
+	StateMinerSectors(context.Context, address.Address, *bitfield.BitField, types.TipSetKey) ([]*miner.SectorOnChainInfo, error)
```
lanzafame, line 57: Odd one out with not naming the parameters

tasks/actorstate/miner.go
```diff
@@ -98,10 +103,171 @@ func (m StorageMinerExtractor) Extract(ctx context.Context, a ActorInfo, node Ac
...
+				Event:     minermodel.PreCommitAdded,
+			})
+		}
+	}
+
+	return out, nil
+}
+
+func (m StorageMinerExtractor) minerPartitionsDiff(ctx context.Context, prevState, curState miner.State) (*minermodel.PartitionStatus, error) {
+	ctx, span := global.Tracer("").Start(ctx, "MinerPartitionDiff")
```
lanzafame, line 230:
```suggestion
	ctx, span := global.Tracer("").Start(ctx, "StorageMinerExtractor.minerPartitionDiff")
```

tasks/actorstate/miner.go
```diff
@@ -98,10 +103,171 @@ func (m StorageMinerExtractor) Extract(ctx context.Context, a ActorInfo, node Ac
...
 		SectorChanges:    sectorChanges,
-		PartitionDiff:    partitionsDiff,
+		SectorEvents:     sectorEvents,
 	}, nil
 }
 
-func (m StorageMinerExtractor) minerPartitionsDiff(ctx context.Context, prevState, curState miner.State) (map[uint64]*minermodel.PartitionStatus, error) {
-	return nil, nil
+func (m StorageMinerExtractor) computeSectorEvents(ctx context.Context, node ActorStateAPI, a ActorInfo, sc *miner.SectorChanges, pc *miner.PreCommitChanges, ps *minermodel.PartitionStatus) (minermodel.MinerSectorEventList, error) {
+	out := minermodel.MinerSectorEventList{}
```
lanzafame, line 111:
```suggestion
	ctx, span := global.Tracer("").Start(ctx, "StorageMinerExtractor.computeSectorEvetns")
	defer span.End()
	
	out := minermodel.MinerSectorEventList{}
```

lanzafame (changes requested):

lanzafame (approved):

willscott (approved):

PR #77, opened 2020-10-08
feat(task): Create chainvis views and refresher

placer14:
This PR serves two purposes:
1. It is a proposal for supporting optional views within Visor. The idea here is that certain views are not always worth maintaining during normal processing. Within, we define a pattern for a materialized view which can be instantiated without data and optionally refreshed by passing in `--chainvis-refresh-rate=1m` with some value greater than 0 (which disables the refresh loop entirely).
2. It is a proposal to support chain visualization views for queries associated with https://github.com/DigitalMOB2/filecoin-lotus-explorer-playground/. This view is a contract for data we promise to support for this external dependency. This will be a UI dependency which sits alongside Grafana as one of out first custom visualizations on top of Sentinel.

Overall, this pattern intends to add zero operational overhead and minimal developer overhead (specifically in maintaining the underlying tables supporting these views). Refresh is completely optional. Creation of materialized views without data initialized are virtually free to maintain in schema. Cost of refreshing views is managed by configurable refresh rate.

Potential areas of improvement:
- capture metrics for view refresh duration
- CI process to test migrations

### DIFF
--- a/tasks/views/chainvis.go
+++ b/tasks/views/chainvis.go
@@ -2,6 +2,7 @@ package views
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"golang.org/x/xerrors"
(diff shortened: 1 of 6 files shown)


tasks/views/chainvis.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Run starts regularly refreshing until context is done or an error occurs
+func (r *ChainVisRefresher) Run(ctx context.Context) error {
+	if r.refreshRate == 0 {
+		return nil
+	}
+	return wait.RepeatUntil(ctx, r.refreshRate, r.refreshView)
+}
+
+func (r *ChainVisRefresher) refreshView(ctx context.Context) (bool, error) {
```
iand, line 35: Minor nit: would be cleaner to have these views in a slice that is looped over

iand (approved):
LGTM with a minor nit

Nice use of the task interface.

frrist (approved):
:ship:

PR #78, opened 2020-10-08
perf: minor optimization of market actor diffing

iand:
The first commit adds a test for market actor state diffing and adds a mock API that will help with writing future unit tests for actor state extraction.

The second commit changes the implementation of market state extraction to combine the deal and proposal diffing into a single predicate so the old and new actor states are only loaded once.


frrist (approved):
LGTM

PR #81, opened 2020-10-09
chore(deps): Update to lotus@v0.9.0

placer14:


willscott (approved):

iand (approved):

version/version.go
```diff
@@ -13,7 +15,7 @@ var reVersion = regexp.MustCompile(`^(v\d+\.\d+.\d+)(?:-)?(.+)?$`)
 func String() string {
 	matches := reVersion.FindStringSubmatch(GitVersion)
 	if matches == nil || len(matches) < 3 {
-		return "v0.0.0+" + GitVersion
+		return ReleaseVersion + "+" + GitVersion
```
iand, line 18: This is incorrect. This line is only encountered when there is no git tag on the repo. We should tag our versions properly.

placer14: Oops. Thanks for raising. Will revert and manage via git tags.

placer14:
Closed in favor of fix included in https://github.com/filecoin-project/sentinel-visor/pull/84

PR #84, opened 2020-10-09
feat: support v2 actor codes

iand:
Note that this branched from #81

Will be needed before the v2 switchover


willscott (approved):

placer14 (approved):

frrist:
I suspect some changes will be needed here for inclusion and exclusion to work as expect? https://github.com/filecoin-project/sentinel-visor/blob/master/commands/run.go#L354

PR #94, opened 2020-10-13
feat(task): add chain economics processing

iand:


tasks/chain/economics.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/filecoin-project/sentinel-visor/storage"
+	"github.com/filecoin-project/sentinel-visor/wait"
+)
+
+const (
+	idleSleepInterval = 60 * time.Second       // time to wait if the processor runs out of blocks to process
+	batchInterval     = 100 * time.Millisecond // time to wait between batches
+)
+
+var log = logging.Logger("message")
```
willscott, line 28: name the logger "ChainEconomics"?

tasks/chain/economics.go
```diff
@@ -0,0 +1,151 @@
...
+			return false, nil // Don't propagate cancelation error so we can resume processing cleanly
+		default:
+		}
+
+		if err := p.processItem(ctx, item); err != nil {
+			log.Errorw("failed to process tipset", "error", err.Error(), "height", item.Height)
+			if err := p.storage.MarkTipSetEconomicsComplete(ctx, item.TipSet, item.Height, p.clock.Now(), err.Error()); err != nil {
+				log.Errorw("failed to mark tipset economics complete", "error", err.Error(), "height", item.Height)
+			}
+			continue
```
willscott, line 99: if processing fails, we'll never actually error out of processing. is that intentional?

willscott (approved):

tasks/chain/economics.go
iand: Yes. A single failed tipset should not fail the entire batch or stop the processor from moving onto a new batch

tasks/chain/economics.go
iand: Renamed to "chain" to match the package name

PR #95, opened 2020-10-13
feat: add --no-default-tasks flag to run subcommand

iand:
**For discussion. An alternative would be simply to set all task worker defaults to false or 0**

This new flag is intended to help us deploy new versions of visor without unexpected
behaviour. We want the `run` subcommand to use sensible defaults for the various tasks
but every time we add a new task with a new default number of workers we may accidentally
start that task on a instance that is configured in a specific way (e.g. just for chain
head indexing)

When --no-default-tasks is specified visor will treat all tasks as though they had zero
workers by default. The number of workers can be set by explicitly specifying them using
a cli flag or environment variable.

For example:

    visor run --no-default-tasks

will not run any tasks at all, whereas

    visor run --no-default-tasks -mw 5 --indexhead

will run only 5 message workers and the head indexer

### DIFF
--- a/commands/run.go
+++ b/commands/run.go
@@ -23,6 +23,11 @@ var Run = &cli.Command{
 	Name:  "run",
 	Usage: "Index and process blocks from the filecoin blockchain",
 	Flags: []cli.Flag{
+		&cli.BoolFlag{
+			Name:    "no-default-tasks",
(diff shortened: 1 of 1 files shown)


willscott:
having all tasks default to false seems a bit more clear to me, but not a strong preference either way

lanzafame:
I agree with @willscott on this. Maybe a `--default` flag that provides "sane" defaults that don't kill node

iand:
Think I would prefer to default to zero workers

PR #96, opened 2020-10-13
feat: add visor processing stats table

iand:
Add a task that writes summary stats to a new table visor_processing_stats.

For each tipset processor (message discovery, state change, chain economics) and
for each message processor (currently only gas outputs) and for each actor state
change it writes the following metrics:

 - total number of items that have been discovered for processing
 - total number of items that have been processed
 - total number of items that have been processed but reported an error
 - total number of items that are currently being processed
 - highest epoch that has been processed
 - highest epoch that has not been processed

Stats are written every minute and only a single global task is used.

Note: branched from #94

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 # vendor/
 .idea
 visor
+!model/visor
 sentinel-visor
 
 build/.*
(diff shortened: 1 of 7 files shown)


iand:
Creating the indexes takes a couple of minutes on a populated database.

PR #97, opened 2020-10-13
feat: add heaviest chain materialized view

iand:
This is the basic view against which we will write grafana queries. The view can be altered as we optimize it but its interface (i.e. columns) should remain the same.

Schema number may need adjusting depending on sequence of commits


storage/migrations/8_heaviest_chain.go
```diff
@@ -0,0 +1,42 @@
...
+	up := batch(`
+CREATE MATERIALIZED VIEW IF NOT EXISTS heaviest_chain_view AS
+WITH RECURSIVE heaviest_chain AS (
+	SELECT
+		b.cid,
+		b.height,
+		b.miner,
+		b.timestamp,
+		b.parent_state_root,
+		b.win_count
```
lanzafame, line 19: Why the `b.win_count` field?

iand: I'm not sure, @placer14 wrote this originally

placer14: Because we want all of our data in our derived view without needing to join on write-heavy DBs.

placer14 (approved):

PR #99, opened 2020-10-14
Base fee and chain capacity tracking.

frrist:
### What
- Implement Message Gas Economy model for tracking Chain Capacity Usage and Base fee.
- Fix what I believe to be a bug in the migration logic by including a null migration at version 0 (I _think_ this might relate to #89 but am not sure).

### DIFF
--- a/tasks/message/message.go
+++ b/tasks/message/message.go
@@ -2,8 +2,13 @@ package message
 
 import (
 	"context"
+	"math"
+	"math/big"
 	"time"
 
+	"github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/build"
+	"github.com/filecoin-project/lotus/chain/store"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/go-pg/pg/v10"
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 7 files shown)


tasks/message/message.go
```diff
@@ -173,54 +184,75 @@ func (p *MessageProcessor) processTipSet(ctx context.Context, ts *types.TipSet)
...
 		for _, message := range vmm {
-			bmsgs = append(bmsgs, &messagemodel.BlockMessage{
+			// record which blocks had which messages
+			bmsgModels = append(bmsgModels, &messagemodel.BlockMessage{
 				Block:   blk.String(),
 				Message: message.Cid().String(),
 			})
 
-			// so we don't create duplicate message models.
+			totalUniqGasLimit += message.GasLimit
```
willscott, line 238: is it useful to double count limit for messages in both bls and secpk?

willscott (approved):

iand:
Rebased onto master and changed the migration number to 10.

iand (approved):
LGTM

iand:
I'm curious what the problem was that required a '0' migration

frrist:
> I'm curious what the problem was that required a '0' migration

When the database is first initialized we run this query (https://github.com/go-pg/migrations/blob/v8/collection.go#L620) against it to get its current version. The gopg_migration table is empty so 0 is returned (https://github.com/go-pg/migrations/blob/v8/collection.go#L624) making the dbVersion at startup 0 (https://github.com/filecoin-project/sentinel-visor/blob/master/storage/migrate.go#L39)

PR #102, opened 2020-10-14
fix(schema): Add index on visor_processing_actors.height

placer14:
Note: This depends on https://github.com/filecoin-project/sentinel-visor/pull/74/files being merged to maintain sequential migrations.

Improve the storage.LeaseActors method which was taking a bit long to select over height.

### DIFF
--- a/storage/migrations/9_processing_actor_height_index.go
+++ b/storage/migrations/9_processing_actor_height_index.go
@@ -1,0 +1,17 @@
+package migrations
+
+import (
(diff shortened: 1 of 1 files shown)


iand:
Note I incorporated this in https://github.com/filecoin-project/sentinel-visor/pull/96

Also you can't use CONCURRENT inside a migration since it runs in a transaction. I suggest we add the indexes by hand using CONCURRENT before migrating.

PR #105, opened 2020-10-14
perf(db): reduce batch size for chain history indexer

iand:
Makes batch size configurable. The high batch size was causing enormous insert queries
to be sent to postgres which were always taking > 10s to process, holding locks in the
process.

### DIFF
--- a/commands/run.go
+++ b/commands/run.go
@@ -59,7 +59,7 @@ var Run = &cli.Command{
 			Aliases: []string{"ihb"},
 			Value:   25,
 			Usage:   "Batch size for the chain history indexer",
-			EnvVars: []string{"VISOR_indexhistory_BATCH"},
+			EnvVars: []string{"VISOR_INDEXHISTORY_BATCH"},
 		},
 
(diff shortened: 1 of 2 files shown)


commands/run.go
```diff
@@ -54,6 +54,13 @@ var Run = &cli.Command{
 			Usage:   "Start indexing tipsets by walking the chain history",
 			EnvVars: []string{"VISOR_INDEXHISTORY"},
 		},
+		&cli.IntFlag{
+			Name:    "indexhistory-batch",
+			Aliases: []string{"ihb"},
+			Value:   25,
+			Usage:   "Batch size for the chain history indexer",
+			EnvVars: []string{"VISOR_indexhistory_BATCH"},
```
willscott, line 62:
```suggestion
			EnvVars: []string{"VISOR_INDEXHISTORY_BATCH"},
```

willscott (approved):

commands/run.go
iand: Thanks. That was sloppy of me!

PR #106, opened 2020-10-14
fix: use hash index type for visor_processing_actors_code_idx

iand:


willscott (approved):
there should be quite few keys on this index, so hopefully it shouldn't matter too much

PR #108, opened 2020-10-14
fix(migration): message gas economy uses bigint

frrist:
I don't think we have deployed a version of visor beyond this, given that I feel comfortable modifying the migration. If there are strong feelings against this I will modify it by adding another migration instead of altering this one.


placer14 (approved):
This is no problem. Will wait for green CI and can merge right away. Thanks!

PR #111, opened 2020-10-15
feat: convert processing, block and message tables to hypertables

iand:
This is a major rework of the database schema and may only be applied
to an empty database. It is not reversible without dropping the schema
and restating.

This change adds a height column to processing, message and actor
tables as part of the primary key and uses it
to partition the table as a timescale hypertable. Rough calculations of
expected number of rows per epoch and row widths are included to estimate
the potential chunk sizes of each hypertable. Timescale recommend that chunks
are small enough that the most recent can fit into 25% of main memory, summed
over all tables.

Additionally the indexes on the visor_processing tables have been analysed to
give the best performance for the work selection queries used by visor. The
tipsets and message processing table indexes fully cover the query so only an
index scan is performed. The actors processing table requires two indexes to
cover the cases where all codes are specified vs zero codes.

Also fixes the sector_id columns for miner_pre_commit_infos, miner_sector_infos and
miner_deal_sectors which were created as autoincrement serials by the go-pg
orm package.

### DIFF
--- a/storage/migrations/14_height_partitoning.go
+++ b/storage/migrations/14_height_partitoning.go
@@ -1,0 +1,479 @@
+package migrations
+
+import (
+	"github.com/go-pg/migrations/v8"
+)
+
+// Schema version 14 paritions most tables by height as timescale hypertables
+
+func init() {
+	up := batch(`
+
+-- Ideally we want chunks to fit into memory so aim to keep them fairly small
+-- Best practise is to restrict chunk sizes so total across all active hypertables fits 25% main memory
+-- Space partitioning can also be used (i.e. hash over another column) but main benefit comes from
+-- parallelising IO by placing those chunks on separate disks which we are not planning at the moment.
+-- Space partitioning can be added at a later date via the add_dimension function.
+
+-- TimescaleDB only supports UNIQUE constraints that have the partitioning key as their prefix. This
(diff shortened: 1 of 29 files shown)


iand:
Note that github hides the migration file by default since it is large

willscott (approved):

storage/migrations/14_height_partitoning.go
```diff
@@ -0,0 +1,465 @@
...
+);
+
+
+-- ----------------------------------------------------------------
+-- visor_processing_actors
+-- ----------------------------------------------------------------
+
+-- Make sure height is in the primary key of visor_processing_actors
+ALTER TABLE visor_processing_actors ALTER COLUMN height SET NOT NULL;
+ALTER TABLE visor_processing_actors DROP CONSTRAINT IF EXISTS IF EXISTS visor_processing_actors_pkey;
```
lanzafame, line 84:
```suggestion
ALTER TABLE visor_processing_actors DROP CONSTRAINT IF EXISTS visor_processing_actors_pkey;
```

lanzafame (changes requested):

storage/migrations/14_height_partitoning.go
iand: Must have leaned on the keyboard!

PR #113, opened 2020-10-15
feat: add deferred migrations

iand:
Deferred migrations are migrations that do not need to be applied immediately and may be
slow to execute. An example would be adding a new index to an existing large table.
A normal migration would lock the entire database until the indexing has completed which
may be unacceptably long.

All pending deferred migrations may be applied at a later date using `visor migrate --deferred`
or a up a specific schema version by using `visor migrate --deferred-to=VERSION`

Schema migration 9 is modified to use deferred migrations for the new indexes.

### DIFF
--- a/commands/migrate.go
+++ b/commands/migrate.go
@@ -20,6 +20,16 @@ var Migrate = &cli.Command{
 			Value: false,
 			Usage: "Migrate the schema to the latest version.",
 		},
+		&cli.BoolFlag{
+			Name:  "deferred",
+			Value: false,
+			Usage: "Run deferred migrations.",
+		},
+		&cli.IntFlag{
+			Name:  "deferred-to",
+			Usage: "Run deferred migrations to `VERSION`.",
+			Value: 0,
+		},
 	},
 	Action: func(cctx *cli.Context) error {
 		if err := setupLogging(cctx); err != nil {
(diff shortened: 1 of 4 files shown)


iand:
> I think the semantics here are going to be a bit painful. eventually we're going to run into cases where a subsequent schema does depend on the presence of things that we've previously marked as able to be deferred.

I agree this is a risk and we'd have to minimize use of the deferred capability, preferably just to indexing and backfills. 

> We could maybe split out indexes from table creation as separate schema paths, but I'd probably be in favor of just having a single migration path for simplicity, even at the cost of migrations potentially being slow.

I'd also prefer to have a single path. But I think we need to be have some control over when migrations are run. Currently they are run whenever we deploy a new version and lock the database for an unbounded duration.  The deferred migration idea gives us the ability to run at other times, perhaps when we have planned to have less load on the database.

For the current set of problematic indexes a better strategy may be to migrate to new tables and backfill from a particular height. That could look something like https://github.com/filecoin-project/sentinel-visor/pull/111

placer14:
Even if you could enumerate all the cases which were appropriate, there's no way to properly protect the semantics of only using deferred migrations for X and Y. I think migrations are easy enough to manage that we could simply hold a migration until we're operationally ready to deliver it (either in a PR or in a branch). Less code to manage in exchange for some human process?

iand:
The indexes in migration 9 are problematic to create against a well populated database. Holding that migration means we are blocked on deployment of schema versions 10 to 12.

The intention of deferring is that we would run the standard migration, start visor and then kick off the deferred migration. They should not be held for long periods of time. They are a mechanism to allow features such as concurrent indexing to be used (since they can't be used inside a transaction). The `--allow-schema-migrations` flag could be modified to run deferred migrations automatically once visor has started but this could block visor for some tasks such a backfilling or copying data into new tables. For those cases I think it's better to run `visor migrate --deferred` separately after an upgrade of visor.

iand:
Deferred indefinitely 😏

PR #115, opened 2020-10-16
add 30s bucket to metrics distribution as that is our block time

lanzafame:


willscott (approved):

willscott:
should we put some dense buckets right around that epoch interval as well?

lanzafame:
I don't think there is much benefit to having `20s, 25s, 30s, 35s, 50s` but am happy to add them. The overhead isn't much.

PR #116, opened 2020-10-16
feat: set default numbers of workers to zero in run subcommand

iand:
Sets the defaults for tasks to "off". The expectation should be that new tasks added won't unexpectedly start on existing deployments when Visor is upgraded.


willscott (approved):

willscott:
should check to confirm that when deploying things are still started as expected with this change / where the instruction to run new tasks is now migrated to

lanzafame (approved):

PR #119, opened 2020-10-16
include schema with messages parsed in json field

willscott:
* adds a `ParsedMessages` table including derived fields for useful message analytics.


model/messages/parsedmessage.go
```diff
@@ -0,0 +1,55 @@
...
+
+type ParsedMessage struct {
+	Cid string `pg:",pk,notnull"`
+
+	From   string `pg:",notnull"`
+	To     string `pg:",notnull"`
+	Value  string `pg:",notnull"`
+	Method string `pg:",notnull"`
+
+	Params string `pg:""`
```
iand, line 24: Params could be: 

```
Params []byte `pg:",type:jsonb,notnull"`
```

Making it []byte would save an allocation in the MessageProcessor

storage/migrations/13_parsed_message.go
```diff
@@ -0,0 +1,24 @@
...
+CREATE TABLE IF NOT EXISTS public.message_parsed (
+	"cid" text NOT NULL,
+	"from" text NOT NULL,
+    "to" text NOT NULL,
+	"value" text NOT NULL,
+	"method" text NOT NULL,
+	"params" jsonb,
+	PRIMARY KEY ("cid")
+);
+`)
```
iand, line 18: We should think about indexing the params. If the JSON is fairly flat mostly key/value then we could get a lot of mileage from a GIN index like:

```
CREATE INDEX public.message_parsed_params_idx ON message_parsed USING GIN (params);
```

A good overview of some of the options: [Using JSONB in PostgreSQL](https://scalegrid.io/blog/using-jsonb-in-postgresql-how-to-effectively-store-index-json-data-in-postgresql/#6)

tasks/message/message.go
```diff
@@ -204,9 +202,13 @@ func (p *MessageProcessor) fetchMessages(ctx context.Context, ts *types.TipSet)
...
 
-func (p *MessageProcessor) extractMessageModels(ctx context.Context, ts *types.TipSet, blkMsgs map[cid.Cid]*api.BlockMessages) (messagemodel.Messages, messagemodel.BlockMessages, visor.ProcessingMessageList, *messagemodel.MessageGasEconomy, error) {
-	msgModels := messagemodel.Messages{}
-	bmsgModels := messagemodel.BlockMessages{}
+func (p *MessageProcessor) extractMessageModels(ctx context.Context, ts *types.TipSet, blkMsgs map[cid.Cid]*api.BlockMessages) (*messagemodel.MessageTaskResult, visor.ProcessingMessageList, error) {
+	result := &messagemodel.MessageTaskResult{
+		Messages:       messagemodel.Messages{},
+		BlockMessages:  messagemodel.BlockMessages{},
+		ParsedMessages: messagemodel.ParsedMessages{},
+	}
```
iand, line 210: Initialise MessageGasEconomy too for consistency

tasks/message/message.go
```diff
@@ -266,7 +268,20 @@ func (p *MessageProcessor) extractMessageModels(ctx context.Context, ts *types.T
 				Nonce:      message.Nonce,
 				Method:     uint64(message.Method),
 				Params:     message.Params,
```
iand, line 270: Can we drop the Params column from Messages now we have ParsedMessages? Seems wasteful to store the data twice and it would make the messages row width a lot smaller.

iand (approved):
Overall LGTM with some suggestions

model/messages/parsedmessage.go
willscott: do you mean in the task? the `string()` cast shouldn't lead to a memory copy, if you do.

storage/migrations/13_parsed_message.go
willscott: It looks like a GIN index allows for fast filtering on presence of top-level keys in the json. I think we do better to put a direct index on `Method` - the keys for each message of a given method should be identical, so the filter on the GIN would be equivalent and still probably not as good as being able to filter by method, right?

tasks/message/message.go
willscott: I'm okay with that. the only downside is that it's destructive. if we have messages we've seen on previous forks that a back-fill won't re-find, we shouldn't delete that data.

model/messages/parsedmessage.go
```diff
@@ -0,0 +1,55 @@
...
+	"go.opentelemetry.io/otel/label"
+
+	"github.com/filecoin-project/sentinel-visor/metrics"
+)
+
+type ParsedMessage struct {
+	Cid string `pg:",pk,notnull"`
+
+	From   string `pg:",notnull"`
+	To     string `pg:",notnull"`
```
frrist, line 20: Would it be worth capturing the actor types for the `From` and `To` fields?

frrist (approved):

model/messages/parsedmessage.go
willscott: `Method` implicitly allows determination of the type of the `To` address.
I haven't seen any cases where we've made use of the type of `From` addresses. Can you think of a use case for that?

frrist: Off the top of my head the only case for `From` I can think of is differentiating multisig actors from account actors, but I don't know how frequently we will want that. I think this is fine as is.

model/messages/parsedmessage.go
iand: If that were true then you would be able to edit a string by modifying the byte slice's backing array.

PR #121, opened 2020-10-19
include height with parsed messages

willscott:
fixup to #119


model/messages/parsedmessage.go
```diff
@@ -16,6 +16,7 @@ import (
 type ParsedMessage struct {
 	Cid string `pg:",pk,notnull"`
 
+	Height int64  `pg:",notnull"`
```
frrist, line 19: I _think_ we'll also want `use_zero`.

frrist (approved):

iand (approved):

PR #122, opened 2020-10-19
fix index creation syntax in parsed message index

willscott:


frrist (approved):
Assuming this works for you locally, looks good to me :ship: :shipit:

willscott:
The migrate worked with this change. sorry for not doing that migration in #122

frrist:
I feel like CI should be catching this sort of thing...

PR #130, opened 2020-10-21
fix(schema): remove blocking processing indexes and improve processing stats table

iand:
The processing indexes are very slow to create against populated databases. Remove them
and add them back in as part of a later conversion of the processing tables to
hypertables (see https://github.com/filecoin-project/sentinel-visor/pull/111)

While we are amending this migration, also change the structure of the visor_processing_stats
table. Fixes https://github.com/filecoin-project/sentinel-visor/issues/120

### DIFF
--- a/tasks/stats/processing.go
+++ b/tasks/stats/processing.go
@@ -6,102 +6,104 @@ import (
 	"strings"
 	"time"
 
+	sa0builtin "github.com/filecoin-project/specs-actors/actors/builtin"
+	sa2builtin "github.com/filecoin-project/specs-actors/v2/actors/builtin"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/sentinel-visor/storage"
 	"github.com/filecoin-project/sentinel-visor/wait"
 )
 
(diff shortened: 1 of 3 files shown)


iand:
This should unblock releasing features that are held off pending this index migration.

tasks/stats/processing.go
```diff
@@ -126,7 +128,7 @@ func (r *ProcessingStatsRefresher) Run(ctx context.Context) error {
 }
 
 func (r *ProcessingStatsRefresher) collectStats(ctx context.Context) (bool, error) {
-	subQueries := []string{statsActors}
+	subQueries := []string{fmt.Sprintf(statsActors, actorCodeCase)}
```
willscott, line 131: does this mean we're templating the case statement into each of the sub-select's in the union?
is there a way to select that into a place that can then be directly used in each of the union'd selects directly?

iand: I could try something clever like inserting into a temporary table and joining against it, but I'm not sure it's worth it. The case is evaluated once per actor code per subquery (since we have already grouped by code). Perhaps about 120 times which I think is quite low compared to the cost of grouping over 30 million rows.

PR #132, opened 2020-10-21
add sql lens

willscott:
This applies the same blockstore structure that is used in https://github.com/ribasushi/ltsh/commit/b92851d074d8b1764ec063fcd836ed17840d6a00 for a visor consumer

Use with `visor --lens sql --repo "postgres:///prelotus?user=lotus&host=/var/run/postgresql/" run` or similar


lanzafame:
@willscott does this require a `ltsh` node that is synced to be running?

willscott:
> @willscott does this require a `ltsh` node that is synced to be running?

nope

lanzafame:
I am confused, how and when does that postgres db get blocks into it?

willscott:
> I am confused, how and when does that postgres db get blocks into it?

working on that in parallel (it'll eventually be a lotus PR): https://github.com/filecoin-project/lotus/tree/feat/dbrepo

willscott:
looking for reviewer. this is safe to merge, and i'd rather it not languish in a PR

lens/sqlrepo/blockstore.go
```diff
@@ -0,0 +1,235 @@
...
+	//return c.Encode(multibase.MustNewEncoder(multibase.Base64))
+
+	k, _ = multibase.Encode(multibase.Base64, c.Hash())
+	return
+}
+
+func (sbs *SqlBlockstore) Has(c cid.Cid) (has bool, err error) {
+	err = sbs.db.QueryRow(
+		context.Background(),
+		"SELECT EXISTS( SELECT 42 FROM blocks WHERE multiHash = $1 )",
```
frrist, line 81: 42?

lens/sqlrepo/blockstore.go
```diff
@@ -0,0 +1,235 @@
...
+func DB(connStr string) *pgxpool.Pool {
+	if dbpool == nil {
+		var err error
+		dbpool, err = pgxpool.Connect(context.Background(), connStr)
+		if err != nil {
+			log.Fatalf("failed to connect to %s: %s", connStr, err)
+		}
+
+		for _, ddl := range []string{
+			"CREATE TABLE IF NOT EXISTS blocks(" +
```
frrist, line 33: Assuming this means blockstore blocks and not blockchain blocks -- would it be worth using the [Large Object type](https://www.postgresql.org/docs/10/lo-intro.html) here instead? This also may be the wrong place to ask this.

frrist (approved):
Happy to have this in now and for incremental changes to land in follow ons.

lens/sqlrepo/blockstore.go
willscott: i think this is just for the where, and it's just a constant that we check if any rows are returned.

lens/sqlrepo/blockstore.go
willscott: cc @ribasushi

willscott: really i suspect that this as a consumer shouldn't be making tables at all, but it's the same blockstore object that riba is using on the lotus producer side for writing into.

lens/sqlrepo/blockstore.go
frrist: Could I bother you to add a comment mentioning that?

lens/sqlrepo/blockstore.go
ribasushi: @frrist it means blockstore-blocks-specific-to-the-blockchain: the largest block is something like 38k, definitely not a candidate for a large object

What @willscott is saying is correct re "this code should not be creating tables". Moreover: this is all highly WiP/MVP: the "schema" was for a test, it is **definitely** not how it actually should look...

If you want to incorporate this into the visor now ( the producer work is placed on hold for a few days ) - I recommend making the actual queries configurable. Then you do not care what the database looks like.

lens/sqlrepo/blockstore.go
ribasushi: We can change it to 23 if you guys *really* insist...

lens/sqlrepo/blockstore.go
frrist: Thanks for the solid explanation @ribasushi. I'm not meaning to be overly critical, just curious. Still happy to have this in as to avoid PR rot @willscott, my approval is sticky.

> it means blockstore-blocks-specific-to-the-blockchain: the largest block is something like 38k, definitely not a candidate for a large object.

Would the IPLD object representing the state tree at a specific state root also be in this store? Where is the code for this library?

ribasushi: @frrist no worries, just giving context why the parts-to-be-critiqued exist at all. I wasn't expecting us to make this into an official visor plugin when I was playing with it ;)

PR #138, opened 2020-10-23
add car repo lens

willscott:
This has the same structure as #132 but uses a carbs db as the underlying blockstore.


placer14 (approved):
LGTM

PR #139, opened 2020-10-23
chore: add issue and PR templates

frrist:


.github/pull_request_template.md
```diff
@@ -0,0 +1,5 @@
+### Motivation
```
lanzafame, line 1: Does this interfere with Github taking the long description part of a commit message and using that as the PR description?

frrist: I'm not sure, I suspect it does. From a quick look at PR's we've merged the past couple weeks it appears we don't use that feature very often.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,10 @@
+---
+name: Bug report
+about: Create a report to help us improve
+labels: kind/bug
+assignees: ''
+
+---
+### Describe the bug:
```
placer14, line 8:
```suggestion
### Visor Version: vx.x.x

### Describe the bug:
```

placer14 (approved):

.github/pull_request_template.md
iand: I'm probably the only person who writes descriptions of changes in the commit messages. As such I'd prefer to keep the current behaviour so the commit message becomes the PR text.

frrist: Alright, in that case would dropping the PR template and keeping the issue template be a reasonable compromise?

iand: Sure

PR #140, opened 2020-10-25
Fix Repo loading of Car files

willscott:
* Use a `MemRepo` rather than a `FSRepo` in read-only mode for metadata storage
* Use working [`v0.0.3`](https://github.com/willscott/carbs/releases/tag/v0.0.3) of carbs.


placer14 (approved):
GTM.

Issue #135, opened 2020-10-22
Migration from 13->14 with populated DB fails


placer14:
If I try to delete everything visor migrate --to 0
It fails with
```
running destructive schema migration from version 2 to version 1
2020-10-22T13:38:47.513-0700	FATAL	visor	sentinel-visor/main.go:116	run migration: ERROR #0A000 cannot alter type of a column used by a view or rule
```

Note: This included the following report which was separated into https://github.com/filecoin-project/sentinel-visor/issues/173 to resolution separately.

Report from @willscott :

On master, migrations are unhappy. the migration from schema 13->14 fails on a non-empty DB
```
run migration: ERROR #0A000 table "block_headers" is not empty
run migration: ERROR #23502 column "height" contains null values
```

iand:
This is working as designed. We decided adding indexes to existing data was too costly and we would start with a fresh database at this point.

iand:
Possible solutions:

 1. We could truncate data in the tables beforehand but this seems dangerous since it potentially could wipe data unexpectedly if someone redeploys with --allow-schema-migration set
 2. We could alter this schema change to create entirely new tables, leaving the old data in place

placer14:
Will break out first case into a separate issue to close the second case with the PR provided.

PR #144, opened 2020-10-26
fix: remove irreversable migration instruction

placer14:
Address 2nd case of #135


iand (approved):

iand:
I approved but it seems like the wrong fix. The error says `cannot alter type of a column used by a view or rule` so it seems we should drop that view first.

placer14:
Closing due to age. Will resolve again in new PR.

PR #149, opened 2020-10-27
feat: rpc reconnection on failure

### LINKED ISSUE #98: Visor failed to handle daemon restarts
Restarted the lotus daemon backing a visor instance and it failed to reconnect

https://gist.github.com/travisperson/4bb895ee50fa537a58049e4824bea6a0

iand:
Tasks are now passed an APIOpener which they use to acquire a lens. On
encountering a fatal error the task exits and closes the lens. The scheduler
restarts the task after a delay allowing the task to attempt a new connection
to the lens.

The lotus lens opener returns a new rpc connection. The lotusrepo and carrepo
openers return a shared instance of their repo since they hold exclusive locks
over the store.

Fixes #98

### DIFF
--- a/lens/interface.go
+++ b/lens/interface.go
@@ -1,6 +1,8 @@
 package lens
 
 import (
+	"context"
+
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/chain/vm"
@@ -14,3 +16,7 @@ type API interface {
 }
 
 type APICloser func()
+
+type APIOpener interface {
+	Open(context.Context) (API, APICloser, error)
+}
--- a/lens/lotus/lotus.go
+++ b/lens/lotus/lotus.go
@@ -6,60 +6,74 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/filecoin-project/sentinel-visor/lens"
-
+	"github.com/filecoin-project/go-jsonrpc"
+	lotus_api "github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/api/client"
+	lcli "github.com/filecoin-project/lotus/cli"
+	lru "github.com/hashicorp/golang-lru"
 	logging "github.com/ipfs/go-log/v2"
 	ma "github.com/multiformats/go-multiaddr"
 	manet "github.com/multiformats/go-multiaddr/net"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/xerrors"
 
-	"github.com/filecoin-project/go-jsonrpc"
-	lotus_api "github.com/filecoin-project/lotus/api"
-	"github.com/filecoin-project/lotus/api/client"
-	lcli "github.com/filecoin-project/lotus/cli"
+	"github.com/filecoin-project/sentinel-visor/lens"
 )
 
 var log = logging.Logger("visor/lens/lotus")
 
-func GetFullNodeAPI(cctx *cli.Context) (context.Context, lens.API, lens.APICloser, error) {
+type APIOpener struct {
+	// TODO: replace dependency on cli.Context with tokenMaddr and repo path
+	cctx  *cli.Context
+	cache *lru.ARCCache // cache shared across all instances of the api
+}
+
+func NewAPIOpener(cctx *cli.Context, cacheSize int) (context.Context, *APIOpener, lens.APICloser, error) {
+	ctx := lcli.ReqContext(cctx)
+
+	ac, err := lru.NewARC(cacheSize)
+	if err != nil {
+		return ctx, nil, nil, xerrors.Errorf("new arc cache: %w", err)
+	}
+
+	o := &APIOpener{
+		cctx:  cctx,
+		cache: ac,
+	}
+
+	return ctx, o, lens.APICloser(func() {}), nil
+}
+
+func (o *APIOpener) Open(ctx context.Context) (lens.API, lens.APICloser, error) {
 	var api lotus_api.FullNode
 	var closer jsonrpc.ClientCloser
 	var err error
 
-	if tokenMaddr := cctx.String("api"); tokenMaddr != "" {
+	if tokenMaddr := o.cctx.String("api"); tokenMaddr != "" {
 		toks := strings.Split(tokenMaddr, ":")
 		if len(toks) != 2 {
-			return nil, nil, nil, fmt.Errorf("invalid api tokens, expected <token>:<maddr>, got: %s", tokenMaddr)
+			return nil, nil, fmt.Errorf("invalid api tokens, expected <token>:<maddr>, got: %s", tokenMaddr)
 		}
 
-		api, closer, err = getFullNodeAPIUsingCredentials(cctx.Context, toks[1], toks[0])
+		api, closer, err = getFullNodeAPIUsingCredentials(ctx, toks[1], toks[0])
 		if err != nil {
-			return nil, nil, nil, xerrors.Errorf("get full node api with credentials: %w", err)
+			return nil, nil, xerrors.Errorf("get full node api with credentials: %w", err)
 		}
 	} else {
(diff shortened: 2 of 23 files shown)


lens/lotus/lotus.go
```diff
@@ -6,60 +6,74 @@ import (
...
-	lcli "github.com/filecoin-project/lotus/cli"
+	"github.com/filecoin-project/sentinel-visor/lens"
 )
 
 var log = logging.Logger("visor/lens/lotus")
 
-func GetFullNodeAPI(cctx *cli.Context) (context.Context, lens.API, lens.APICloser, error) {
+type APIOpener struct {
+	// TODO: replace dependency on cli.Context with tokenMaddr and repo path
+	cctx  *cli.Context
```
willscott, line 27: if open takes a context, why do we need a context on NewAPIOpener? we can take it for the length of opening, but i'm not seeing what the value of storing it on the opener struct is

tasks/chain/economics.go
```diff
@@ -54,11 +54,19 @@ type ChainEconomics struct {
 // Run starts processing batches of tipsets until the context is done or
 // an error occurs.
 func (p *ChainEconomics) Run(ctx context.Context) error {
+	node, closer, err := p.opener.Open(ctx)
+	if err != nil {
+		return xerrors.Errorf("open lens: %w", err)
+	}
+	defer closer()
```
willscott, line 61: how does this relate to the lifetime of the API?
if the api fails, this whole task will need to restart, right?

lens/interface.go
```diff
@@ -14,3 +16,7 @@ type API interface {
 }
 
 type APICloser func()
+
+type APIOpener interface {
```
willscott, line 20: would a pattern where there's a wrapped 'reconnecting API' that presents the lens interface, but which internally will re-open the underlying lens after a delay upon failures be an easier abstraction than this one?

having each task have to worry about the lifetime / reconnecting to the lens on failures seems like it may result in duplicate reconnecting code in more places

lens/lotus/lotus.go
iand: Are you referring to the cli Context?  It's stored because we pass it to visor which adds a tracing key and (now I look closely) sets up a signal handler. I don't know why we're doing this. I'd prefer not to return the context at all from NewAPIOpener but I was trying to limit the scope of this change to connection handling.

tasks/chain/economics.go
iand: It does. But tasks are designed to be restarted by the scheduler. When processBatch returns an error then `wait.RepeatUntil` will return and closer will be called. For the lotus api that closes the rpc client so next time the task starts (after a 1 minute delay by default) it will acquire a new connection to the lotus api. 

The repo and sql lenses manage a single reference to the repo and calling closer here is a noop.

lens/lotus/lotus.go
willscott: more that you also pass a ctx to the `Open` function, and that's the one that should probably get tagged / live for the length of the open connection rather than the one passed in constructing the opener object itself.

lens/interface.go
iand: Each task needs to be able to close and reopen its connection independently. Otherwise we have to manage concurrent access to the lens state. There's not really any specific reconnecting code in the tasks: the open a connection and close it when they exit. The scheduler ensures that tasks are restarted after they fail. The connection logic itself is encapsulated in the opener that is passed to the task.

lens/lotus/lotus.go
iand: I'm note sure I follow

NewAPIOpener accepts a cli.Context (and uses it for reading flags/settings),  Open accepts a context.Context (and uses it for calls to the lotus API)

willscott: i think it's just the disjoint between the singleton APIOpener lifetime of the structs vs the duration of connections from the `Open` call where the semantics of how the cctx on the opener object are used is non-obvious.

iand: cctx is required (unfortunately) because we have a hard dependency on passing a `*cli.Context` to `github.com/filecoin-project/lotus/cli.GetFullNodeAPI` where it expects to be able to read flags and environment to locate the repo on disk.  We could unpick that dependency.

iand: Updated so we don't return the context from the creation of api openers. Does this address your concerns @willscott

lens/lotus/lotus.go
```diff
@@ -6,60 +6,72 @@ import (
...
 
-		api, closer, err = getFullNodeAPIUsingCredentials(cctx.Context, toks[1], toks[0])
+		api, closer, err = getFullNodeAPIUsingCredentials(ctx, toks[1], toks[0])
 		if err != nil {
-			return nil, nil, nil, xerrors.Errorf("get full node api with credentials: %w", err)
+			return nil, nil, xerrors.Errorf("get full node api with credentials: %w", err)
 		}
 	} else {
-		api, closer, err = lcli.GetFullNodeAPI(cctx)
+		api, closer, err = lcli.GetFullNodeAPI(o.cctx)
```
willscott, line 61: why when we have token credentials do we use the `ctx` passed in to `Open` as the lifetime of the API, versus when we don't (and look on disk for the api endpoint) we use the context passed to create the APIOpener instead?

iand: As mentioned above they are different types. As far as I can tell GetFullNodeAPI expects the cli context so it can read the flags and environment.

iand: Perhaps @frrist or @mg could explain further since this code comes from the original spike

willscott: but in the first case you could provide `o.cctx.Context` for consistency though

iand: We should always prefer the context.Context passed to a function because it may have tracing or metrics keys in it. o.cctx.Context would be the original context derived from context.Background by the CLI package.

willscott: then lets change `GetFullNodeAPI` take in an explicit context so that one can also use the context passed in to the opener?

What i'm worried about is that the decision to pass in credentials as a flag shouldn't change the lifetime of the API as it's unrelated. that seems like it can easily lead to a fairly counter-intuitive bug to track down at some point in the future.

iand: Agree this is a wart. 

GetFullNodeAPI is in lotus. Can we decouple this change and that one?

willscott: ah. i see the issue now 😢 

we could do something like:
```
o.cctx.Context = ctx
api, closer, err = lcli.GetFullNodeAPI(o.cctx)
```

maybe would want to save/restore the full context to make sure it doesn't get lost.

iand: Can you expand on "the decision to pass in credentials as a flag shouldn't change the lifetime of the API"  - I don't think I understand why it would change it.

iand: (since the lifetime of the API is always the duration of a single run of the task that is using it)

willscott: i guess i meant more "which context is used" - if an extra tag (or a cancel context) is added to the context passed to the `Open` method, it will be applied if there is a `--api ` flag and the upper case of instantiation is used, but not if there isn't and only the original app context goes to the API.

Having consistency of the context used for the API node / json RPC seems desirable

frrist: > Perhaps frrist or mg could explain further since this code comes from the original spike

Not 100% sure on the question, but it looks like the methods called interanlly by `lcli.GetFullNodeAPI` are all exported meaning `lcli.GetFullNodeAPI` could be re-implemented here for simpler context handling.

iand:
@willscott I refactored the lotus lens to remove the dependency on lotus lcli and the long term reference to cli.Context. The opener now holds only the address and token which is uses to reconnect to the api when Open is called.

willscott (approved):

PR #157, opened 2020-10-27
Limit cases where message parsing fails

willscott:
* use a child tipset when available to learn type of destination actor for message parsing

### DIFF
--- a/tasks/message/message.go
+++ b/tasks/message/message.go
@@ -278,6 +278,8 @@ func (p *MessageProcessor) extractMessageModels(ctx context.Context, ts *types.T
 			}
 			result.Messages = append(result.Messages, msg)
 
+			msgsSeen[message.Cid()] = struct{}{}
+
 			if p.parseMessages {
 				dstAddr, err := address.NewFromString(msg.To)
 				if err != nil {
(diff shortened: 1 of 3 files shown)


tasks/message/message.go
```diff
@@ -278,13 +279,31 @@ func (p *MessageProcessor) extractMessageModels(ctx context.Context, ts *types.T
...
 				dstAddr, err := address.NewFromString(msg.To)
 				if err != nil {
 					return nil, nil, xerrors.Errorf("parse to address: %w", err)
 				}
 
-				st, err := state.LoadStateTree(p.node.Store(), ts.ParentState())
+				child, err := p.node.ChainGetTipSetByHeight(ctx, ts.Height()+1, types.NewTipSetKey())
+				if err != nil {
+					// If we aren't finalized, we fail for now, because a child tipset may occur
+					if head, err := p.node.ChainGetTipSet(ctx, types.NewTipSetKey()); err == nil && head.Height()-ts.Height() < 900 {
```
placer14, line 293:
```suggestion
					if head, err := p.node.ChainGetTipSet(ctx, types.NewTipSetKey()); err == nil && head.Height()-ts.Height() < build.Finality {
```
Prefer to avoid magic numbers. This const indirects to abi.ChainEpoch(900). https://github.com/filecoin-project/lotus/blob/master/build/params_shared_vals.go#L38

tasks/message/message.go
```diff
@@ -278,13 +279,31 @@ func (p *MessageProcessor) extractMessageModels(ctx context.Context, ts *types.T
...
+						log.Warn("Failing derivation for messate that does not yet have children, but might get them later")
```
placer14, line 294:
```suggestion
						log.Warn("Delaying derivation for message which is not yet finalized")
```
Is this error misleading? Looks like it's not because of lack of children, but more that it's still suseptible to reorg.

tasks/message/message.go
placer14: What does passing in a Zero TipSetKey do for `ChainGetTipSet`? Is that a secret way to get ChainHead? And if so, why not `api.ChainHead()`?

placer14 (approved):
Some questions but GTM otherwise.

tasks/message/message.go
willscott: they call exactly the same `getHeaviestTipset` method in the implementation.

tasks/message/message.go
```diff
@@ -278,13 +279,31 @@ func (p *MessageProcessor) extractMessageModels(ctx context.Context, ts *types.T
...
 					return nil, nil, xerrors.Errorf("parse to address: %w", err)
 				}
 
-				st, err := state.LoadStateTree(p.node.Store(), ts.ParentState())
+				child, err := p.node.ChainGetTipSetByHeight(ctx, ts.Height()+1, types.NewTipSetKey())
+				if err != nil {
+					// If we aren't finalized, we fail for now, because a child tipset may occur
+					if head, err := p.node.ChainHead(ctx); err == nil && head.Height()-ts.Height() < build.Finality {
+						log.Warn("Delaying derivation for message which is not yet finalized")
+						return nil, nil, xerrors.Errorf("Failed to load child tipset: %w", err)
```
placer14, line 295: err is `nil` here

placer14 (approved):
Logic bug in error handling. Still GTM

Issue #145, opened 2020-10-26
Track Miner Actor State Changes


frrist:
### Motivation & Background
We need to track Miner State Information. There are static and dynamic parts of the miner state:

**Static Info:**
  - MinerID
  - SectorSize
  - WindowPoStPartitionSectors

**Dynamic Info:** (excluding complex fields represented by a CID, e.g. PreCommitSecotrs, VestingFunds, Sectors, AllocatedSectors, etc.)
  - Owner, Worker, and ControlAddresses
  - PeerID, MultiAddrs
  - ConsensusFaultElapsed
  - PreCommitDeposits
  - LockedFunds
  - FeeDebt
  - InitialPledge
  - CurrentDeadline

The `Owner`, `Worker`, `Control addresses`, `PeerID`, `MultiAddrs`, and `ConsesnsuFaultsElapsed` are contained in the [MinerInfo field](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_state.go#L81) and changes to these values can be detected by inspecting the [MinerInfo CID](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_state.go#L32) for changes across state changes.

The `PreCommitDeposits`, `LockedFunds`, `FeeDebt`, `InitialPledge`, and `CurrentDeadline` fields are contained in the Miner State structure and changes to these values can be detected by inspecting the fields at the previous state.

### Acceptance Criteria
We need to persist these values when they change for analysis and tracking. Currently there are `miner_state` and `miner_info` tables that do not take the dynamic vs. static nature of the miner actor state into account, its probably simplest to remove these and start fresh but this is left to the implementer. The below table names are _suggestions_, naming is hard.

- Record miner static info to a table. `miner_static_info`
- Record dynamic fields contained in MinerInfo structure to a table along with the `StateRoot` and `MinerID` `miner_dynamic_info`
- Record the dynamic fields contained in the MinerState structure to a table along with the `StateRoot` and `MinerID` `miner_dynamic_state`

placer14:
Thoughts:
- Instead of dynamic/static as a means of describing the lifetime of the data, let's instead talk about them in terms of how they are persisted/used in the schema. So, something like `miner_current_state`/`miner_state_latest` vs `miner_states`.
[...]

PR #158, opened 2020-10-27
polish: update miner processing logic

frrist:
### Motivation

Allow the genesis task to share miner state extraction
logic with the storage miner task.

Extract miner feedebt, locked funds, miner info, and current deadline
info

Remove MinerState table, this data is already contained in the
ActorState table.

Remove MinerPower table, this logic will be moved to the power actor
task.

addresses #145


go.mod
```diff
@@ -43,6 +43,7 @@ require (
 	go.opentelemetry.io/otel/exporters/trace/jaeger v0.12.0
 	go.opentelemetry.io/otel/sdk v0.12.0
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1
+	gotest.tools v2.2.0+incompatible // indirect
```
willscott, line 46: go mod tidy?

model/actors/miner/minerinfo.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"context"
+
+	"github.com/go-pg/pg/v10"
+	"go.opentelemetry.io/otel/api/global"
+	"golang.org/x/xerrors"
+)
+
+type MinerInfo struct {
```
willscott, line 11: do we need a migration? this looks different from what i see in my current db schema

frrist: Yeah we need quite a few (:crying_cat_face:), once I get those in, in the next hour or so, I'll take this out of draft.

go.mod
frrist: Yeah, could probably make that apart of CI, along with linting... * shudders *

iand: go mod tidy shouldn't be part of ci since it makes permanent modifications to go.mod and go.sum. We should add linting though.

storage/migrations/15_miner_actor_refactor.go
```diff
@@ -0,0 +1,69 @@
...
+		"worker_change_epoch" bigint not null,
+		"consensus_faulted_elapsed" bigint not null,
+		"peer_id" text,
+		"control_addresses" jsonb,
+		"multi_addresses" jsonb,
+		PRIMARY KEY ("height", "miner_id", "state_root")
+	);
+
+	ALTER TABLE public.miner_deal_sectors RENAME TO miner_sector_deals;
+`)
```
iand, line 58: Let's make these hypertables. Add something like the following for each one:

```
-- Height chunked per 7 days (20160 epochs)
SELECT create_hypertable(
	'miner_current_deadline_infos',
	'height',
	chunk_time_interval => 20160,
	if_not_exists => TRUE
);
```

iand (approved):
LGTM but I suggest we make the new tables into hypertables so they partition by height automatically.

go.mod
frrist: I mean adding a step in CI that checks if running `go mod tidy` returns a non-empty diff e.g. what's done in specs-actors here: https://github.com/filecoin-project/specs-actors/blob/master/.circleci/config.yml#L32

iand: That would be cool

frrist: Tracked here: https://github.com/filecoin-project/sentinel-visor/issues/170

storage/migrations/15_miner_actor_refactor.go
frrist: Done.

tasks/actorstate/miner.go
```diff
@@ -31,16 +33,71 @@ func init() {
...
 
 	stop := metrics.Timer(ctx, metrics.ProcessingDuration)
 	defer stop()
 
+	ec, err := NewMinerStateExtractionContext(ctx, a, node)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO add this to the power actor processing.
```
willscott, line 48: is this in a subsequent PR?

tasks/actorstate/miner.go
```diff
@@ -56,77 +113,383 @@ func (m StorageMinerExtractor) Extract(ctx context.Context, a ActorInfo, node Ac
...
 	}
 
-	minfo, err := curState.Info()
-	if err != nil {
-		return nil, xerrors.Errorf("loading miner info: %w", err)
+	prevState := curState
+	if a.Epoch != 0 {
+		prevActor, err := node.StateGetActor(ctx, a.Address, a.ParentTipSet)
+		if err != nil {
+			return nil, xerrors.Errorf("loading previous miner actor: %w", err)
```
willscott, line 120: context of actor and epoch might be useful

willscott (approved):

tasks/actorstate/miner.go
frrist: Yes, issue for it here: https://github.com/filecoin-project/sentinel-visor/issues/171. Picking it up next.

tasks/actorstate/miner.go
frrist: agreed, will add.

PR #159, opened 2020-10-27
feat: add dynamic panel creation based on tags

lanzafame:
Panels are added to the dashboard if there are tag values for them in the visor_processing_stats.tag column.


iand (approved):
LGTM

PR #163, opened 2020-10-28
Create codeql-analysis.yml

willscott:
this is the standard github codeql analysis CI workflow. It will provide alerts on security releases of dependencies. While dependabot can be quite overwhelming, the codeql alerts in my experience are typically fairly uncommon and reasonable signal-to-noise-ratio.


frrist (approved):
Nice!

PR #165, opened 2020-10-28
fallback to address actor when interpreting messages

willscott:
fix issue occasionally observed when parsing messages.

Also enhance errors to indicate which message is source of failure.

### DIFF
--- a/tasks/message/message.go
+++ b/tasks/message/message.go
@@ -3,6 +3,7 @@ package message
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


tasks/message/message.go
```diff
@@ -3,6 +3,7 @@ package message
 import (
```
frrist, line 3: I see what happened, this is duplicated in #166. This looks fairly mechanical, I'd be fine closing this and landing everything in #166 if that's easier for you -- will leave up to you.

frrist (approved):

PR #166, opened 2020-10-28
Replace filecoin-ffi with stubs to simplify build process

willscott:
visor shouldn't be actually shelling out to ffi code to create or verify proofs.

As such, it's easier to use native golang stubs of the ffi methods, rather than needing to build the rust binary and play the cgo dance.

simplifies the path for #164

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "extern/filecoin-ffi"]
-	path = extern/filecoin-ffi
-	url = https://github.com/filecoin-project/filecoin-ffi
 [submodule "extern/fil-blst"]
 	path = extern/fil-blst
 	url = https://github.com/filecoin-project/fil-blst
--- a/tasks/message/message.go
+++ b/tasks/message/message.go
@@ -3,6 +3,7 @@ package message
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"math/big"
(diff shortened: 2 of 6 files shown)


tasks/message/message.go
```diff
@@ -284,39 +295,45 @@ func (p *MessageProcessor) extractMessageModels(ctx context.Context, ts *types.T
...
 				}
 
 				st, err := state.LoadStateTree(p.node.Store(), child.ParentState())
 				if err != nil {
-					return nil, nil, xerrors.Errorf("load state tree: %w", err)
+					return nil, nil, xerrors.Errorf("load state tree when considering message %s: %w", message.Cid().String(), err)
 				}
 
 				dstActor, err := st.GetActor(dstAddr)
+				dstActorCode := accountActorCodeID
```
frrist, line 323: :baby: nit: would prefer we check the error returned `st.GetActor(dstAddr)` on the line below it, but not a big deal

tasks/message/message.go
```diff
@@ -3,6 +3,7 @@ package message
 import (
```
frrist, line 3: did you mean to land this in the same PR?

.gitmodules
```diff
@@ -1,6 +1,3 @@
-[submodule "extern/filecoin-ffi"]
-	path = extern/filecoin-ffi
-	url = https://github.com/filecoin-project/filecoin-ffi
 [submodule "extern/fil-blst"]
```
frrist, line 1: commentary: I wonder if we can remove this too..

frrist (approved):
Good Stuff

tasks/message/message.go
willscott: i did not. this is from #165. I can rebase to remove that commit

.gitmodules
willscott: this one i don't have strong feelings about because it doesn't have the extra build step to make the go interface available.

tasks/message/message.go
willscott: fixed on this PR, and i'll merge this as a combined PR.

PR #167, opened 2020-10-28
remove line that appeared in duplicate state

willscott:


willscott:
never mind. this was my problem >.>

PR #172, opened 2020-10-29
feat: add power actor claim extraction

### LINKED ISSUE #171: Extract Power Actor Claimed Power HAMT
### Description
The power actors state maintains a map (HAMT) of miner addresses to claimed power. Extracting the state of the Claimed HAMT will yield information on the power of every miner on the network.

### Acceptance criteria
- Extract the state of the claimed power hamt to a table `power_actor_claims`.
e.g.
```go
type PowerActorClaim struct {
	Height          int64  `pg:",pk,notnull,use_zero"`
	StateRoot       string `pg:",pk,notnull"`
	MinerID         string `pg:",notnull"`
	RawBytePower    string `pg:",notnull"`
	QualityAdjPower string `pg:",notnull"`
}
```

### Where to begin
- Will require an addition to the lotus power shim for determining if the claimed hamt has changed. https://github.com/filecoin-project/lotus/pull/4628

frrist:
### What
This PR does the following:
- adds a `power_actor_claims` table to track claimed power by all miners in the power actor's state.
- updates `power_test.go` and mock api to work with power actor changes.
- shares power actor extractor logic with genesis task

closes #171
~blocked on https://github.com/filecoin-project/lotus/pull/4628~

### DIFF
--- a/tasks/actorstate/power.go
+++ b/tasks/actorstate/power.go
@@ -3,6 +3,8 @@ package actorstate
 import (
 	"context"
 
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/lotus/chain/types"
 	"go.opentelemetry.io/otel/api/global"
 	"golang.org/x/xerrors"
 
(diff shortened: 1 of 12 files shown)


tasks/actorstate/power.go
```diff
@@ -25,46 +27,102 @@ func init() {
...
+	}
+
+	curState, err := power.Load(node.Store(), curActor)
+	if err != nil {
+		return nil, xerrors.Errorf("loading current power state: %w", err)
+	}
+
+	prevState := curState
+	if a.Epoch != 0 {
+		prevActor, err := node.StateGetActor(ctx, a.Address, a.ParentTipSet)
```
willscott, line 48: has this been fixed in lotus to not be a VM run?

willscott (approved):

tasks/actorstate/power.go
frrist: :vomiting_face: ughh probably not. This isn't an issue for the API lens (since lotus has a cache) but it will be expensive in the repolens, thanks for catching this. 

Is the canonical way of dealing with this to load the state tree like [this](https://github.com/filecoin-project/sentinel-visor/blob/master/tasks/message/message.go#L331)?

willscott: for consistency that seems like a good plan

frrist: I'd like to avoid implementing this logic everywhere we call `StateGetActor` and have filed #196  to track/eventually address this concern. I propose we address this as a fix for #196, thoughts?

willscott: i'm fine with that

Issue #126, opened 2020-10-20
Reported Slow Queries


placer14:
A list of long running queries we've seen and should consider for optimization.

PRIORITY (high max and mean executions times)
```sql
WITH leased AS ( UPDATE visor_processing_actors a SET claimed_until = $1 FROM ( SELECT * FROM visor_processing_actors WHERE completed_at IS null AND (claimed_until IS null OR claimed_until < $2) AND height >= $3 AND height <= $4 AND code IN ($5,$6,$7,$8,$9) ORDER BY height DESC LIMIT $10 FOR UPDATE SKIP LOCKED ) candidates WHERE a.head = candidates.head AND a.code = candidates.code RETURNING a.head, a.code, a.nonce, a.balance, a.address, a.parent_state_root, a.tip_set, a.parent_tip_set) SELECT head, code, nonce, balance, address, parent_state_root, tip_set, parent_tip_set from leased
```

```sql
WITH leased AS ( UPDATE visor_processing_messages pm SET gas_outputs_claimed_until = '2020-10-17 21:21:35.624142746+00:00:00' FROM ( SELECT pm.cid, m.from, m.to, m.size_bytes, m.nonce, m.value, m.gas_fee_cap, m.gas_premium, m.gas_limit, m.method, r.state_root, r.exit_code,r.gas_used, bh.parent_base_fee FROM visor_processing_messages pm JOIN receipts r ON pm.cid = r.message JOIN messages m ON pm.cid = m.cid JOIN block_messages bm on pm.cid = bm.message JOIN block_headers bh on bm.block = bh.cid WHERE pm.gas_outputs_completed_at IS null AND (pm.gas_outputs_claimed_until IS null OR pm.gas_outputs_claimed_until < '2020-10-17 21:06:35.71401484+00:00:00') AND pm.height >= 0 AND pm.height <= 9223372036854775807 ORDER BY pm.height DESC LIMIT 500 ) candidates WHERE pm.cid = candidates.cid RETURNING pm.cid, candidates.* ) SELECT * FROM leased
```

```sql
INSERT INTO "visor_processing_tipsets" AS "processing_tip_set" ("tip_set", "height", "added_at", "statechange_claimed_until", "statechange_completed_at", "statechange_errors_detected", "message_claimed_until", "message_completed_at", "message_errors_detected") VALUES ($1, $2, $3, DEFAULT, DEFAULT, DEFAULT, DEFAULT, DEFAULT, DEFAULT), ...etc...
[...]

PR #178, opened 2020-10-30
feat: allow actor state processor to run without leasing

iand:
Leasing of actor states to process causes high contention on the visor_processing_actors
table since many instances are competing to lock rows to update their leases.

With this change, when --actorchange-lease is set to zero then the actor change processor
will not use leasing at all. Instead it assumes that it can work on any actor states
between the height ranges specified with --from and --to. It further subdivides that
height range by the number of workers started so each worker gets its own range of
heights to process.

Note that this requires manual coordination of height ranges allocated to each
instance of visor otherwise each instance may duplicate work performed by another
instance.

Note also that --to should be set to a reasonable value, typically under the current
height of the chain. Since the range specified with --to and --from is subdivided
evenly between workers a very large upper bound could result in some workers being
allocated ranges that have no data on the chain.

### DIFF
--- a/commands/run.go
+++ b/commands/run.go
@@ -91,7 +91,7 @@ var Run = &cli.Command{
 			Name:    "actorstate-lease",
 			Aliases: []string{"asl"},
 			Value:   time.Minute * 15,
-			Usage:   "Lease time for the actor state processor",
+			Usage:   "Lease time for the actor state processor. Set to zero to avoid using leases.",
 			EnvVars: []string{"VISOR_ACTORSTATE_LEASE"},
 		},
 		&cli.IntFlag{
@@ -120,7 +120,6 @@ var Run = &cli.Command{
 			DefaultText: "none",
 			EnvVars:     []string{"VISOR_ACTORSTATE_EXCLUDE"},
 		},
-
 		&cli.DurationFlag{
 			Name:    "message-lease",
 			Aliases: []string{"ml"},
@@ -222,6 +221,11 @@ var Run = &cli.Command{
 		// Validate flags
 		heightFrom := cctx.Int64("from")
 		heightTo := cctx.Int64("to")
+
+		if heightFrom > heightTo {
+			return xerrors.Errorf("--from must not be greater than --to")
+		}
+
 		actorCodes, err := getActorCodes(cctx)
 		if err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


commands/run.go
```diff
@@ -222,6 +221,11 @@ var Run = &cli.Command{
 		// Validate flags
 		heightFrom := cctx.Int64("from")
 		heightTo := cctx.Int64("to")
+
+		if heightFrom > heightTo {
+			return xerrors.Errorf("--from must not be greater than --to")
+		}
+
```
placer14, line 228: ❤️

commands/run.go
```diff
@@ -461,3 +496,32 @@ var actorNamesToCodes = map[string]cid.Cid{
...
+		subs[i].max = subs[i].min + step
+	}
+
+	subs[n-1].max = h.max
+	return subs
+}
+
+func estimateCurrentEpoch() int64 {
+	return int64(time.Since(time.Date(2020, 8, 24, 22, 0, 0, 0, time.UTC)) / (30 * time.Second))
+}
```
placer14, line 527: Not a fan of magic numbers.
```suggestion
var mainnetGenesis = time.Date(2020, 8, 24, 22, 0, 0, 0, time.UTC)
func estimateCurrentEpoch() int64 {
	return int64(time.Since(mainnetGenesis) / (builtin.EpochDurationSeconds))
}
```

placer14 (approved):
Nice opt-in solution!

willscott (approved):

placer14:
Applies to #126.

commands/run.go
```diff
@@ -522,6 +522,7 @@ func (h heightRange) divide(n int) []heightRange {
 	return subs
 }
 
+var mainnetGenesis = time.Date(2020, 8, 24, 22, 0, 0, 0, time.UTC)
 func estimateCurrentEpoch() int64 {
-	return int64(time.Since(time.Date(2020, 8, 24, 22, 0, 0, 0, time.UTC)) / (30 * time.Second))
+	return int64(time.Since(mainnetGenesis) / (builtin.EpochDurationSeconds))
```
tuzkii, line 527: func estimateCurrentEpoch() int64 {
	return int64(time.Since(mainnetGenesis) / (builtin.EpochDurationSeconds * time.Second))
}

iand: Can you open an issue if you think there is a problem here

PR #181, opened 2020-10-30
dep(lotus): Update to lotus@v1.1.2

placer14:


placer14:
Didn't realize current lotus was as close to master as it was. Closing as redundant.

PR #187, opened 2020-11-02
fix(rpc): use http instead of ws for rpc

lanzafame:
This allows reconnections to actually work.

I was woken too many times over the weekend because lotus had restarted and visor had failed to reconnect. I got the idea from magik's latest lotus worker PR which does the same thing.


lanzafame:
Production isn't running @iand reconnection fix... 😭

willscott:
the downside is that http probably has quite a bit more framing overhead than websockets. we should check that the performance difference in steady state is acceptable

PR #199, opened 2020-11-03
feat: add changelog generator

placer14:
Here's something we might use to refine our changelog generation for releases going forward. Intend to use this for our next release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,70 +1,154 @@
-# Changelog
-All notable changes to this project will be documented in this file.
-
-The format is a variant of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) combined
-with categories from [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)
-
-This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). Breaking changes 
-should trigger an increment to the major version. Features increment the minor version and fixes or
-other changes increment the patch number.
-
-
+<a name="unreleased"></a>
 ## [Unreleased]
 
-These are changes that will probably be included in the next release.
+### Build
+- add prometheus, grafana and dashboard images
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -76,3 +76,7 @@ dist-clean:
 	git clean -xdff
 	git submodule deinit --all -f
 .PHONY: dist-clean
+
+.PHONY: changelog
+changelog:
+	git-chglog -o CHANGELOG.md
```
iand, line 82: Can we add this binary to make deps?

CHANGELOG.md
```diff
@@ -1,70 +1,154 @@
...
-First tagged release
-
-
-
-
-
+### Build
+- add prometheus, grafana and dashboard images
+
+### Chore
```
iand, line 7: Is there a way to change the ordering so that breaking changes, features and fixes are first?

Makefile
placer14: Yeah, I'd be good with that.

CHANGELOG.md
placer14: Absolutely. The template is completely configurable.

placer14: On second look, it appears the struct for managing this will not allow custom sorting. Others appear to have wanted this feature so it's probably easy to apply here but the project appears to be abandoned. Though I do like the template-based configuration. Would we object to pulling this wierd commit in the deps and including the custom sort option as shown in https://github.com/khos2ow/git-chglog/commit/df30eead45c1d8eba3be5ed7310f0b3d8d4fc5ba?

placer14: Could we agree to handle the sorting manually right now for this release and circle back to resolve afterward?

iand: Yeah that's fine. We could also fork git-chglog and apply that sort patch.

PR #201, opened 2020-11-04
Frrist/multisig transactions 154

### LINKED ISSUE #154: Define Extractor for Multisig Actor
### Motivation & Background
To analyze and monitor token flows we need to extract multisig actor states. This will require adding a new extractor for the multistage actor type and defining new models for the extracted data.

### Acceptance criteria
Extract the following fields from the Multisig Actor state:
- Signers: a slice of addresses, model as json, dynamic value.
- NumApprovalsThreshold: number of approvals required to execute a transaction, model as uint64, dynamic value
- NextTxnID: Next ID that a transaction will be assigned to when added to the `PendingTxns` HAMT, model as uint64, dynamic value.
- InitialBalance: The token balance locked in the multisig actor, model as string,  dynamic value.
- StartEpoch: Epoch vesting begins, model as int64, dynamic value.
- UnlockDuration: Duration of vesting, model as int64, dynamic value.
- PendingTxns: HAMT mapping transactions ID's to corresponding Transactions, requires separate table,  dynamic value.

Transactions will need to be extracted from the PendingTxns HAMT, the fields of a Multisig Transaction include:
- To: address message will be sent to, model as string, static value.
- Value: value of message, model as string, static value.
- Method: method to call on `To`, model as int64, static value.
- Params: params of message, model as bytes, static value.
- Approved: Slice of addresses that have approved transaction, model as json, dynamic value

Note: The params of the Transaction could be parsed using the statediff tool (@willscott can you confirm this?)

### Where to begin
Multisig actor state: https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/multisig/multisig_state.go#L14-L29)
Lotus Interface Shim: https://github.com/filecoin-project/lotus/blob/master/chain/actors/builtin/multisig/state.go#L39

frrist:
close #154 
blocked on https://github.com/filecoin-project/lotus/pull/4719


tasks/actorstate/multisig.go
```diff
@@ -0,0 +1,157 @@
...
+				Approved:      approved,
+			})
+			return nil
+		}); err != nil {
+			return nil, err
+		}
+		return out, nil
+	}
+
+	changes, err := multisig.DiffPendingTransactions(ec.PrevState, ec.CurrState)
```
frrist, line 71: note: transactions that were removed could either have been canceled, executed and failed or executed and succeeded. Needing to spend a bit more time thinking if we can gather that info here, my gut says no..

storage/migrations/16_msig_transactions.go
```diff
@@ -0,0 +1,31 @@
...
+		"multisig_id" text not null,
+		"state_root" text not null,
+		"transaction_id" text not null,
+
+		"to" text not null,
+		"value" text not null,
+		"method" bigint not null,
+		"params" bytea not null,
+		"approved" json not null,
+		PRIMARY KEY ("height", "multisig_id", "state_root", "transaction_id")
```
willscott, line 22: do you want state_root after multisig_id here?

willscott (approved):

storage/migrations/16_msig_transactions.go
frrist: whoops, good catch

PR #203, opened 2020-11-04
Adds an option to exit on failure. (#202)

coryschwartz:
* Adds an option to exit on failure.

* Remove comment that we always want to be running


placer14:
Superceeded by https://github.com/filecoin-project/sentinel-visor/pull/209

PR #205, opened 2020-11-04
[Experiment] Buffer actor state batch completion

frrist:
motivation is to reduce writes to the visor_processing_actors table

### DIFF
--- a/model/visor/processing.go
+++ b/model/visor/processing.go
@@ -164,6 +164,13 @@ func (p *ProcessingActor) ParentTipSetKey() (types.TipSetKey, error) {
 	return TipSetKeyFromString(p.ParentTipSet)
 }
 
+func (p *ProcessingActor) MarkComplete(completedAt time.Time, errorDetected error) {
+	if errorDetected != nil {
+		p.ErrorsDetected = errorDetected.Error()
+	}
+	p.CompletedAt = completedAt
+}
+
 type ProcessingActorList []*ProcessingActor
 
 func (pl ProcessingActorList) PersistWithTx(ctx context.Context, tx *pg.Tx) error {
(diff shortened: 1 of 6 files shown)


storage/sql.go
```diff
@@ -386,8 +387,8 @@ WITH leased AS (
 	) candidates
 	WHERE a.head = candidates.head AND a.code = candidates.code
  	AND a.height >= ? AND a.height <= ?
-   RETURNING a.head, a.code, a.nonce, a.balance, a.address, a.parent_state_root, a.tip_set, a.parent_tip_set, a.height)
-SELECT head, code, nonce, balance, address, parent_state_root, tip_set, parent_tip_set, height from leased;
+   RETURNING a.head, a.code, a.nonce, a.balance, a.address, a.parent_state_root, a.tip_set, a.parent_tip_set, a.height, a.added_at)
+SELECT head, code, nonce, balance, address, parent_state_root, tip_set, parent_tip_set, height, added_at from leased;
```
frrist, line 391: Reason for this change was to return a valid model.

frrist:
unfortunately its very time consuming to provide a benchmark for this change...

model/visor/processing.go
```diff
@@ -184,6 +191,30 @@ func (pl ProcessingActorList) PersistWithTx(ctx context.Context, tx *pg.Tx) erro
 	return nil
 }
 
+func (pl ProcessingActorList) UpdateWithTx(ctx context.Context, tx *pg.Tx) error {
```
iand, line 194: There is a test in the storage package `TestMarkActorComplete`. Maybe there is a way to keep that working against this code.

model/visor/processing.go
```diff
@@ -164,6 +164,13 @@ func (p *ProcessingActor) ParentTipSetKey() (types.TipSetKey, error) {
 	return TipSetKeyFromString(p.ParentTipSet)
 }
 
+func (p *ProcessingActor) MarkComplete(completedAt time.Time, errorDetected error) {
+	if errorDetected != nil {
```
iand, line 168: I think we will want to make sure ClaimedUntil is zero. It's not strictly necessary but it keeps the data clean.

tasks/actorstate/actorstate.go
```diff
@@ -255,8 +255,15 @@ func (p *ActorStateProcessor) processBatch(ctx context.Context, node lens.API) (
 		ctx, cancel = context.WithDeadline(ctx, claimUntil)
 		defer cancel()
 	}
-
 	stats.Record(ctx, metrics.TipsetHeight.M(batch[0].Height))
+
+	var doneBatch visor.ProcessingActorList
+	defer func() {
```
iand, line 261: There may be a race here. When using leasing we set up a context with a deadline based on the lease. When that expires the remainder of the batch is abandoned and we need to make sure that anything we completed is updated in the database. However the context has been cancelled so any database updates will fail. We could probably fix that here by reordering where we create the deadline context so it's not passed into the update.

However, I think this is a wider problem in general and suspect that some of the issues we see with bad connections and advisory locks not being dropped is due to the use of cancelled contexts in the database calls.

frrist:
Will revisit this once we have benchmarking in place that can measure the effectiveness of this change.

PR #208, opened 2020-11-06
remove null bytes from message params

willscott:
this is a bit unfortunate because for some binary fields this could lead to corruption, but we probably won't be dealing with those fields meaningfully in this table.


iand (approved):
Approved but wouldn't it be better to drop the binary fields. Are they useful inside the database?

willscott:
> Approved but wouldn't it be better to drop the binary fields. Are they useful inside the database?

i think this is a problem where there are 1-2 cases of binary fields being labeled as cbor strings. we don't have the context to do this at a semantic-parsed level here, so it's hard to be selective about it

PR #209, opened 2020-11-06
hotfix to v0.3.0-rc1 restart when lock not acquired

lanzafame:
hotfix to v0.3.0-rc1


coryschwartz (approved):

schedule/scheduler.go
```diff
@@ -30,6 +30,9 @@ type TaskConfig struct {
 	// Locker is an optional lock that must be taken before the task can execute.
 	Locker Locker
 
+	// ExitOnFailure will will cause cause a fatal failure when a task stops with an error.
+	ExitOnFailure bool
```
frrist, line 34: I'm not sure we need this flag. Setting `RestartOnFailuer` to false more or less does the same thing.

schedule/scheduler.go
```diff
@@ -91,9 +94,15 @@ func (s *Scheduler) Run(ctx context.Context) error {
 				if err := tc.Locker.Lock(ctx); err != nil {
 					if errors.Is(err, storage.ErrLockNotAcquired) {
 						log.Infow("task not started: lock not acquired", "task", tc.Name)
+						if tc.ExitOnFailure {
+							log.Fatal("Exiting due to previous error. ExitOnFailure")
```
frrist, line 98: I'm not sure this is the behavior we want:
```
// Fatal uses fmt.Sprint to construct and log a message, then calls os.Exit.
```
https://github.com/uber-go/zap/blob/master/sugar.go#L126

None of the deferred unlocking code will be called if we exit here, and per my above comment I don't think the `ExitOnFailure` flag is needed.

schedule/scheduler.go
```diff
@@ -133,6 +142,9 @@ func (s *Scheduler) Run(ctx context.Context) error {
 					}
 					log.Errorw("task exited with failure", "task", tc.Name, "error", err.Error())
 
+					if tc.ExitOnFailure {
+						log.Fatal("Exiting due to previous error. ExitOnFailure")
+					}
```
frrist, line 147: ditto

willscott (approved):
only looked at changes since previous approval. looks fine.

storage/migrations/17_chainvis_views_not_materialized.go
```diff
@@ -0,0 +1,145 @@
+package migrations
```
frrist, line 1: What is the motivation for calling this before the migration previously holding this number (17)? An alternative might be:
- Leave 17 (multisig transactions) and 18 (drops unused tables) alone. 
- Add migration 19 with the desired changes from here.

frrist (commented):
I'd like to find an alternative to `ExitOnFailure`, I don't see why we need it.

We cannot merge this with the usage of Fatal logs, I believe they will break the locking.

schedule/scheduler.go
placer14: This fix is needed for the indexer when it fails on startup (as the RestartOnFailure flag is not considered until after locks are added which is exactly where the index fails when locks could not be acquired. (https://github.com/filecoin-project/sentinel-visor/blob/master/schedule/scheduler.go#L90-L106)

Agree it's a cludgy fix and would prefer to consider `RestartOnCompletion` inside the locking logic, but this is what was delivered and is working in production right now. Propose we refactor via a separate issue?

schedule/scheduler.go
iand: Strong agree. Using os.Exit almost guarantees that any lock we manage to take will be left active in the database. So in some circumstances this exit behaviour will make our problem worse.

storage/migrations/17_chainvis_views_not_materialized.go
frrist: As discussed in chat we will take these migration as is and flatten our schema https://github.com/filecoin-project/sentinel-visor/issues/225 in a follow on.

schedule/scheduler.go
frrist: As discussed in chat we will remove the `ExitOnFailure` and address the problem it is trying to fix in a follow on.

placer14: As discussed, agreed to remove this cludgy fix entirely and find a better fix for considering `RestartOnCompletion` within the locking code.

schedule/scheduler.go
placer14: As discussed, agreed to remove here.

commands/run.go
```diff
@@ -307,14 +307,12 @@ var Run = &cli.Command{
 			scheduler.Add(schedule.TaskConfig{
 				Name:                fmt.Sprintf("ActorStateChangeProcessor%03d", i),
 				Task:                actorstate.NewActorStateChangeProcessor(rctx.db, rctx.opener, cctx.Duration("statechange-lease"), cctx.Int("statechange-batch"), heightFrom, heightTo),
-				RestartOnFailure:    true,
-				RestartOnCompletion: true,
-				RestartDelay:        time.Minute,
```
frrist, line 312: Note that this change will change the restart delay to 100 milliseconds which is probably reasonable(?)

frrist (approved):
:shipit: Thanks for wrangling this @placer14 :pray:

PR #210, opened 2020-11-09
fix(migrations): CREATE EXTENSION deadlocks inside migrations global lock

### LINKED ISSUE #179: Remove create extension from migration
### Describe the bug:
Migrations require a global lock which deadlocks when migration 1 is executed due to `create extension` statement which also requires a global lock. This statement should be removed from migrations and included in docs for manual application.

### Steps to Reproduce:
1. `./visor migrate --latest` on DB without timescaledb extension already enabled.

hsanjuan:
Instead, run this statement before obtaining the global lock for migrations.

Fixes #179 .

### DIFF
--- a/storage/migrate.go
+++ b/storage/migrate.go
@@ -95,9 +95,12 @@ func (d *Database) MigrateSchemaTo(ctx context.Context, target int) error {
 	}
 
 	// CREATE EXTENSION cannot run inside a transaction (as part of
-	// migration), so we run it explicitally here. (#179).
-	if _, err := d.DB.Exec(`CREATE EXTENSION IF NOT EXISTS timescaledb WITH SCHEMA public`); err != nil {
-		return xerrors.Errorf("creating timescaledb extension: %w", err)
+	// migration), so we run it explicitally here and only when setting
+	// things up for the first time (#179).
+	if dbVersion == 0 {
+		if _, err := d.DB.Exec(`CREATE EXTENSION IF NOT EXISTS timescaledb WITH SCHEMA public`); err != nil {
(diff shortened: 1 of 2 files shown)


hsanjuan:
#179 says that this should documented to be run manually, but I find that suboptimal vs just doing it for the user (doing it when it's not needed does not seem like a bad thing either). If I'm missing something, let me know.

storage/migrate.go
```diff
@@ -89,6 +89,12 @@ func (d *Database) MigrateSchemaTo(ctx context.Context, target int) error {
 		return xerrors.Errorf("check migration sequence: %w", err)
 	}
 
+	// CREATE EXTENSION requires a global lock so it cannot be run inside
```
iand, line 92: Let's do this after we have obtained the exclusive lock below. The lock is advisory and simply used to ensure that there is only one copy of visor using the database. It won't actually cause postgres to lock anything. I believe the restriction on the extension is that it can't be run inside a transaction.

storage/migrate.go
```diff
@@ -94,6 +94,12 @@ func (d *Database) MigrateSchemaTo(ctx context.Context, target int) error {
 		return xerrors.Errorf("acquiring schema lock: %w", err)
 	}
 
+	// CREATE EXTENSION cannot run inside a transaction (as part of
```
iand, line 97: We only need to do this when we are at schema version 0 (when dbVersion = 0)

hsanjuan: Sorry for the delay, added that now.

iand (approved):

PR #213, opened 2020-11-10
fix(scheduler): Add LockerFailureSchedulerAbort

### LINKED ISSUE #207: "Lock not acquired" should be fatal
When the `ChainHeadIndexer` hits the case of [not acquring its lock](https://github.com/filecoin-project/sentinel-visor/issues/51), the task should cause visor to fatally die instead of accept that it cannot start without further action.

hsanjuan:
Fixes #207. Adds a task property that can abort the full scheduler
when the locking fails. This property is enabled for ChainHeadIndexer
tasks.

### DIFF
--- a/commands/run.go
+++ b/commands/run.go
@@ -263,12 +263,13 @@ var Run = &cli.Command{
 		// Add one indexing task to follow the chain head
 		if cctx.Bool("indexhead") {
 			scheduler.Add(schedule.TaskConfig{
-				Name:                "ChainHeadIndexer",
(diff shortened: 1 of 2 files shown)


hsanjuan:
Re-reading #207, it may be this was not what it asked.

placer14:
This is a bit of a hack which we wanted last week to allow self-healing to alleviate the constant alerts. This may not be warranted as strongly now that Sentinel lotus nodes have more memory to operate and visor has less-frequent restart events. Looking at the recent logs from visor, we have not encounted the "lock not acquired" problem since we cut the v0.30-rc. Apologies to have you chase this and not use it, but might be better to root cause the stuck locks (should we start hitting them again) than rely on this hack. Closing this and the original issue.

PR #214, opened 2020-11-10
feat(lens): Optimize StateGetActor calls.

hsanjuan:
Add a utility function called "OptimizedStateGetActorWithFallback()" which
gets the current tipset's state by trying to load the child's state and using
ParentState() on that, so that we avoid recomputing the state for the current
tipset by applying messages.

If anything goes wrong, the function falls back to the original, more
expensive StateGetActor, and logs a warning.


frrist (approved):
LGTM. I'd recommend giving this a read for more context on the nuance of the lotus API methods: https://protocollabs.slack.com/archives/CKT1X7YFN/p1603386966321800

lens/util.go
```diff
@@ -0,0 +1,65 @@
...
+	return act, nil
+}
+
+func efficientStateGetActor(ctx context.Context, api API, actor address.Address, tsk types.TipSetKey) (*types.Actor, error) {
+	ts, err := api.ChainGetTipSet(ctx, tsk)
+	if err != nil {
+		return nil, xerrors.Errorf("Failed to load tipset: %w", err)
+	}
+
+	child, err := api.ChainGetTipSetByHeight(ctx, ts.Height()+1, types.NewTipSetKey())
```
frrist, line 39: I think this is missing handling for null rounds, I believe `child == ts` will be true for that case.

hsanjuan: looking

frrist: I've been able to manually confirm it (that is `child == ts` when there is a null round after `ts) on my local deployment. I think the lotus chain store has a method for handling this case gracefully: https://github.com/filecoin-project/lotus/blob/master/chain/store/store.go#L1261

frrist:
@hsanjuan I have cherry-picked and tweaked these changes to handle null rounds in #238 -- see https://github.com/filecoin-project/sentinel-visor/commit/defb12a515f05e1efc5885005fdc9742c6fad4b7#diff-a771f3c826cfe5ebe05c1a003873af47456ee3e16188c1e5d0a1162c0b3e27acR41. 

In the same PR I have fixed what I believe to be an off by one error (see https://github.com/filecoin-project/sentinel-visor/pull/238#discussion_r525552778) I discovered while reviewing this. Without the off-by-one fix many of the calls to StateGetActor will continue to fail which is why I (sloppy) did this all in a single PR.

Feel free to merge the changes in defb12a515f05e1efc5885005fdc9742c6fad4b7 here and I will land the bug fix as a follow on, or just land #238 directly in your AM.

lens/util.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("load child tipset: %w", err)
+	}
+
+	if !cidsEqual(child.Parents().Cids(), ts.Cids()) {
```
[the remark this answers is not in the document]
hsanjuan: I think I copy-pasted from somewhere. Looking at https://github.com/filecoin-project/lotus/blob/master/chain/types/tipset.go#L238 it seems that lotus does not assume that cids will be in the same order and takes even more troubles to make the check... I am not sure it is a safe assumption to make that the TipSetKeys will serialize to the same thing though (maybe?).

lens/util.go
hsanjuan: I think this is resolved by the change you added right?

hsanjuan:
@frrist I have imported your changes and added some smaller fixes.

lens/util.go
frrist: This is resolved by using `GetTipsetByHeight` and passing `false` in, so yes I believe we can resolve this comment.

lens/util.go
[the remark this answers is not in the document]
frrist: Given the following [definition](http://beta.spec.filecoin.io.ipns.localhost:8080/#systems__filecoin_blockchain__struct__tipset) I'd like to think we can depend on it, but it's probably safer to be explicit here, as you are doing now.

frrist:
My approval is sticky, this LGTM.

frrist:
Hey @hsanjuan is there anything that is preventing you from merging this?

hsanjuan:
I don't remember. Either I forgot or there were discussions about not calling these methods at all and this felt like more crust. Should we merge (I can rebase), or close?

frrist:
`StateGetActor` is still used through out Visor during actor extraction, and as far as I am aware Lotus will still recompute state when the method is called -- meaning this is still worth merging since it will alleviate the performance hit of state computation for some cases.

lens/util.go
```diff
@@ -0,0 +1,58 @@
...
+
+var logger = logging.Logger("visor/lens/lotus")
+
+// OptimizedStateGetActorWithFallback is a helper to obtain an actor in the
+// state of the current tipset without recomputing the full tipset. It does
+// this by obtaining the child tipset (current height+1) and using the
+// pre-computed ParentState().
+//
+// TODO: Remove. See:  https://github.com/filecoin-project/sentinel-visor/issues/196
+func OptimizedStateGetActorWithFallback(ctx context.Context, store cbor.IpldStore, chainAPI full.ChainModuleAPI, fallback full.StateModuleAPI, actor address.Address, tsk types.TipSetKey) (*types.Actor, error) {
```
hsanjuan, line 24: This could be simpler if `FullNode` actually implemented `FullNodeAPI` but it doesn't:

```
lens/util/repo.go:109:74: cannot use ra.FullNodeAPI (type impl.FullNodeAPI) as type api.FullNode in argument to lens.OptimizedStateGetActorWithFallback:
	impl.FullNodeAPI does not implement api.FullNode (AuthNew method has pointer receiver)
```

hsanjuan (commented):
@frrist ok, this should be good to go.

Please review again. The problem was that there was a TODO because I did not manage to enable the optimization for the lotus-api lens as I did not find how to access a `store`. I somehow managed now.

PR #219, opened 2020-11-12
feat: add views to see the distribution of data in visor tables

iand:
Adds a new schema `visorstats` that contains views for each table visor populates showing the distribution of data in each table in groups of 10,000 epochs. Where the underlying table is a hypertable the view is created as a continuous agg.

Example:

```
> select * from visorstats.market_deal_states;
 lower  | upper  |  count  
--------+--------+---------
 170000 | 178800 | 1307486
 160000 | 169999 |  474772
 150000 | 159999 |  481812
 140000 | 149760 |  503161
 130000 | 138240 |  236842
 120000 | 126720 |  179243
 110000 | 115200 |  187132
      0 |      0 |    2304
(8 rows)
```

Renumbered the migration to 19 so it comes after https://github.com/filecoin-project/sentinel-visor/pull/240

### DIFF
--- a/storage/migrations/19_visorstats.go
+++ b/storage/migrations/19_visorstats.go
@@ -1,0 +1,220 @@
+package migrations
+
+import (
+	"github.com/go-pg/migrations/v8"
+)
+
+// Schema version 19 adds the visorstats scheme with views that show the progress of visor over ranges of epochs
+
+func init() {
+	up := batch(`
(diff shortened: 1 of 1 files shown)


iand:
These views are not that great and the continuous aggregates interfere with some bulk loading activities. Will return to them if needed.

PR #222, opened 2020-11-12
Add s3 lens for working against s3 data

willscott:
cc @ribasushi


iand (commented):
Can you make this more generic. The only part that depends on S3 is the S3BlockStore, so can you split out a fake full node constructor that takes a blockstore.Blockstore implementation. Then we can avoid reimplementing the fake verifier and other node logic.

lens/s3repo/store.go
```diff
@@ -0,0 +1,116 @@
...
+		return -1, err
+	}
+	if resp.StatusCode == 200 {
+		return int(resp.ContentLength), nil
+	}
+	return -1, fmt.Errorf("does not exist")
+}
+
+func (sbs *S3Blockstore) getMasterTsKey(ctx context.Context, lookback int) (*types.TipSetKey, error) {
+	resp, err := sbs.client.Get(sbs.prefix + "/head")
```
iand, line 73: This didn't work when I tried fetching the url manually

lens/s3repo/repo.go
```diff
@@ -0,0 +1,217 @@
...
+	rapi.Context = c.Context
+	rapi.cacheSize = c.Int("lens-cache-hint")
+	return &APIOpener{rapi: &rapi}, sf, nil
+}
+
+func (o *APIOpener) Open(ctx context.Context) (lens.API, lens.APICloser, error) {
+	return o.rapi, lens.APICloser(func() {}), nil
+}
+
+type S3API struct {
```
iand, line 98: Will this support ChainGetTipSetByHeight?

willscott: yep.
The current metadata setup either will happen via the lotus metadata in-memory store which will keep an index to answer this.

Here's a janky re-implementation of that which may be the path of least-resistance here too:
https://github.com/filecoin-project/statediff/blob/master/lib/lazysource.go#L63

lens/s3repo/store.go
ribasushi: @iand note that the s3 bucket is frozen at 230700
more this weekend

lens/s3repo/store.go
```diff
@@ -0,0 +1,116 @@
...
+	resp, err := sbs.client.Get(sbs.prefix + c.String() + "/data.raw")
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	buf, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	return blocks.NewBlockWithCid(buf, c)
```
iand, line 58: Check resp.StatusCode

willscott:
pulled out the common repo code. ready for another look.

frrist (approved):

PR #228, opened 2020-11-16
fix: Make ChainVis into basic views

placer14:
Make existing materialized views for ChainVis project into basic views.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,6 @@
 version: 2.1
 
 executors:
-  golang:
-    docker:
-      - image: cimg/go:1.15.2
   dockerizer:
     docker:
       - image: cimg/go:1.15.2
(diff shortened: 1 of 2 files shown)


storage/migrations/17_chainvis_views_not_materialized.go
```diff
@@ -0,0 +1,145 @@
...
+		main_block.miner,
+		main_block.height,
+		main_block.parent_weight AS parentweight,
+		main_block.timestamp,
+		main_block.parent_state_root AS parentstateroot,
+		parent_block.timestamp AS parenttimestamp,
+		parent_block.height AS parentheight,
+		mp.raw_bytes_power AS parentpower,
+		synced.synced_at AS syncedtimestamp,
+		(SELECT COUNT(*) FROM block_messages WHERE block_messages.block = main_block.cid) AS messages
```
willscott, line 29: name this column 'message_count`?

storage/migrations/17_chainvis_views_not_materialized.go
```diff
@@ -0,0 +1,145 @@
...
+	SELECT
+		main_block.cid AS block,
+		bp.parent AS parent,
+		main_block.miner,
+		main_block.height,
+		main_block.parent_weight AS parentweight,
+		main_block.timestamp,
+		main_block.parent_state_root AS parentstateroot,
+		parent_block.timestamp AS parenttimestamp,
+		parent_block.height AS parentheight,
```
willscott, line 26: name as `parent_height` rather than `parentheight`?

willscott (approved):

storage/migrations/17_chainvis_views_not_materialized.go
willscott: this is an instance of more general: "prefer '_' or camel case over... just pushing the words together"

.circleci/config.yml
```diff
@@ -1,9 +1,6 @@
 version: 2.1
```
frrist, line 1: This is unrelated to the ChainVis code, yeah?

placer14: Correct, these changes only enable the migration test within Circle CI.

storage/migrations/17_chainvis_views_not_materialized.go
placer14: These views go to an externally supported app. Changing column names will require updates from the devs (something I wanted to avoid at the moment to get this change into production for them quickly).

PR #229, opened 2020-11-16
fix: don't update go modules when running make

frrist:
I'd like to avoid fetching `git-chnglg` when I run make (also the `-u` in the `go get` was cauing go modules to be updated on calls to `make`). Instead `make changelog` runs the binary using:
```
go run go run github.com/git-chglog/git-chglog/cmd/git-chglog ...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,6 @@ build/.update-modules:
 
 .PHONY: deps
 deps: build/.update-modules
-	go get -u github.com/git-chglog/git-chglog/cmd/git-chglog
 
 # test starts dependencies and runs all tests
 .PHONY: test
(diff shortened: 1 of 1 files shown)


willscott (approved):

placer14:
I don't feel strongly about this change, but it was recommended by @iand here: https://github.com/filecoin-project/sentinel-visor/pull/199#discussion_r517007276. Might be worth to get his opinion before deciding one way or the other. 

Three ideas:
1) maybe we'll just include a line in the documentation about it, let the dev manually resolve it.
2) Or maybe make the `go get` a part of `make changelog` if the binary isn't already installed? 
3) Or another idea, we change to GOPATH mode when we install it so it doesn't update modules with the dep on installation?

willscott:
we don't need (and i also prefer not to) `go get` external tools, but could directly invoke them as `go run github.com/...` in tooling. I don't keep go binaries on my system path normally, so the `go get` path would break for me anyway.

placer14:
Maybe `make changelog` should `go run` instead of calling it locally? That'd be fine with me, too.

frrist:
`go run github.com/...` sounds good to me, let's do that.

PR #233, opened 2020-11-17
fix: gracefully disconnect from postgres on exit

iand:
Extracted from iand/beefy branch where I have been using this for a while. Ensures the context used to run queries is cancelled properly when the process exits.


willscott (approved):

frrist (approved):
Nice!

PR #238, opened 2020-11-17
Fix off by one in actor state change processor

frrist:
While reviewing #214 I found what I believe to be an off-by-one error in the actor state change logic. This is a bit complicated so I'd love a second set of eyes on this to confirm/deny my suspicion.

### DIFF
--- a/lens/lotus/api.go
+++ b/lens/lotus/api.go
@@ -131,7 +131,10 @@ func (aw *APIWrapper) StateGetActor(ctx context.Context, actor address.Address, 
 	ctx, _ = tag.New(ctx, tag.Upsert(metrics.API, "StateGetActor"))
 	stop := metrics.Timer(ctx, metrics.LensRequestDuration)
 	defer stop()
-	return aw.FullNode.StateGetActor(ctx, actor, tsk)
+
+	return aw.StateGetActor(ctx, actor, tsk)
+	// TODO idk how to get a chain store here, probably just need to change the lotus api?
+	//return lens.OptimizedStateGetActorWithFallback(ctx, aw, aw.FullNode, actor, tsk)
 }
 
 func (aw *APIWrapper) StateListActors(ctx context.Context, tsk types.TipSetKey) ([]address.Address, error) {
--- a/lens/util.go
+++ b/lens/util.go
@@ -1,0 +1,67 @@
+package lens
+
+import (
+	"context"
+	"errors"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/lotus/chain/state"
+	"github.com/filecoin-project/lotus/chain/store"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/node/impl/full"
+	"github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log/v2"
+	"golang.org/x/xerrors"
+)
+
+var logger = logging.Logger("visor/lens/lotus")
+
+// OptimizedStateGetActorWithFallback is a helper to obtain an actor in the
+// state of the current tipset without recomputing the full tipset. It does
+// this by obtaining the child tipset (current height+1) and using the
+// pre-computed ParentState().
+//
+// TODO: Remove. See:  https://github.com/filecoin-project/sentinel-visor/issues/196
+func OptimizedStateGetActorWithFallback(ctx context.Context, store *store.ChainStore, fallback full.StateModuleAPI, actor address.Address, tsk types.TipSetKey) (*types.Actor, error) {
+	act, err := efficientStateGetActorFromChainStore(ctx, store, actor, tsk)
+	if err != nil {
+		logger.Warnf("Optimized StateGetActorError: %s. Falling back to default StateGetActor().")
+		return fallback.StateGetActor(ctx, actor, tsk)
+	}
+	return act, nil
+}
+
+func efficientStateGetActorFromChainStore(ctx context.Context, store *store.ChainStore, actor address.Address, tsk types.TipSetKey) (*types.Actor, error) {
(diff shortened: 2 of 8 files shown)


lens/util.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func efficientStateGetActorFromChainStore(ctx context.Context, store *store.ChainStore, actor address.Address, tsk types.TipSetKey) (*types.Actor, error) {
+	ts, err := store.GetTipSetFromKey(tsk)
+	if err != nil {
+		return nil, xerrors.Errorf("Failed to load tipset: %w", err)
+	}
+
+	// heaviest tipset means look on the main chain and false means return tipset following null round.
+	child, err := store.GetTipsetByHeight(ctx, ts.Height()+1, store.GetHeaviestTipSet(), false)
```
frrist, line 41: Riffing on https://github.com/filecoin-project/sentinel-visor/pull/214. This change in logic handles null rounds.

tasks/actorstate/actorstatechange.go
```diff
@@ -182,35 +181,15 @@ func (p *ActorStateChangeProcessor) processTipSet(ctx context.Context, node lens
...
 			Code:            act.Code.String(),
 			Nonce:           strconv.FormatUint(act.Nonce, 10),
 			Balance:         act.Balance.String(),
 			Address:         addr.String(),
-			TipSet:          pts.Key().String(),
-			ParentTipSet:    pts.Parents().String(),
-			ParentStateRoot: pts.ParentState().String(),
+			TipSet:          ts.Key().String(),
+			ParentTipSet:    pts.Key().String(),
+			ParentStateRoot: ts.ParentState().String(),
```
frrist, line 192: I believe we have an off-by-one error here:
The actors state changed between `pts` (parent) and `ts` (child) and the stateroot containing the change is the childs stateroot, which is equal to the evaluation of the messages in the parent tipset.

tasks/actorstate/actorstatechange.go
```diff
@@ -182,35 +181,15 @@ func (p *ActorStateChangeProcessor) processTipSet(ctx context.Context, node lens
...
-		_, err = node.StateGetActor(ctx, addr, pts.Parents())
-		if err != nil {
-			if strings.Contains(err.Error(), "actor not found") {
-				ll.Debugw("parent actor not found", "addr", str)
-				// TODO consider tracking deleted actors
-				continue
-			}
-			return xerrors.Errorf("get actor parent: %w", err)
-		}
-
```
frrist, line 204: This lookup will always fail for the case when a message that creates an actor exists in `pts` -- meaning the actor state exists in `ts`'s stateroot -- since it contains the result of message evaluation from pts.

frrist: I _think_ we are not always guaranteed to see an `actor not found` error here since some lens's will run the VM and the VM may not return `actor not found`. When this happens we don't persist any actor state changes for a tipset, meaning we will be missing data.

frrist: I have removed these calls because they are no longer necessary give the below change to model data.

lens/lotus/api.go
```diff
@@ -131,7 +131,10 @@ func (aw *APIWrapper) StateGetActor(ctx context.Context, actor address.Address,
 	ctx, _ = tag.New(ctx, tag.Upsert(metrics.API, "StateGetActor"))
 	stop := metrics.Timer(ctx, metrics.LensRequestDuration)
 	defer stop()
-	return aw.FullNode.StateGetActor(ctx, actor, tsk)
+
+	return aw.StateGetActor(ctx, actor, tsk)
+	// TODO idk how to get a chain store here, probably just need to change the lotus api?
```
willscott, line 136: what do you need for chain store that isn't in `aw.Store()`?

lens/util.go
```diff
@@ -0,0 +1,67 @@
...
+func OptimizedStateGetActorWithFallback(ctx context.Context, store *store.ChainStore, fallback full.StateModuleAPI, actor address.Address, tsk types.TipSetKey) (*types.Actor, error) {
+	act, err := efficientStateGetActorFromChainStore(ctx, store, actor, tsk)
+	if err != nil {
+		logger.Warnf("Optimized StateGetActorError: %s. Falling back to default StateGetActor().")
+		return fallback.StateGetActor(ctx, actor, tsk)
+	}
+	return act, nil
+}
+
+func efficientStateGetActorFromChainStore(ctx context.Context, store *store.ChainStore, actor address.Address, tsk types.TipSetKey) (*types.Actor, error) {
```
willscott, line 34: It's unfortunate that this is going to end up going across the API boundary several times (load tipset, load child tipset, loadstatetree (expensive)).
we should think about if there's a way to cache some or any of these things before the API boundary.

iand:
There was a slack discussion about this last month. 

What is the impact of this change? Do we simply have the wrong height associated with the stateroots? Do we need to reprocess all actor states?

tasks/actorstate/actorstatechange.go
hsanjuan: I don't think this is right. We are looking at changes between pts.ParentState() (state corresponding to ts.Height()-2) and ts.ParentState() (state corresponding to ts.Height()-1).

The ProcessingActor is getting recorded to process an actor that lives in ts.ParentState(). When processing the actor, the State is obtained with `StateReadState(address, tipset)` where tipset is `pts.Key()`. That is correct, since that corresponds to ts.Height()-1 which was the place where the new values for that actor took place.

At least that is how I understand it, even though this might be unefficient and Lotus has this full mess of what the "state for a tipset" means.

frrist:
Per discussion in Sentinel Office hours we will be closing this to remain consistent with the lotus API. We will revisit when https://github.com/filecoin-project/lotus/pull/4545 lands.

PR #240, opened 2020-11-18
chore: remove unused tables and views

### LINKED ISSUE #217: Remove unused tables from sentinel-analysis database
### Description

There are a bunch of old tables that are not populated or used by visor and are confusing for users (examples: `miner_info` vs `miner_infos`, `miner_powers` vs `power_actor_claims`) We should remove them. Drone doesn't run against this database so the only tables in `public` are the ones created by visor.

### Acceptance criteria

- Only tables that are used by visor should be in the `public` schema
- Analysis related views and tables are in the `analysis` schema
- Remove tables which don't belong in either place.

### Where to begin

 - Cross reference the schema used by Visor with the existing database and create a migration to remove unused tables and views.

frrist:
closes #217


storage/migrations/18_remove_unused_tables.go
```diff
@@ -0,0 +1,36 @@
...
+	DROP TABLE IF EXISTS public.miner_power; 				-- replaced by power_actor_claims
+	DROP TABLE IF EXISTS public.miner_powers; 				-- replaced by power_actor_claims
+	DROP TABLE IF EXISTS public.minerid_dealid_sectorid; 	-- replaced by miner_sector_deals
+	DROP TABLE IF EXISTS public.sector_info; 				-- replaced by miner_sector_infos
+	DROP TABLE IF EXISTS public.sector_precommit_infos; 	-- replaced by miner_pre_commit_infos
+
+`)
+
+	// no take backsies.
+	down := batch(`SELECT 1;`)
```
frrist, line 33: I'd really like to avoid writing the down SQL for this if we think its reasonable. Almost all of the tables have constants, these can be seen in migration 1. Adding back the required constants here will be very tedious and adds almost zero value since we never intend to use these tables again, thoughts?

iand: Means we can't roll back a failed deployment. We have some other migrations like that. I'm ok with it since any failure state can be corrected by adding another forward migration.

iand (approved):

storage/migrations/18_remove_unused_tables.go
placer14: Not a fan. :( Hope we don't have to deal with this later.

iand: To clarify what I meant. If you have migration `n` that creates a table which you need to undo, then you can create migration `n+1` to remove it. This way we can keep migrations immutable and everyone who has migrated to `n` has a path forward.

PR #248, opened 2020-11-18
Reorder migrations after merging latest master

placer14:
Just want to be explicit about the changes I'm making to make latest hotfix migrations agree with what is already deployed to `master`.

NOTE: This will FBUAR anyone's visor database who has already migrated up to v17.

### DIFF
--- a/storage/migrations/19_remove_unused_tables.go
+++ b/storage/migrations/19_remove_unused_tables.go
@@ -8,6 +8,34 @@ import (
 
 func init() {
 	up := batch(`
+DROP VIEW IF EXISTS chain_visualizer_chain_data_view;
+CREATE VIEW chain_visualizer_chain_data_view AS
+	SELECT
+		main_block.cid AS block,
+		bp.parent AS parent,
+		main_block.miner,
+		main_block.height,
+		main_block.parent_weight AS parentweight,
+		main_block.timestamp,
+		main_block.parent_state_root AS parentstateroot,
+		parent_block.timestamp AS parenttimestamp,
+		parent_block.height AS parentheight,
+		-- was miner_power.raw_bytes_power (plural bytes)
+		pac.raw_byte_power AS parentpower,
+		-- was blocks_synced.synced_at
+		main_block.timestamp AS syncedtimestamp,
+		(SELECT COUNT(*) FROM block_messages WHERE block_messages.block = main_block.cid) AS messages
+	FROM
+		block_headers main_block
+	LEFT JOIN
+		block_parents bp ON bp.block = main_block.cid
+	LEFT JOIN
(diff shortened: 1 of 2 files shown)


storage/migrations/19_remove_unused_tables.go
```diff
@@ -4,11 +4,10 @@ import (
 	"github.com/go-pg/migrations/v8"
 )
 
-// Schema version 18 drops unused tables.
+// Schema version 19 drops unused tables.
 
 func init() {
 	up := batch(`
-	DROP MATERIALIZED VIEW IF EXISTS public.chain_visualizer_chain_data_view; 		-- artifact from chainwatch
```
placer14, line 11: This view is being used by an external project and is removed in the hotfix's v17 migration. Removing this to prevent an error here.

frrist: This view depends on some of the tables being dropped below:
https://github.com/filecoin-project/sentinel-visor/blob/master/storage/migrations/7_chainvis_views.go#L35
https://github.com/filecoin-project/sentinel-visor/blob/master/storage/migrations/7_chainvis_views.go#L37

frrist (commented):
I don't intend to block this change from landing in the release branch, but I don't see how this will make it into master, let's discuss when we reach that point.

storage/migrations/19_remove_unused_tables.go
placer14: The view supports [this tool](https://github.com/DigitalMOB2/filecoin-lotus-explorer-playground/). If we are dubious about its value, I'd like to talk that through first, but maybe not as part of this PR.

Agree that the view needs to be corrected for the dropped tables. Will fix that.

frrist: An alternative to fixing it here could be updating the view here: https://github.com/DigitalMOB2/filecoin-lotus-explorer-playground/blob/master/2.create-views.sql#L1. But happy to go with either option.

storage/migrations/19_remove_unused_tables.go
```diff
@@ -4,11 +4,39 @@ import (
...
+		synced.synced_at AS syncedtimestamp,
+		(SELECT COUNT(*) FROM block_messages WHERE block_messages.block = main_block.cid) AS messages
+	FROM
+		block_headers main_block
+	LEFT JOIN
+		block_parents bp ON bp.block = main_block.cid
+	LEFT JOIN
+		block_headers parent_block ON parent_block.cid = bp.parent
+	LEFT JOIN
+		blocks_synced synced ON synced.cid = main_block.cid
```
frrist, line 34: I believe this table has been removed.

PR #252, opened 2020-11-19
feat: functions to convert unix epoch to fil epoch

olizilla:
adds `unix_to_height` and `height_and_unix` to translate a unix epoch in seconds to a fil epoch. Makes searching the messages tables more fun.

This is mainnet specific, so this may not be the right approach. I'd need to be able to programmatically set the initial unix epoch time to suport other networks.

License: MIT

### DIFF
--- a/storage/migrations/19_height_functions.go
+++ b/storage/migrations/19_height_functions.go
@@ -8,13 +8,13 @@ import (
 
 func init() {
 	up := batch(`
-CREATE OR REPLACE FUNCTION public.unix_to_height(unix_epoch bigint) RETURNS bigint AS $$
-	SELECT ((unix_epoch - 1598306400) / 30)::bigint;
-$$ LANGUAGE SQL;
+	CREATE OR REPLACE FUNCTION public.unix_to_height(unix_epoch bigint) RETURNS bigint AS $$
+		SELECT ((unix_epoch - 1598306400) / 30)::bigint;
+	$$ LANGUAGE SQL;
 
-CREATE OR REPLACE FUNCTION public.height_to_unix(fil_epoch bigint) RETURNS bigint AS $$
-	SELECT ((fil_epoch * 30) + 1598306400)::bigint;
-$$ LANGUAGE SQL;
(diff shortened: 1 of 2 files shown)


storage/migrations/19_height_functions.go
```diff
@@ -0,0 +1,25 @@
...
+	SELECT ((unix_epoch - 1598306400) / 30)::bigint;
+$$ LANGUAGE SQL;
+
+CREATE OR REPLACE FUNCTION public.height_to_unix(fil_epoch bigint) RETURNS bigint AS $$
+	SELECT ((fil_epoch * 30) + 1598306400)::bigint;
+$$ LANGUAGE SQL;
+`)
+
+	down := batch(`
+	DROP FUNCTION IF EXISTS public.actor_tips;
```
frrist, line 21: Should this be
```sql
DROP FUNCTION IF EXISTS public.unix_to_height
DROP FUNCTION IF EXISTS public.height_to_unix
```
instead?

frrist (approved):
Good Stuff!

storage/migrations/19_height_functions.go
olizilla: it very very should

iand (approved):
LGTM but it needs to be changed to migration 20 now

alanshaw (approved):

olizilla:
Updated to be migration 21. What's the etiquette round here? This seems innocuous and helpful... I'm ok with them being hardcodded for mainnet only for now. If we made them select the info from a table at runtime, they would be slower, but work for every network... perhaps visor could provide the genesis info to the template at migration time? That sounds more complicated, but seems optimal for perf. 

I have no numbers on the perf impact of selecting the genesis info at rumtime, but we want to be able to run this function on every row returned, to convert the height to a time that can be passed to grafana, so I'd assume it'd be worth hardcoding it if we can.

I'm going to deploy the functions manually on the analysis and montoring db today, as they are mainnet specific already, unless anyone objects.

iand:
Happy for you to merge when ready

PR #257, opened 2020-11-20
fix: totalGasLimit and totalUniqueGasLimit are correct

frrist:
- values were flipped

### DIFF
--- a/tasks/message/message.go
+++ b/tasks/message/message.go
@@ -271,11 +271,11 @@ func (p *MessageProcessor) extractMessageModels(ctx context.Context, node lens.A
 				Message: message.Cid().String(),
 			})
 
-			totalUniqGasLimit += message.GasLimit
+			totalGasLimit += message.GasLimit
 			if _, seen := msgsSeen[message.Cid()]; seen {
 				continue
 			}
-			totalGasLimit += message.GasLimit
+			totalUniqGasLimit += message.GasLimit
 
 			// record this message for processing by later stages
 			pmsgModels = append(pmsgModels, visor.NewProcessingMessage(message, int64(ts.Height())))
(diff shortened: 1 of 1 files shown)


iand:
Happy with the fix but we also need to correct existing data

frrist:
@iand in 258c3f3503ce06a354da78369f02b5b414dac32c I have added a command that will attempt to recompute the incorrect gas values.

commands/fixgas.go
```diff
@@ -0,0 +1,107 @@
...
+	if err := db.DB.RunInTransaction(ctx, func(tx *pg.Tx) error {
+		err := tx.ModelContext(ctx, &msgs).Select()
+		if err != nil {
+			return err
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+	for _, msg := range msgs {
```
iand, line 73: There are 383k rows in this table. Maybe the select will work but I don't know about the update. Is there pagination support in go-pg?

frrist: I'm not sure, looking now.

commands/fixgas.go
```diff
@@ -0,0 +1,130 @@
...
+
+		msg.GasLimitTotal = total
+		msg.GasLimitUniqueTotal = unique
+		msg.BaseFee = baseFee
+		msg.BaseFeeChangeLog = math.Log(baseFeeChangeF) / math.Log(1.125)
+		msg.GasFillRatio = float64(total) / float64(len(ts.Blocks())*build.BlockGasTarget)
+		msg.GasCapacityRatio = float64(unique) / float64(len(ts.Blocks())*build.BlockGasTarget)
+		msg.GasWasteRatio = float64(total-unique) / float64(len(ts.Blocks())*build.BlockGasTarget)
+	}
+	batchSize := 1000
```
frrist, line 103: I've added batch here, if you'd like I can make this configurable via an env var.

iand:
I haven't run this. I don't expect to merge this PR, will just run from branch

frrist:
@iand I have moved the command that fixes the table values to: #258. I think we will want to merge this PR to address the bug but agree that command should not be merged.

frrist:
@iand any objection to merging this?

iand:
No objections

PR #258, opened 2020-11-23
[Do Not Merge] fix: command to fix gas value mismatch

frrist:
Command for fixing incorrect gas value, a follow on to #257


iand:
Ran this against `sentinel-analysis`. Have not run against `sentinel-monitoring` since that database requires migrating and I'm not sure if that will affect the current visor setup that is feeding alerting.

frrist:
Closing, but leaving the branch around in case we need this for other visor instances.

PR #261, opened 2020-11-23
test: add init actor extractor tests

### LINKED ISSUE #262: Init actor extractor produces actorID instead of address for genesis extraction
### Describe the bug:
This bug will only occur during genesis state extraction:
https://github.com/filecoin-project/sentinel-visor/blob/4350191538e0dfe573f074357359062a9022474b/tasks/actorstate/init.go#L37-L42
The extractor code wrongly uses an actorID instead of the actor ID address when creating the model, this will result in the genesis state containing actorID's instead of ID addresses -- `100` vs `f0100`.

### Steps to Reproduce:
Run init actor extraction logic over genesis state

### Fix:
I found this while adding test coverage for the init actor in #261. I have fixed the bug in the same PR. Applying the fix to database state should be relatively simple, we just need to reprocess the init actor state at genesis once the fix is applied.

frrist:
adds init actor extraction test coverage and fixes #262

### DIFF
--- a/tasks/actorstate/actorstate_test.go
+++ b/tasks/actorstate/actorstate_test.go
@@ -12,12 +12,15 @@ import (
 	miner "github.com/filecoin-project/lotus/chain/actors/builtin/miner"
 	"github.com/filecoin-project/lotus/chain/types"
 	bstore "github.com/filecoin-project/lotus/lib/blockstore"
+	sa0init "github.com/filecoin-project/specs-actors/actors/builtin/init"
 	samarket "github.com/filecoin-project/specs-actors/actors/builtin/market"
 	sa0power "github.com/filecoin-project/specs-actors/actors/builtin/power"
 	sa0reward "github.com/filecoin-project/specs-actors/actors/builtin/reward"
 	"github.com/filecoin-project/specs-actors/actors/util/adt"
+	sa2init "github.com/filecoin-project/specs-actors/v2/actors/builtin/init"
 	sa2power "github.com/filecoin-project/specs-actors/v2/actors/builtin/power"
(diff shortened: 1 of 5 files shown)


tasks/actorstate/init.go
```diff
@@ -35,8 +35,12 @@ func (InitExtractor) Extract(ctx context.Context, a ActorInfo, node ActorStateAP
 
 		out := initmodel.IdAddressList{}
 		if err := initActorState.ForEachActor(func(id abi.ActorID, addr address.Address) error {
+			idAddr, err := address.NewIDAddress(uint64(id))
+			if err != nil {
+				return err
+			}
 			out = append(out, &initmodel.IdAddress{
-				ID:        id.String(),
+				ID:        idAddr.String(),
```
frrist, line 43: bug fix: https://github.com/filecoin-project/sentinel-visor/issues/262

tasks/actorstate/actorstate_test.go
```diff
@@ -319,3 +322,19 @@ func (m *MockAPI) newEmptyRewardStateV0(currRealizedPower abi.StoragePower) (*sa
 func (m *MockAPI) newEmptyRewardStateV2(currRealizedPower abi.StoragePower) (*sa2reward.State, error) {
 	return sa2reward.ConstructState(currRealizedPower), nil
 }
+
+func (m *MockAPI) newEmptyInitStateV0() (*sa0init.State, error) {
```
willscott, line 326: should MockAPI move to it's own file?

willscott (approved):

tasks/actorstate/actorstate_test.go
frrist: Definitely should. Could also accept a `*testing.T` and stop returning errors -- going to make these fixups my next PR.

PR #263, opened 2020-11-24
polish(mockapi): accepts testing.TB and produces less errors

frrist:
- remove returned errors from mock api helper methods
- condense tipset creation and storeage of tipset in mock api
- rename methods to reflect their action -- `must` will do the thing or fail the test.

### DIFF
--- a/tasks/actorstate/actorstate_test.go
+++ b/tasks/actorstate/actorstate_test.go
@@ -2,6 +2,7 @@ package actorstate_test
 
 import (
 	"context"
+	"testing"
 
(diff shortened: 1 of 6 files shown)


willscott (approved):

iand (approved):
Change looks ok, but I generally prefer not to embed testing.TB since it limits opportunities for parallelising tests against a mock that has been pre-prepared. Also handling the errors in the test allows you to log more context around the failure, rather than the generic "failed to make empty adt".

PR #264, opened 2020-11-25
Export some helpers and add some types

hsanjuan:
The purpose of this rather superficial changes is to better support custom, ad-hoc tooling that re-uses logic present in sentinel-visor.

### DIFF
--- a/commands/setup.go
+++ b/commands/setup.go
@@ -34,10 +34,11 @@ import (
 
 var log = logging.Logger("visor")
 
+// RunContext wraps lens and DB connections.
 type RunContext struct {
-	opener lens.APIOpener
-	closer lens.APICloser
-	db     *storage.Database
+	Opener lens.APIOpener
+	Closer lens.APICloser
+	DB     *storage.Database
 }
 
 // SetupStorageAndAPI setups of the sentinel database and returns a
@@ -98,9 +99,9 @@ func SetupStorageAndAPI(cctx *cli.Context) (context.Context, *RunContext, error)
 	}
 
 	return ctx, &RunContext{
-		opener: opener,
-		closer: closer,
-		db:     db,
+		Opener: opener,
+		Closer: closer,
+		DB:     db,
 	}, nil
 }
 
--- a/commands/setup.go
+++ b/commands/setup.go
@@ -40,7 +40,9 @@ type RunContext struct {
 	db     *storage.Database
 }
 
-func setupStorageAndAPI(cctx *cli.Context) (context.Context, *RunContext, error) {
+// SetupStorageAndAPI setups of the sentinel database and returns a
+// ready-to-use RunContext with Openers and Closers using the chosen lens.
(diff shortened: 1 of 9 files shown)


commands/setup.go
```diff
@@ -34,13 +34,16 @@ import (
...
-	db     *storage.Database
+	Opener lens.APIOpener
+	Closer lens.APICloser
+	DB     *storage.Database
 }
 
-func setupStorageAndAPI(cctx *cli.Context) (context.Context, *RunContext, error) {
+// SetupStorageAndAPI setups of the sentinel database and returns a
+// ready-to-use RunContext with Openers and Closers using the chosen lens.
+func SetupStorageAndAPI(cctx *cli.Context) (context.Context, *RunContext, error) {
```
hsanjuan, line 46: Purpose is that we can call SetupStorageAndAPI from a different program easily and then be able to handle DB closing from there.

model/actors/common/actors.go
```diff
@@ -28,13 +31,33 @@ func (a *Actor) PersistWithTx(ctx context.Context, tx *pg.Tx) error {
...
+	if len(actors) == 0 {
+		return nil
+	}
+	if _, err := tx.ModelContext(ctx, &actors).
+		OnConflict("do nothing").
+		Insert(); err != nil {
+		return fmt.Errorf("persisting actors: %w", err)
+	}
+	return nil
+}
```
hsanjuan, line 51: Doing this should be much faster than doing one insert per actor.

model/actors/miner/sectorevents.go
```diff
@@ -42,6 +42,11 @@ type MinerSectorEventList []*MinerSectorEvent
 func (l MinerSectorEventList) PersistWithTx(ctx context.Context, tx *pg.Tx) error {
 	ctx, span := global.Tracer("").Start(ctx, "MinerSectorEventList.PersistWithTx", trace.WithAttributes(label.Int("count", len(l))))
 	defer span.End()
+
+	if len(l) == 0 {
+		return nil
+	}
```
hsanjuan, line 48: psql errors when being given a nil slice.

model/actors/miner/task.go
```diff
@@ -82,10 +84,74 @@ func (res *MinerTaskResult) Persist(ctx context.Context, db *pg.DB) error {
...
+type MinerTaskLists struct {
+	MinerInfoModel           MinerInfoList
+	FeeDebtModel             MinerFeeDebtList
+	LockedFundsModel         MinerLockedFundsList
+	CurrentDeadlineInfoModel MinerCurrentDeadlineInfoList
+	PreCommitsModel          MinerPreCommitInfoList
+	SectorsModel             MinerSectorInfoList
+	SectorEventsModel        MinerSectorEventList
+	SectorDealsModel         MinerSectorDealList
+}
```
hsanjuan, line 108: Allows processing all miners in a tipset, collecting all the models and then persisting that. Should help towards making a single Tx per tipset and be faster than using MinerTaskResultList.

frrist (approved):
Good stuff!

PR #266, opened 2020-11-25
build(ci): add go mod tidy check

frrist:


willscott:
prefer doing this in the CI task, or as a github action?

frrist:
I don't have a strong opinion on either. I've gone with circleCI here since there is prior art for this type of thing in Lotus and Specs-Actors. What do you think?

willscott:
seems reasonable

willscott (approved):

PR #268, opened 2020-11-26
fix: optimisable height functions

olizilla:
add attributes to the height functions to allow postgres to optimise them

- IMMUTABLE - the function cannot modify the database and always returns the same result when given the same argument value
- RETURNS NULL ON NULL INPUT - the function is not executed when there are null arguments; instead a null result is assumed automatically.
- PARALLEL SAFE - safe to run in parallel mode without restriction.

with these in place my test query execution time goes down from 3m40s to 12s, which is as fast as the inlined, functionless version.

see: https://www.postgresql.org/docs/12/sql-createfunction.html

License: MIT

### DIFF
--- a/storage/migrations/22_height_functions_fast.go
+++ b/storage/migrations/22_height_functions_fast.go
@@ -1,0 +1,30 @@
+package migrations
+
+import (
+	"github.com/go-pg/migrations/v8"
+)
+
+// Schema version 22 updates adds hints to our helper fns so postgres can optimise them.
+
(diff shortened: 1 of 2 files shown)


olizilla:
This change is idempotent and is already applied to the monitoring and analysis dbs

alanshaw:
Just, facinating.

`IMMUTABLE RETURNS NULL ON NULL INPUT PARALLEL SAFE` - I bet that's a sentence you never thought you'd type.

alanshaw (approved):

iand (approved):

PR #270, opened 2020-11-26
feat(gas outputs): Add Height and ActorName

hsanjuan:
Gas outputs are very important for many base-fee and gas-usage related views.

The table has no height, which means it needs to be joined with blocks all the
time on something like state_root to get a timestamp. This is likely
innefficient.

Additionally, it needs to be joined with actors to get the actor code. Since
we are "deriving" and already include redundant info, we might also get this
one. This allows easy filtering of messages by actor type and method.


model/derived/gasoutputs.go
```diff
@@ -65,8 +67,3 @@ func (l GasOutputsList) PersistWithTx(ctx context.Context, tx *pg.Tx) error {
 	}
 	return nil
 }
-
-type ProcessingGasOutputs struct {
-	Height int64
-	GasOutputs
-}
```
hsanjuan, line 72: I am not sure why this was added this way, but if we include height in the output view, then we don't need it.

model/derived/gasoutputs.go
```diff
@@ -12,6 +12,7 @@ import (
 
 type GasOutputs struct {
 	tableName          struct{} `pg:"derived_gas_outputs"` //nolint: structcheck,unused
+	Height             int64    `pg:",pk,use_zero,notnull"`
```
hsanjuan, line 15: We will need to backfill the value for this column.

hsanjuan: The value corresponds to the height of the message I think (not that of the receipts).

willscott, line 15: will this change need a schema migration?

tasks/message/gasoutputs.go
```diff
@@ -125,6 +132,35 @@ func (p *GasOutputsProcessor) processItem(ctx context.Context, node lens.API, it
...
+	var dstActorCode cid.Cid
+	dstActor, err := st.GetActor(dstAddr)
+	if err != nil {
+		if !errors.Is(err, types.ErrActorNotFound) {
+			return xerrors.Errorf("get destination actor for gas outputs %s failed: %w", item.Cid, err)
+		}
+	} else {
+		dstActorCode = dstActor.Code
+	}
+
```
hsanjuan, line 163: I wish there was a better way, but this should be ok if the state is cached at some layer.

model/derived/gasoutputs.go
iand: I think I added it because I realised height was missing from the underlying table but I needed to return it here to optimize the query

model/derived/gasoutputs.go
iand: Yes, it's the height of the message. Backfilling is a problem. With the current version it will take some time. With the new in-order processing it takes <1s per tipset, but that's still ~211000 seconds = ~2.4 days unless we shard the processing.

An alternative is to figure out a query that can update this table by joining with messages. That will still be slow to run, several hours at least.

tasks/message/gasoutputs.go
iand: https://github.com/filecoin-project/sentinel-visor/pull/249 is much more efficient but will be a few days before it's ready. Currently checking that the data it produces matches the existing data.

model/derived/gasoutputs.go
hsanjuan: added below, forgot in first commit.

hsanjuan: Custom in-order processing that just does messages and just updates this should be faster (?). That said, 2.4 days should be ok.

tasks/message/gasoutputs.go
hsanjuan: yeah, still need to support it in the "standard" way which I trust will become way faster with store improvement and block caching at some point.

storage/migrations/22_gasoutputs_height_code.go
```diff
@@ -0,0 +1,18 @@
+package migrations
+
+import "github.com/go-pg/migrations/v8"
+
+// Schema version 22 adds Height and ActorName to gas outputs table
+
+func init() {
+	up := batch(`
+	ALTER TABLE public.derived_gas_outputs ADD COLUMN height int8;
```
iand, line 9: height needs to be `bigint`

iand: We ought to add an index, but we can't until we populate the height. We could add it in a later migration.

iand:
Can we drop the message size column? Is it useful?

storage/migrations/22_gasoutputs_height_code.go
hsanjuan: fixed the type

hsanjuan:
> Can we drop the message size column? Is it useful?

I haven't used it for anything, but it seems like it could give an interesting metric.

storage/migrations/22_gasoutputs_height_code.go
```diff
@@ -0,0 +1,18 @@
+package migrations
+
+import "github.com/go-pg/migrations/v8"
+
+// Schema version 22 adds Height and ActorName to gas outputs table
+
+func init() {
+	up := batch(`
+	ALTER TABLE public.derived_gas_outputs ADD COLUMN height bigint;
+	ALTER TABLE public.derived_gas_outputs ADD COLUMN actor_name text;
```
frrist, line 10: I believe constraints need to be added for these new columns since neither are null-able, and `height` is a primary key.

tasks/message/gasoutputs.go
```diff
@@ -125,6 +132,35 @@ func (p *GasOutputsProcessor) processItem(ctx context.Context, node lens.API, it
 	stop := metrics.Timer(ctx, metrics.ProcessingDuration)
 	defer stop()
 
+	// Note: this item will only be processed if there are receipts for
+	// it, which means there should be a tipset at height+1.  This is only
+	// used to get the destination actor code, so we don't care about side
+	// chains.
```
frrist, line 138: Perhaps its worth adding a check to ensure `child != item`? Alternatively, null rounds can be handled via: https://github.com/filecoin-project/sentinel-visor/pull/214/files#diff-a771f3c826cfe5ebe05c1a003873af47456ee3e16188c1e5d0a1162c0b3e27acR40

storage/migrations/22_gasoutputs_height_code.go
hsanjuan: Do you mean `height` should be made primary key along with `cid`?

hsanjuan: Please have a look to the latest commits because I am not sure I know what I'm doing with the migrations..

tasks/message/gasoutputs.go
hsanjuan: Given that I just want to extract the dstActorCode, is it a problem if child == parent? The actor code will be the same even if it comes from a slightly wrong state. But maybe I'm missing something?

storage/migrations/24_gasoutputs_height_code.go
```diff
@@ -0,0 +1,21 @@
...
+
+import "github.com/go-pg/migrations/v8"
+
+// Schema version 22 adds Height and ActorName to gas outputs table
+
+func init() {
+	up := batch(`
+	ALTER TABLE public.derived_gas_outputs ADD COLUMN height bigint NOT NULL;
+	ALTER TABLE public.derived_gas_outputs ADD COLUMN actor_name text NOT NULL;
+	ALTER TABLE public.derived_gas_outputs ADD PRIMARY KEY (height);
```
frrist, line 11: :heavy_check_mark:

tasks/message/gasoutputs.go
frrist: I believe the destination actor may no exist in `child` (when child == parent) if this is the first message being sent to the destination actor. But based on the [below code](https://github.com/filecoin-project/sentinel-visor/pull/270/files#diff-869b68f5505c6de42fcd2d730c0d2e06d684800e092a071b45ff9ff3a385c5adR154-R162) maybe that isn't an issue -- We just won't know the designation actor code if the actor is created in a block preceding a null round.

frrist (approved):

storage/migrations/25_gasoutputs_height_code.go
```diff
@@ -0,0 +1,22 @@
...
+
+// Schema version 22 adds Height and ActorName to gas outputs table
+
+func init() {
+	up := batch(`
+	ALTER TABLE public.derived_gas_outputs ADD COLUMN height bigint NOT NULL;
+	ALTER TABLE public.derived_gas_outputs ADD COLUMN actor_name text NOT NULL;
+	ALTER TABLE public.derived_gas_outputs DROP CONSTRAINT derived_gas_outputs_pkey;
+	ALTER TABLE public.derived_gas_outputs ADD PRIMARY KEY (cid, height);
+`)
```
hsanjuan, line 13: @iand can shall we leave like this, or add state_root too?

storage/migrations/25_gasoutputs_height_code.go
```diff
@@ -0,0 +1,22 @@
...
+
+import "github.com/go-pg/migrations/v8"
+
+// Schema version 22 adds Height and ActorName to gas outputs table
+
+func init() {
+	up := batch(`
+	ALTER TABLE public.derived_gas_outputs ADD COLUMN height bigint NOT NULL;
+	ALTER TABLE public.derived_gas_outputs ADD COLUMN actor_name text NOT NULL;
+	ALTER TABLE public.derived_gas_outputs DROP CONSTRAINT derived_gas_outputs_pkey;
```
hsanjuan, line 11: I also see in other migrations we do `_pk` instead of `_pkey`, should I change that?

iand: pkey is the convention used by postgres when you create a primary key

storage/migrations/25_gasoutputs_height_code.go
iand: let's add state_root - I'll do it

PR #272, opened 2020-11-27
fix: do not error when no specialized state processor

alanshaw:
Hopefully I'm reading this right but if there's no _specialized_ state processor, `processActor` will return an error and abort the _whole_ batch. There's no specialized state processor for account actors so I imagine this happens frequently and slows down the process significantly.

This PR allows there to be no specialized processor, allowing general state information to be extracted and saved and the rest of the batch processed.

### DIFF
--- a/tasks/actorstate/actorstate.go
+++ b/tasks/actorstate/actorstate.go
@@ -314,7 +314,7 @@ func (p *ActorStateProcessor) processActor(ctx context.Context, node lens.API, i
 	// Find a specific extractor for the actor type
 	extractor, exists := p.extractors[info.Actor.Code]
 	if !exists {
-		return xerrors.Errorf("no extractor defined for actor code %q", info.Actor.Code.String())
+		return nil
 	}
 
 	ctx, _ = tag.New(ctx, tag.Upsert(metrics.TaskType, ActorNameByCode(info.Actor.Code)))


tasks/actorstate/actorstate.go
```diff
@@ -314,7 +314,7 @@ func (p *ActorStateProcessor) processActor(ctx context.Context, node lens.API, i
 	// Find a specific extractor for the actor type
 	extractor, exists := p.extractors[info.Actor.Code]
 	if !exists {
-		return xerrors.Errorf("no extractor defined for actor code %q", info.Actor.Code.String())
+		return nil
```
iand, line 317: Can we log instead

alanshaw: I considered this but it'll be a lot of noise since the same message will be logged over and over for every actor we don't have a specialized extractor for. I came the the conclusion it wasn't useful information. We have a [`Debug` method that returns this same error](https://github.com/filecoin-project/sentinel-visor/blob/5bef7da1e9a6aead50d6585fc3e08ba582d7c354/tasks/actorstate/actorstate.go#L183) and I expect if we were to log it we'd want it to be _debug_ level.

iand: Fair enough. In my branch I count the number that we didn't have extractors for and report it at the end. https://github.com/filecoin-project/sentinel-visor/blob/iand/watcher/chain/actor.go#L185

alanshaw:
Ignore this PR, records in the batch are already filtered by actor codes we have specialized extractors for.

PR #278, opened 2020-11-30
feat: extract basic account actor states

alanshaw:
~~This PR enables Sentinel to persist state for the burnt funds actor without having to extract and persist data for ALL accounts.~~

~~It registers an `AccountExtractor` for the Account Actor CID and introduces the concept of a `FilteredExtractor`. If an extractor is a `FilteredExtractor` then `processActor` will call it's `Filter` method to see if it wants to extract data for a given actor.~~

~~At the moment the `AccountExtractor` is only interested in the burnt funds account and has no specialized extraction needs beyond what is automatically extracted by the common extractor. It means it's `Extract` method is basically a noop.~~

This PR adds an account actor extractor so that basic state is extracted for account actors.

### DIFF
--- a/tasks/actorstate/account.go
+++ b/tasks/actorstate/account.go
@@ -4,14 +4,16 @@ import (
 	"context"
 
 	"github.com/filecoin-project/sentinel-visor/model"
-	builtin "github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	sa0builtin "github.com/filecoin-project/specs-actors/actors/builtin"
+	sa2builtin "github.com/filecoin-project/specs-actors/v2/actors/builtin"
 )
 
 // AccountExtractor is a state extractor that deals with Account actors.
 type AccountExtractor struct{}
 
 func init() {
-	Register(builtin.AccountActorCodeID, AccountExtractor{})
+	Register(sa0builtin.AccountActorCodeID, AccountExtractor{})
+	Register(sa2builtin.AccountActorCodeID, AccountExtractor{})
 }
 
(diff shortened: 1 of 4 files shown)


iand (commented):
What's the advantage of this over simply extracting basic actor state and json for every actor? That step is very fast compared to parsing the actor state into separate tables.

alanshaw:
Extracting the basic state for all actors will have a time and storage cost and querying the table will become more difficult due to the amount of data.

Given sentinel's remit perhaps this is acceptable and necessary?

frrist:
I would prefer to modify the `ActorStateProcessor` such that all actor types are extracted to the `actor_state` table.

> querying the table will become more difficult due to the amount of data

Is this a hunch or have we observed this to be the case with the `actor_state` table?

alanshaw:
I've modified so that it collects basic state for all account actors.

tasks/actorstate/account.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"github.com/filecoin-project/sentinel-visor/model"
+	builtin "github.com/filecoin-project/specs-actors/v2/actors/builtin"
+)
+
+// AccountExtractor is a state extractor that deals with Account actors.
+type AccountExtractor struct{}
+
+func init() {
+	Register(builtin.AccountActorCodeID, AccountExtractor{})
```
frrist, line 14: Register the account actor v1 codeID in addition to the v2 (this) one.

frrist (approved):

PR #291, opened 2020-12-02
fix(schema): fix primary key for miner_sector_deals table

iand:
Fixes https://github.com/filecoin-project/sentinel-visor/issues/280

### DIFF
--- a/storage/migrations/23_miner_sector_deals_pk.go
+++ b/storage/migrations/23_miner_sector_deals_pk.go
@@ -1,0 +1,24 @@
+package migrations
+
+import (
+	"github.com/go-pg/migrations/v8"
+)
+
+// Schema version 23 fixes the miner sector deals primary key
+
+func init() {
+	up := batch(`
+		-- old index from when table was named miner_deal_sectors
+		ALTER TABLE public.miner_sector_deals DROP CONSTRAINT IF EXISTS miner_deal_sectors_pkey;
+
+		ALTER TABLE public.miner_sector_deals DROP CONSTRAINT IF EXISTS miner_sector_deals_pkey;
+	   	ALTER TABLE public.miner_sector_deals ADD PRIMARY KEY (height, miner_id, sector_id, deal_id);
+`)
+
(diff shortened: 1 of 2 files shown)


storage/migrations/23_miner_sector_deals_pk.go
```diff
@@ -0,0 +1,24 @@
...
+
+// Schema version 23 fixes the miner sector deals primary key
+
+func init() {
+	up := batch(`
+		-- old index from when table was named miner_deal_sectors
+		ALTER TABLE public.miner_sector_deals DROP CONSTRAINT IF EXISTS miner_deal_sectors_pkey;
+
+		ALTER TABLE public.miner_sector_deals DROP CONSTRAINT IF EXISTS miner_sector_deals_pkey;
+	   	ALTER TABLE public.miner_sector_deals ADD PRIMARY KEY (height, miner_id, sector_id, deal_id);
```
hsanjuan, line 15: Would it be better to have just deal_id as primary key? A deal can only belong in one sector (afaik). Although this might be slightly better in the case of forks, where a deal ends up on a different sector_id or height.

General question, is it better to have more or less columns as primary keys? 

LGTM otherwise.

hsanjuan (approved):

storage/migrations/23_miner_sector_deals_pk.go
iand: height has to be in the primary key because we use it to partition the table as a timescale hypertable. 

It depends on our use cases. I think in this case we would benefit from having all the columns indexed since pg will be able to use it for a large number of queries where we join against miner_id or sector_id.

It's not clear to me from the spec whether deal_id is globally unique or simply unique for a miner. The GetDeal api call suggests it is globally unique and my scan of the actors code suggests it's generated as a unique sequence, so I assume we could have a separate unique index on deal_id.

hsanjuan: i leave it up to you.

PR #297, opened 2020-12-03
test(chain): test chain economics processor

frrist:
This feels like a lot of typing for little reward. For now I'm wanting try this with `github.com/stretchr/testify/mock` but remain open to alternatives.


iand (commented):
I think the mock of the lens is valuable, as is reducing the node api to a smaller interface. We should always be writing code that accepts the smallest possible interface anyway.

However, much of this code is going to be redundant as soon as we move away from the random access approach used by the `run` command.  LeaseTipSetEconomics and MarkTipSetEconomicsComplete won't be needed and the chain economics extraction interface becomes a lot simpler.

Currently have a second implementation of chain economics extraction in the chain package, which I expect to become the primary. I'm going to spend some time today to try and reduce that duplication by refactoring  `extractChainEconomicsModel` to take a TipSet instead of a TipSetKey and exporting it.

iand:
See https://github.com/filecoin-project/sentinel-visor/pull/298 for a refactoring based on this

frrist:
Closing in favor of #298. I appreciate the help here @iand

PR #298, opened 2020-12-03
test(chain): refactor and test chain economics extraction

iand:


tasks/chain/economics.go
```diff
@@ -134,27 +141,38 @@ func (p *ChainEconomics) processItem(ctx context.Context, node lens.API, item *v
...
+	if err := p.storage.RunInTransaction(ctx, func(tx *pg.Tx) error {
 		return ce.PersistWithTx(ctx, tx)
 	}); err != nil {
 		return xerrors.Errorf("persist: %w", err)
 	}
 
 	return nil
 }
+
+type ChainEconomicsLens interface {
```
willscott, line 160: maybe disambiguate this from the current definition of lens used in sentinel. This is the subset of the API interface used, not an implementation of a backend.

frrist (approved):

PR #306, opened 2020-12-09
create visor schema

### LINKED ISSUE #218: Move visor tables to visor schema
### Description

Currently all tables are in the public schema. Postgres allows tables and views to be namespaced into separate schemas. We should organise our database so visor tables are grouped into a schema called `visor`.

### Acceptance criteria

 - No tables used by visor are in the `public` schema

frrist:
### Status
Cannot merge until we develop a plan for preventing dashboard breakages.
 
### What
- remove the table `blocks_synced` as it is not longer used: 506d09c 
- create a visor schema and move everything to it: b1696c7 
  - Note: this leave the table `gopg_migrations` in the public schema.

### Impact
Inorder for existing queries to remain functional after this migration the executing users `search_path` must be set to `visor`. Without that set all existing queries (so dashboards) will fail since the default search_path of the is usually `public`.

closes #218

### DIFF
--- a/storage/migrations/26_remove_unused_tables_again.go
+++ b/storage/migrations/26_remove_unused_tables_again.go
@@ -1,0 +1,24 @@
+package migrations
+
+import "github.com/go-pg/migrations/v8"
+
+// Schema version 26 removes unused blocks_synced table;
+
+func init() {
+	up := batch(`
+	DROP TABLE IF EXISTS public.blocks_synced;
+`)
+
+	down := batch(`
+CREATE TABLE public.blocks_synced (
+    cid text NOT NULL,
+    synced_at integer NOT NULL,
+    processed_at integer
+);
+ALTER TABLE ONLY public.blocks_synced
+    ADD CONSTRAINT blocks_synced_pk PRIMARY KEY (cid);
+CREATE UNIQUE INDEX blocks_synced_cid_uindex ON public.blocks_synced USING btree (cid, processed_at);
+`)
+
+	migrations.MustRegisterTx(up, down)
+}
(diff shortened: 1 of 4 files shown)


iand (commented):
I'd like to see this tested against staging first. Pretty sure we'll need to change some user permissions too.

frrist:
At the very least the `search_path` will need to be set to `visor` in order for existing queries to continue working ([example in testing code](https://github.com/filecoin-project/sentinel-visor/pull/306/files#diff-7a254d0ebd5726aba3c5f50953bd3b744023c2ff1c90e348c86aea88d77d6414R67)). Without that set all dashboard queries using the `postgres` user will fail since the default `search_path` of the `postgres` user is `public`. 

One way to address this is by creating a `sentinel` user who's default `search_path` is set to `visor`. Dashboards could then execute their queries as as the `sentinel` user without needing to change their queries.

iand (commented):
We can't merge this until we have a plan for how to avoid breaking dashboards. This relates to our recent discussion where we identified that dashboards/grafana should track the visor schema version somehow.

hsanjuan:
This seems like a good thing to do. Let's merge this and see what breaks during a planned rollout. We only have a handful of dashboards and users and the more we wait the more complex it is going to get to fix it.

iand:
Needs reviewing to include recent schema updates

frrist:
Any objection to keeping this change on ice until we decide to flatten our schema?

iand:
Closed in favour of https://github.com/filecoin-project/sentinel-visor/pull/528

PR #308, opened 2020-12-10
chore(api): trim the lens API to required methods

frrist:
### Motivation
There has been some discussion of embedding visor into a lotus daemon. Reducing the API required by visor aims to decrease friction of making that change.


iand (commented):
Test failure is real

iand (approved):

PR #324, opened 2020-12-15
Allow model upsertion

### LINKED ISSUE #292: Support upserts
### Description

Currently all Visor models are inserted using `ON CONFLICT DO NOTHING` which discards the data if it conflicts with existing data. This is fine for appending new data but unsatisfactory when backfilling after changes to extraction logic or bug fixes. Visor should support upserting of data if a conflict is encountered. We may want to make this an optional flag for use when backfilling since in normal operation all writes are appends.

### Acceptance criteria

 - Visor can be run in a mode where rows are updated if their primary key already exists in the database

### Where to begin

 - Investigate go-pg support for upserting. If no built in support then change the `OnConflict` clause call in each model's persist method to update the non primary key columns.

frrist:
closes #292

### DIFF
--- a/storage/sql.go
+++ b/storage/sql.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"reflect"
+	"sort"
 	"strings"
 	"time"
 
@@ -89,7 +90,7 @@ var (
 
 const MaxPostgresNameLength = 64
 
-func NewDatabase(ctx context.Context, url string, poolSize int, name string) (*Database, error) {
+func NewDatabase(ctx context.Context, url string, poolSize int, name string, upsert bool) (*Database, error) {
 	if len(name) > MaxPostgresNameLength {
 		return nil, ErrNameTooLong
 	}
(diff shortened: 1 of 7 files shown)


iand:
Love this. Will take a proper look in the morning.

storage/sql.go
```diff
@@ -755,10 +762,100 @@ func (s *TxStorage) PersistModel(ctx context.Context, m interface{}) error {
...
+// update string:
+// 	"owner_id = EXCLUDED.owner_id, worker_id = EXCLUDED.worker_id"
+func GenerateUpsertStrings(model interface{}) (string, string) {
+	// thank you kind open source maintainer: https://github.com/fatih/structs
+	st := structs.New(model)
+	var cf []string
+	var ucf []string
+
+	// walk all exported fields in the model
+	for name := range st.Map() {
```
iand, line 818: I think it would be safer to use go-pg's orm to parse the pg struct tags. We don't need to keep in sync with any modifiers that are introduced. That assumes go-pg exports all the information we need here.

iand (approved):
I like this, it's nice and clean and gives an opt-in to upserting. I made a suggestion about parsing the pg struct tags with go-pg but I'm happy to leave it up to you.

storage/sql.go
frrist: I think you're right, I didn't explore go-pg very much wrt this. I'll take another look.

PR #301, opened 2020-12-03
feat: change derived_gas_outputs value columns to numerics

iand:
This changes the model and database table to use the postgres numeric type. The struct fields remain as strings and postgres automatically converts to numeric when inserted (tested against a real postgres db to verify).

Related issue: https://github.com/filecoin-project/sentinel-visor/issues/244

PR #327, opened 2020-12-16
chore(sql): user numeric type to represent numbers

### LINKED ISSUE #244: Store fil value in db as numeric type
### Description

We currently store FIL values as `TEXT` in the db. We often want to aggregate over the numeric value of the FIL, so lets store FIL as `NUMERIC`.

> The type numeric can store numbers with a very large number of digits. It is especially recommended for storing monetary amounts and other quantities where exactness is required. Calculations with numeric values yield exact results where possible, e.g., addition, subtraction, multiplication. However, calculations on numeric values are very slow compared to the integer types, or to the floating-point types described in the next section.
> – https://www.postgresql.org/docs/12/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL

Caveat: 
> These [sentinel] tables are supposed to be the "raw" form, optimised for ingestion. The plan is to then create the queryable schema by joining and transforming into new tables or views.

Even so, I propose that we consider the raw, ingest form of the value column to be numeric rather than text.

### Acceptance criteria

The `value` column is `NUMERIC` in all tables and views that record them.

### Where to begin

- Determine if there is any reason not to do this?
- Write a migration.

frrist:
together with https://github.com/filecoin-project/sentinel-visor/pull/301 closes #244. will need a rebase if #301 goes in first.

### DIFF
--- a/model/actors/miner/feedebt.go
+++ b/model/actors/miner/feedebt.go
@@ -13,7 +13,7 @@ type MinerFeeDebt struct {
 	MinerID   string `pg:",pk,notnull"`
 	StateRoot string `pg:",pk,notnull"`
 
-	FeeDebt string `pg:",notnull"`
+	FeeDebt string `pg:"type:numeric,notnull"`
 }
 
 func (m *MinerFeeDebt) Persist(ctx context.Context, s model.StorageBatch) error {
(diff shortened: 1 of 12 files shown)


iand (approved):
Let's fold https://github.com/filecoin-project/sentinel-visor/pull/301 into this one so we have all the changes to numerics in a singfe schema change

PR #328, opened 2020-12-17
Remove Run command and refactor tasks

### LINKED ISSUE #311: Remove run command and old tasks
### Description

The new `watch` and `walk` commands supercede what we were attempting with `run`. They have a simpler CLI interface and perform better than the random access pipeline approach used by `run`.

We should remove `run` and all its associated code. This gives us more freedom to refactor `watch` and `walk` and move their core functionality into the `tasks` package. Currently we support two different paradigms which is causing friction when refactoring.

### Acceptance criteria

 - `run` command does not exist
 - All `Lease*` and `Mark*Complete` functions in `storage` package are removed
 - A migration to remove `visor_processing_tipsets`, `visor_processing_actors`, `visor_processing_messages` and `visor_processing_stats` tables
 - Under the `tasks` package
    - `actorstate.ActorStateChangeProcessor`, `actorstate.ActorStateProcessor`, `chain.ChainEconomics`, `message.GasOutputsProcessor` and `message.MessageProcessor` are removed.
   - `indexer.ChainHeadIndexer` is moved to the top level `chain` package and renamed to `chain.Watcher` 
   - `indexer.ChainHistoryIndexer` is moved to the top level `chain` package and renamed to `chain.Walker` 
   - `indexer.TipSetCache` is moved to top level `chain` package
   - `indexer.UnindexedBlockData` is removed
   - The only remaining types in `tasks/actorstate` should be actor extracters such as `StorageMarketExtractor`
   - The `tasks/chain` package should still contain `ExtractChainEconomicsModel` and its test.
   - The `tasks/indexer` package should be removed
   - The `tasks/message` package should be removed 
   - The `tasks/stats` package should be removed 
   - The `tasks/views` package should be kept (but see https://github.com/filecoin-project/sentinel-visor/issues/310)

### Where to begin

 - find the delete key on the keyboard and then :wastebasket:

frrist:
Tried to keep this as mechanical as possible. Made a couple changes to tests that I have called out below. Thank you @iand for the well written issue :heart: 
- Closes #311
- removes debug command since it was dependent on schema we have migrated away from. If this is a feature we wish to keep it will be simplest to re-implement it as a follow on to this work.
- Separate commits will be squished when this merges.


chain/watcher_test.go
```diff
@@ -80,14 +87,15 @@ func TestChainHeadIndexer(t *testing.T) {
...
-
-	tipSetKeys := []string{chainHead.Key().String()}
 
 	err = idx.index(ctx, nh)
 	require.NoError(t, err, "index")
 
+	// TODO NewTipSetIndexer runs its processors in their own go routines (started when TipSet() is called)
+	// this causes this test to behave nondeterministicly so we sleep here to ensure all async jobs
+	// have completed before asserting results
+	time.Sleep(time.Second * 3)
```
frrist, line 97: This test is still useful and I have done my best to keep it working without changing implementation logic (Trying to keep this change as mechanical as possible). If we are alright with this fix/hack for now I can file an issue to fix this as a follow on :+1: :-1:

iand: Happy to keep it like this

iand (approved):
Great :smiley:

PR #337, opened 2021-01-04
Market actor: Export some methods and improve batching

hsanjuan:
I had these changes laying around. The market actor model was still looping every element on the slice when doing a batched-persist, which is very slow. That said, the whole market-actor processing does NOT get sped-up too much by this.

Additionally, export some extractor methods for the market actor, as that allowed me playing with a custom extractor, even though it was still too slow.

### DIFF
--- a/model/actors/market/dealproposal.go
+++ b/model/actors/market/dealproposal.go
@@ -3,11 +3,10 @@ package market
 import (
 	"context"
 
+	"github.com/filecoin-project/sentinel-visor/model"
 	"go.opentelemetry.io/otel/api/global"
 	"go.opentelemetry.io/otel/api/trace"
 	"go.opentelemetry.io/otel/label"
-
-	"github.com/filecoin-project/sentinel-visor/model"
 )
 
 type MarketDealProposal struct {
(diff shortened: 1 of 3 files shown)


model/actors/market/dealproposal.go
```diff
@@ -41,10 +40,5 @@ type MarketDealProposals []*MarketDealProposal
 func (dps MarketDealProposals) Persist(ctx context.Context, s model.StorageBatch) error {
 	ctx, span := global.Tracer("").Start(ctx, "MarketDealProposals.Persist", trace.WithAttributes(label.Int("count", len(dps))))
 	defer span.End()
-	for _, dp := range dps {
-		if err := s.PersistModel(ctx, dp); err != nil {
-			return err
-		}
-	}
-	return nil
+	return s.PersistModel(ctx, dps)
```
hsanjuan, line 43: Note: PersistModel checks for length = 0.

iand (approved):

hsanjuan:
Is it ok to merge? Failures seem unrelated..

iand:
> Is it ok to merge? Failures seem unrelated..

Yes, I agree

PR #344, opened 2021-01-05
fix(miner): extract miner PoSt's from parent messages

### LINKED ISSUE #342: Not all sector PoSt messages are processed
### Describe the bug:

In the [miner actor extractor](https://github.com/filecoin-project/sentinel-visor/blob/master/tasks/actorstate/miner.go#L388) we only obtain messages from the first block in a tipset, which we then use to extract sector posts. We should be reading all messages in the tipset.

This means we are missing PoSts from any current miner extractions 

Visor Version: latest (ae6d3a86ca19e28acead2c6b5b3d21cd946a62aa)

frrist:
- closes #342

Due to the complexity of the miner actor state there there is no test coverage for this change, manual verification is required.

Manual validation was performed by running:
```
./visor --log-level=debug --db="postgres://postgres:password@localhost:5432/postgres?sslmode=disable" --lens="lotusrepo" --repo="~/.lotus" --repo-read-only=false walk --from 10000 --to 11000 --tasks=actorstatesminer
``` 
A CSV of the `miner_sector_posts` table after running the above command can be seen here: https://ipfs.io/ipfs/QmV4RbVzgtD5Nc9ujpMRqyuAsjBSzELSqo7eWY4RP7Hn26

When the aformentioned command was ran against master the `miner_sector_posts` table was empty.

### DIFF
--- a/lens/interface.go
+++ b/lens/interface.go
@@ -26,7 +26,6 @@ type StoreAPI interface {
 	Store() adt.Store
 }
 
-
 type ChainAPI interface {
 	ChainNotify(context.Context) (<-chan []*api.HeadChange, error)
 	ChainHead(context.Context) (*types.TipSet, error)
@@ -56,7 +55,6 @@ type StateAPI interface {
 	StateReadState(ctx context.Context, addr address.Address, tsk types.TipSetKey) (*api.ActorState, error)
 	StateGetReceipt(ctx context.Context, bcid cid.Cid, tsk types.TipSetKey) (*types.MessageReceipt, error)
 	StateVMCirculatingSupplyInternal(context.Context, types.TipSetKey) (api.CirculatingSupply, error)
-
 }
 
 type APICloser func()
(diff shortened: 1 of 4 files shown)


lens/interface.go
```diff
@@ -26,7 +26,6 @@ type StoreAPI interface {
 	Store() adt.Store
 }
 
-
```
frrist, line 29: gofmt changes

tasks/actorstate/actorstate_test.go
```diff
@@ -83,6 +83,10 @@ func (m *MockAPI) ChainGetBlockMessages(ctx context.Context, msg cid.Cid) (*api.
 	}, nil
 }
 
+func (m *MockAPI) ChainGetParentMessages(ctx context.Context, msg cid.Cid) ([]api.Message, error) {
+	return []api.Message{}, nil
+}
+
```
placer14, line 89: Is there a `ChainGetBlockMessages` to remove as well?

placer14 (approved):

iand (approved):

iand:
I rebased this on master and fixed some minor linting based conflicts, however the test failure looks real but unrelated to the actual change made here. Would really like to get this merged so we can start the backfill with this fix.

PR #353, opened 2021-01-13
docs: Add release process details and workflow

placer14:
Documentation of the release process I follow for `visor` so others may participate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,6 @@ A component of [**Sentinel**](https://github.com/filecoin-project/sentinel), a c
 
 A **Visor** process collects _permanent_ Filecoin chain meterics from a [**Lotus**](https://github.com/filecoin-project/lotus/) daemon, and writes them to a [**TimescaleDB**](https://github.com/timescale/timescaledb) time-series and relational datastore.
 
-
 ## Getting Started
 
 Clone the repo and build the dependencies:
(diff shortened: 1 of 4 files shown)


frrist (commented):
Perhaps the release process deserves its own file? 

I generally expect README's to contain tips on getting started, a couple usage examples, and pointers for contributing -- short and sweet. Since publishing a release is the responsibility of maintainers I'd be okay with it moving to its own document.

frrist (approved):

README.md
```diff
@@ -145,12 +145,85 @@ contains a copy of it's own release notes.
...
 
+A non-prescriptive example of the release process might look like the following:
 
+```sh
+git checkout master
+git pull                                # checkout/pull latest master
+git checkout -b vx.x.x-release          # create release branch
+vi CHANGELOG.md                         # update CHANGELOG.md
+make visor                              # validate build
+go mod tidy                             # ensure tidy go.mod for release
```
iand, line 161: tidy should be done before build so the build uses the exact same versions as specified in go.mod/go.sum. Ideally go build should be run with `-mod=readonly` so it doesn't change the files. This is the new default in the upcoming version of Go.

iand (commented):
Meta question: why isn't cutting a release as simple as tagging master once the changelog has been updated?

I think the current method is process heavy and more complex than we need at this stage of development. I'd like to see us tagging minor releases every few day and deploying them, which means keeping a lightweight approach for as long as possible. 

I'd also like to avoid having long lived version branches with hotfixes. Unless it's unavoidable we should prefer to fix in master and increment the patch version. 

The main friction is the schema version. We should be very clear in the changelog when a schema version changes and whether it is destructive or breaking.

iand:
As a follow-on thought: a new schema version should always result in at least a minor version increment.

placer14:
> why isn't cutting a release as simple as tagging master once the changelog has been updated?

I've made the following assumptions:
1.  that release candidates are a worthwhile step in validating our release. 
2. I also believe that deltas between release candidates should ONLY include fixes and not features or interface changes. 

If we simply tag latest master, there is no guarantee of the 2nd assumption being true.

iand:
> > why isn't cutting a release as simple as tagging master once the changelog has been updated?
> 
> I've made the following assumptions:
> 
>     1. that release candidates are a worthwhile step in validating our release.
> 
>     2. I also believe that deltas between release candidates should ONLY include fixes and not features or interface changes.
> 
> 
> If we simply tag latest master, there is no guarantee of the 2nd assumption being true.

I think release candidate management is a complexity that we don't need yet. We don't have a pattern of regular releases to base our decisions on nor do we have a good enough testing regime to be confident of catching regressions or problems in a release candidate. This is made even more difficult if a candidate involves a breaking schema change.

At the moment I don't feel we have any release momentum and each release seems to involve a huge amount of work.

At this stage of our development I think we should be more flexible and responsive. Tag minor releases regularly and attempt deployment as soon as we tag. Prefer to roll forward, fixing issues in a new minor release. Roll back only if there is no choice.

README.md
```diff
@@ -145,12 +145,85 @@ contains a copy of it's own release notes.
...
+    git tag -asm vx.x.x vx.x.x
     git push --tags
 
+A non-prescriptive example of the release process might look like the following:
 
+```sh
+git checkout master
+git pull                                # checkout/pull latest master
+git checkout -b vx.x.x-release          # create release branch
+vi CHANGELOG.md                         # update CHANGELOG.md
```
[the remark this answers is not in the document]
placer14: I don't know if the `make` command is useful as this workflow requires providing arguments (to define the next tag being created within `--next-tag` arg) which `make` requires envvars for. I'm considering removing the target from the makefile and letting the doc stand alone.

placer14:
I fully support a leaner release cycle, but also note that we are adding many dependencies onto the schema outputs of visor. I am looking for room in the release process to validate these dependencies (automatically or otherwise). That said, I can be satisfied with defaulting to the fast path until it becomes too painful.

README.md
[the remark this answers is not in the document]
frrist: In favor of dropping the makefile target.

placer14:
Update with feedback applied. Looking for one last review since we're discussing process here.

docs/migrations.md
```diff
@@ -0,0 +1,29 @@
+# Schema Migrations
+
+The database schema is versioned and every change requires a migration script to be executed. See [storage/migrations/README.md](storage/migrations/README.md) for more information.
```
frrist, line 3: The migration link 404's for me, but maybe it's a PR/formatting thing?

frrist (approved):
LGTM!

iand (approved):

docs/migrations.md
placer14: Good catch!

PR #356, opened 2021-01-14
fix: adjust calculation of gas outputs for FIP-0009

### LINKED ISSUE #354: Unexpected derived_gas_output.base_fee_burn values since FIP-0009
Since FIP-0009 the value for `BaseBurnFee` should be `0` for `SubmitWindowedPoSt` messges as per https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0009.md#test-cases

The visor `derived_gas_ouput` table does not show any zero values in the `base_burn_fee` column for `SubmitWindowedPoSt`

```sql
SELECT o.cid, o.actor_name, o.method, o.base_fee_burn, o.over_estimation_burn
    FROM derived_gas_outputs o
    WHERE height >= 300000
      AND height < 400000
      AND actor_name = 'fil/2/storageminer'
      AND method = 5
      AND base_fee_burn = '0'

- returns no results
```

We rely on the lotus/chain/vm `ComputeGasOutputs` function to populate the values... and that module changed with the 1.4 release of lotus that included FIP-0009 so we need to upgrade our lotus dep and reingest.

https://github.com/filecoin-project/sentinel-visor/blob/4ae639cbea3f5826ee42ef32fdc587ed39e60d06/chain/message.go#L166-L191

Visor Version: `v0.4`

iand:
Fixes #354 

Gas output calculation is now moved to the GetExecutedMessagesForTipset method since we have all the information required at that point.

Upgrades lotus dependency to 1.4.0 but https://github.com/filecoin-project/lotus/pull/5355 is required

Also requires https://github.com/willscott/carbs/pull/9


willscott:
Carbs update now tagged as v0.0.4

willscott (approved):

frrist (approved):

PR #358, opened 2021-01-15
revert: use numeric type to represent numbers

iand:
Migration 27 changes the column types in several very large tables which is extremely slow. I will add this migration back again in a new PR which we should hold until we have planned a large enough maintenance window in which to run the migration.

Reverts filecoin-project/sentinel-visor#327


iand:
As a data point, I ran `ALTER TABLE public.derived_gas_outputs ALTER COLUMN value TYPE numeric USING (value::numeric);` against the staging database and it still hasn't returned after 2.5 hours

frrist (approved):

PR #364, opened 2021-01-21
fix: improve inferred json encoding for csv output

iand:
Nil JSON values should be encoded as lower-case `null` in the CSV output, but other nil values should remain uppercase

Slices and other non-scalar types are automatically inferred to be stored as JSON by go-pg so we can use that to automatically encode. However `string` types that have the `type:jsonb` annotation in our models refer to values that have already been encoded as JSON, so the CSV output needs to avoid encoding them again.

### DIFF
--- a/storage/csv.go
+++ b/storage/csv.go
@@ -45,6 +45,7 @@ func getCSVModelTable(v interface{}) table {
 	for _, fld := range m.Fields {
 		t.columns = append(t.columns, fld.SQLName)
 		t.fields = append(t.fields, fld.GoName)
+		t.types = append(t.types, fld.SQLType)
 	}
 	csvModelTables[name] = t
 
@@ -68,6 +69,7 @@ type table struct {
 	name    string
 	columns []string
 	fields  []string
+	types   []string
 }
 
 func NewCSVStorage(path string) (*CSVStorage, error) {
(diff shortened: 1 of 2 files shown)


storage/csv.go
```diff
@@ -196,6 +194,24 @@ func (c *CSVBatch) PersistModel(ctx context.Context, m interface{}) error {
 				continue
 			}
 
+			var encodeAsJSON bool
+
+			// Strings marked as json type are assumed to already be encoded
+			if fk != reflect.String && (t.types[i] == "json" || t.types[i] == "jsonb") {
+				encodeAsJSON = true
```
frrist, line 201: Should this be false if its already encoded?

storage/csv.go
```diff
@@ -196,6 +194,24 @@ func (c *CSVBatch) PersistModel(ctx context.Context, m interface{}) error {
...
+			} else if fk == reflect.Interface {
+				encodeAsJSON = true
+			}
```
frrist, line 204: This means slices of strings/bytes/structs will be encoded as json?

frrist (approved):

storage/csv.go
iand: No, this is checking if it has a json datatype but is not a string

iand: Comment should be clearer

storage/csv.go
iand: Slices of strings encode as json. For example `MultisigTransaction.Approved` is a `[]string` with no column type tag. go-pg assigns that a json type by default, which matches our schema. Same with `MinerInfo.ControlAddresses` and `MinerInfo.MultiAddresses`

PR #371, opened 2021-01-26
chore: update to lotus 1.5.0-pre1 and actors v3

iand:
Tests for actors v3 are copy/paste of v2 tests with fixups for the types. The main change for v3 is the structure of the HAMT.


tasks/actorstate/reward_test.go
```diff
@@ -41,7 +43,7 @@ func TestRewardExtractV0(t *testing.T) {
 	stateCid, err := mapi.Store().Put(ctx, state)
 	require.NoError(t, err)
 
-	minerAddr := tutils.NewIDAddr(t, 00)
+	minerAddr := tutils.NewIDAddr(t, 0o0)
```
placer14, line 46: May be nothing, but curious why this was needed?

placer14 (approved):
Confused how the `ffi-stub` was building before the types or constants were present. I don't see where it was changed but checking out locally it obviously points to that module. Just some small confusion, but looks good otherwise. 🤝

willscott:
only the subset of ffi stub actually referenced by us or lotus is needed when a module replacement happens.

frrist (approved):
LGMT! :pray:

willscott:
ℹ️ There is a dependency on statediff for parsing message parameters. that may also need to get pulled up to the v3 dependencies to continue functioning. I expect to do that this week.

tasks/actorstate/reward_test.go
iand: Not my doing. I think gofumpt does it. Numbers starting with 0 are octal in Go, and they added `0o` as a consistent way to write them to match `0x`

iand:
> Confused how the `ffi-stub` was building before the types or constants were present. I don't see where it was changed but checking out locally it obviously points to that module. Just some small confusion, but looks good otherwise.

They were in the real ffi package but not imported into Lotus until recently.

PR #376, opened 2021-02-01
feat: improve metric reporting

iand:
Various cleanups and improvements:
 - ensure only tasks passed on command line are used in task tag on metrics
 - add tag for actor type (for processing metrics)
 - add tag for table name (for persistence metrics)
 - add count of failed processing and persistence functions
 - export the count of lens calls so it is reported properly
 - add metrics to some missed lens calls

### DIFF
--- a/model/actors/common/actors.go
+++ b/model/actors/common/actors.go
@@ -3,10 +3,12 @@ package common
 import (
 	"context"
 
+	"go.opencensus.io/tag"
 	"go.opentelemetry.io/otel/api/global"
 	"go.opentelemetry.io/otel/api/trace"
 	"go.opentelemetry.io/otel/label"
 
+	"github.com/filecoin-project/sentinel-visor/metrics"
 	"github.com/filecoin-project/sentinel-visor/model"
 )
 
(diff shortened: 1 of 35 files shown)


model/actors/common/actors.go
```diff
@@ -34,6 +41,10 @@ func (actors ActorList) Persist(ctx context.Context, s model.StorageBatch) error
 	ctx, span := global.Tracer("").Start(ctx, "ActorList.Persist", trace.WithAttributes(label.Int("count", len(actors))))
 	defer span.End()
 
+	ctx, _ = tag.New(ctx, tag.Upsert(metrics.Table, "actors"))
```
placer14, line 44: Should we be tagging metrics for list persistence the same as individual persistence? Seems this would record duration for a batch and individuals just the same and might make the data unreliable. (If you change this, it appears this is the same for all the models.)

placer14 (approved):
One question about metric capturing around batches vs individual model persistence. Approval is sticky. 🤝

model/actors/common/actors.go
iand: I don't think it matters. The list could be a single item and we would want that to be treated identically to the individual case. We're aiming to measure how long visor is taking to persist all the data it extracts, not necessarily per item timings .

PR #377, opened 2021-02-01
chore: increase lens object cache size

iand:
Metrics show that we are continually fetching ~20k blocks per tipset which suggests that the LRU cache used is too small. Increasing it to 500k made a big difference in performance as shown in grafana screenshot below. This change bumps the LRU to 50k to balance memory consumption with performance.

Top-right: note number of requests being made to API drop off dramatically with a larger LRU
Mid-right: note processing times are lower
Bottom-left: memory consumption is up, but not greatly so

The spikes appear to be unrelated and are not present once visor has been running for a while.

Screenshot from 2021-02-01 17-40-31


placer14 (approved):
Huge props on supporting data. Thank you!

frrist (approved):

PR #378, opened 2021-02-02
schema: add integer now function to each hypertable

iand:


storage/migrations/28_integer_now_funcs.go
```diff
@@ -0,0 +1,49 @@
...
+	SELECT set_integer_now_func('miner_fee_debts', 'current_height', replace_if_exists => true);
+	SELECT set_integer_now_func('miner_infos', 'current_height', replace_if_exists => true);
+	SELECT set_integer_now_func('miner_locked_funds', 'current_height', replace_if_exists => true);
+	SELECT set_integer_now_func('miner_pre_commit_infos', 'current_height', replace_if_exists => true);
+	SELECT set_integer_now_func('miner_sector_deals', 'current_height', replace_if_exists => true);
+	SELECT set_integer_now_func('miner_sector_events', 'current_height', replace_if_exists => true);
+	SELECT set_integer_now_func('miner_sector_infos', 'current_height', replace_if_exists => true);
+	SELECT set_integer_now_func('miner_sector_posts', 'current_height', replace_if_exists => true);
+	SELECT set_integer_now_func('parsed_messages', 'current_height', replace_if_exists => true);
+	SELECT set_integer_now_func('receipts', 'current_height', replace_if_exists => true);
```
placer14, line 38: Suppose we decide to have user-specific schemas assigned sometime later... (ex: user `sentinel-analysis` defaults to `analysis` schema)... would these break given the function is assigned on `public` and not on `analysis` when that user is logged in? Should we include the schema qualifier on the function name?

placer14 (approved):
One small question for a case we're not hitting yet, but might sometime later. GTM

storage/migrations/28_integer_now_funcs.go
iand: I don't know if that works but it is worth trying

frrist (approved):

storage/migrations/29_integer_now_funcs.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/go-pg/migrations/v8"
+)
+
+// Schema version 29 adds an integer now function to each hypertable
+
+func init() {
+	up := batch(`
+	-- A function that estimates the current filecoin epoch
+	CREATE OR REPLACE FUNCTION public.current_height() RETURNS bigint AS $$
+		SELECT floor((extract(epoch from now() AT TIME ZONE 'UTC') - extract(epoch from TIMESTAMP '2020-8-24 22:00:00' AT TIME ZONE 'UTC')) / 30)::bigint;
```
hsanjuan, line 13: Would it be bad to rely on the unix_to_height() function here?

hsanjuan:
This has enough approvals. Let's merge?

PR #382, opened 2021-02-15
fix(repo-util): set chainstore on repo lens util

frrist:
Ensure the chain store is set on LensAPI to avoid a null pointer when StateGetActor is called.


iand (approved):
Is the lint error related?

frrist:
Linting locally returns nothing for me; not sure what it's unhappy about in CI -- filed https://github.com/filecoin-project/sentinel-visor/issues/383 to address it.

PR #392, opened 2021-02-18
run linting in circleci

frrist:


hsanjuan (commented):
I'm never going to get used to having the equivalent of `golint ./...` needing so much setup.

Anyways, just checking that the effort here does not go to waste. :)

frrist:
This was a low effort, mostly copy-pasta from the specs-actors circleci config (was hoping it would just work :( ). I view this as low priority so I'll close this and pick back up at a later date (work is already tracked in an issue).

PR #395, opened 2021-02-19
fix: msapprovals missing pending transaction

### LINKED ISSUE #394: msapprovals task cannot find pending transaction
### Describe the bug:

From a recent run

At height 207959
```
"[{""Addr"":""f2qriauwoxp3jgdlkiweurjp3fpirqaslecdnwtei"",""Error"":""failed to find transaction: failed to read transaction details: pending transaction 38 not found""},{""Addr"":""f24qkofnqmuowbmgwm3wdq6f2jiebktp4qzpu3dxy"",""Error"":""failed to find transaction: failed to read transaction details: pending transaction 38 not found""},{""Addr"":""f2jecovubdwiid6ip7pmrrulp3fqymjtbx6fe2ebi"",""Error"":""failed to find transaction: failed to read transaction details: pending transaction 38 not found""}]"
```

At  height 207958

```
"[{""Addr"":""f2xjxf52aqa4d4sq2cdidx5fohlnpurxv3xxugfhq"",""Error"":""failed to find transaction: failed to read transaction details: pending transaction 38 not found""},{""Addr"":""f2i7eahz2xcfts6pdx2fabfpyttx434sais6xaxkq"",""Error"":""failed to find transaction: failed to read transaction details: pending transaction 42 not found""}]"
```

Looks like a defect in the code that processes approve messages to a multisig

iand:
Fixes #394

### DIFF
--- a/tasks/msapprovals/msapprovals.go
+++ b/tasks/msapprovals/msapprovals.go
@@ -83,7 +83,7 @@ func (p *Task) ProcessMessages(ctx context.Context, ts *types.TipSet, pts *types
 			continue
 		}
 
-		applied, tx, err := p.getTransactionIfApplied(ctx, m.Message, m.Receipt, ts, pts)
+		applied, tx, err := p.getTransactionIfApplied(ctx, m.Message, m.Receipt, pts)
 		if err != nil {
 			log.Debugw("failed to get transaction", "error", err.Error(), "addr", m.Message.To.String())
 			errorsDetected = append(errorsDetected, &MultisigError{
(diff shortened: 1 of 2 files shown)


tasks/msapprovals/msapprovals.go
```diff
@@ -247,7 +247,9 @@ func (p *Task) getTransactionIfApplied(ctx context.Context, msg *types.Message,
 		}
 
 		// Get state of actor before the message was applied
-		act, err := p.node.StateGetActor(ctx, msg.To, pts.Key())
+		// pts is the tipset containing the messages, so we need the state as seen at the start of message processing
+		// for that tipset
+		act, err := p.node.StateGetActor(ctx, msg.To, pts.Parents())
```
hsanjuan, line 252: Did some underlying behaviour change that now we need to use the parents?

hsanjuan (approved):

tasks/msapprovals/msapprovals.go
iand: No, but pts is the tipset containing the message, so we need to look at the actor state before the message was applied

PR #406, opened 2021-03-03
feat: record tipset cache metrics during watch

iand:
Gives us more visibility about whether visor is seeing any tipsets arrive on the chain


iand:
Lint failure seems to be problems with downloading vectors

frrist (approved):

PR #408, opened 2021-03-03
Create a Lily node by embedding lotus into visor to observe tipset apply and revert events

frrist:
### What
The PR imports and embeds the lotus daemon into Visro; currently, this node is called Lily.
Changes required to lotus live here: https://github.com/filecoin-project/lotus/pull/5693

#### Usage
1. `make` (may need to run `git submodule update --init --recursive` first)
1. `./visor lily daemon --import-snapshot minimal_finality_stateroots_517061_2021-02-20_11-00-00.car`
1. `make dockerup` to create a database
1. `./visor lily watch start` will instruct lily to start watching and processing the Filecoin blockchain and indexing it to the database.


frrist:
ohh derp I forgot to add the ffi.

node/impl/lily.go
```diff
@@ -0,0 +1,117 @@
...
+	// If we have a zero confidence window then we need to notify every tipset we see
+	if cfg.Confidence == 0 {
+		if err := indexer.TipSet(ctx, head); err != nil {
+			return err
+		}
+	}
+
+	obs := observer.NewIndexingTipSetObserver(indexer, tsCache)
+
+	if err := m.Events.Observe(obs); err != nil {
```
frrist, line 76: I think we can avoid this dependency in lotus if we go back to using the chain notify API call that is already present in the `impl.FullNodeAPI`.

node/api/apistruct/struct.go
```diff
@@ -0,0 +1,26 @@
+package apistruct
```
iand, line 1: Does this need to be a separate package?

node/cli/daemon.go
```diff
@@ -0,0 +1,282 @@
...
+	opts := node.Options(
+		node.WithRepoType(repo.FullNode),
+		node.Options(fopts...),
+		node.WithInvokesKey(node.ExtractApiKey, resAPI),
+	)
+	*out = resAPI
+	return opts
+}
+
+func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool) (err error) {
```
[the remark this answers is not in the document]
iand: We should find a better package for this function, although it's not urgent to do that

node/api/apistruct/struct.go
```diff
@@ -0,0 +1,26 @@
...
+
+	lotusstruct "github.com/filecoin-project/lotus/api/apistruct"
+	logging "github.com/ipfs/go-log/v2"
+
+	"github.com/filecoin-project/sentinel-visor/node/api"
+)
+
+var log = logging.Logger("lily-api")
+
+type LilyNodeStruct struct {
```
iand, line 14: Since this is the start of the new JSON-RPC API that visor will expose to allow the daemon to be controlled we should name it more concretely, e.g. just API. 

Would be nice if we could stop using the word `node` in other places where we mean `lens.Interface` and we should try to stop using it here too.

node/api/apistruct/struct.go
```diff
@@ -0,0 +1,26 @@
...
+	lotusstruct.FullNodeStruct
```
iand, line 15: We don't need all of this. What is the minimum we need?

node/api/apistruct/struct.go
frrist: It does not -- I had everything broken into packages like this to make it easier to mimic lotus's layout. I have sense created a "lily" lens and moved the LilyNode API and struct into it.

node/api/apistruct/struct.go
frrist: Dropped the "node" word and moved this all to the lens package

node/api/apistruct/struct.go
frrist: We need the following API's from lotus:
- `common.CommonAPI` for the RPC and authentication/permissions.
- `full.ChainAPI` for chain event notification and chain inspection.
- `full.StateAPI` for actor state extraction and inspection.

I have redeclared the `LilyAPI` to require only these.

node/cli/daemon.go
```diff
@@ -0,0 +1,282 @@
+package cli
```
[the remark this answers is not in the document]
frrist: Done.

node/cli/daemon.go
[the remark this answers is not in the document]
frrist: Yes, a copy from lotus; I agree copying this code isn't ideal but I would prefer to punt on extraction in lotus since this copy is low risk.

commands/util/rpc.go
```diff
@@ -0,0 +1,148 @@
+package util
```
frrist, line 1: mostly copy-paste from lotus to get the injection working.

chain/indexing.go
```diff
@@ -0,0 +1,45 @@
...
+func (i *IndexingTipSetObserver) Revert(ctx context.Context, ts *types.TipSet) error {
+	log.Debugw("revert tipset", "height", ts.Height(), "tipset", ts.Key().String())
+	err := i.cache.Revert(ts)
+	if err != nil {
+		log.Errorw("tipset cache revert", "error", err.Error())
+	}
+	return nil
+}
+
+var _ events.TipSetObserver = (*IndexingTipSetObserver)(nil)
```
iand, line 45: This line is the only reason for importing the events package. I don't think we need it since we can't compile without it satisfying the interface anyway

chain/indexing.go
```diff
@@ -0,0 +1,45 @@
+package chain
+
+import (
+	"context"
+
+	"github.com/filecoin-project/lotus/chain/events"
+	"github.com/filecoin-project/lotus/chain/types"
+)
+
+func NewIndexingTipSetObserver(obs TipSetObserver, cache *TipSetCache) *IndexingTipSetObserver {
```
iand, line 10: As you mentioned before, we could simply revert to using ChainNotify and then this type is redundant with Watcher. For now let's keep it and work on refactoring to a more optimized lens interface that could work by calling ChainNotiify or could hook into events directly.

commands/lily/watch.go
```diff
@@ -0,0 +1,84 @@
...
+var watchFlags watchOps
+
+var LilyWatchStartCmd = &cli.Command{
+	Name:   "start",
+	Usage:  "start a watch against the chain",
+	Before: initialize,
+	After:  destroy,
+	Flags: []cli.Flag{
+		&cli.IntFlag{
+			Name:        "indexhead-confidence",
```
iand, line 39: We'll refine the names of the params later

main.go
```diff
@@ -48,113 +50,136 @@ func main() {
 		Version: version.String(),
 		Flags: []cli.Flag{
 			&cli.StringFlag{
-				Name:    "lens",
-				EnvVars: []string{"VISOR_LENS"},
-				Value:   "lotus",
+				Name:        "lens",
+				EnvVars:     []string{"VISOR_LENS"},
+				Value:       "lotus",
+				Destination: &commands.VisorCmdFlags.Lens,
```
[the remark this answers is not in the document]
iand: Yeah maybe. I'm not 100% sold on it but willing to try it out.

iand (approved):
I'm happy to merge this to `lily` and use it as a basis for the next steps.

Issue #411, opened 2021-03-09
Panic in actorstate.NewPowerStateExtractionContext


iand:
### Describe the bug:

Several failures with panic traces like the following. 

```
{"level":"warn","ts":"2021-03-08T22:11:10.746Z","logger":"bufbs","caller":"bufbstore/buf_bstore.go:44","msg":"one of the stores is not viewable; running less efficiently"}
{"level":"info","ts":"2021-03-08T22:11:16.027Z","logger":"chain","caller":"chain/indexer.go:313","msg":"task report","height":564300,"task":"blocks","status":"OK","time":36.835269438}
{"level":"info","ts":"2021-03-08T22:11:16.027Z","logger":"chain","caller":"chain/indexer.go:313","msg":"task report","height":564300,"task":"chaineconomics","status":"OK","time":36.835310322}
{"level":"error","ts":"2021-03-08T22:11:16.027Z","logger":"chain","caller":"chain/indexer.go:285","msg":"task returned with error","height":564300,"task":"msapprovals","error":"context done: context deadline exceeded"}
{"level":"warn","ts":"2021-03-08T22:11:16.027Z","logger":"visor/lens/lotus","caller":"lens/util.go:27","msg":"Optimized StateGetActorError: Failed to load tipset: handler: websocket connection closed. Falling back to default StateGetActor()."}
{"level":"error","ts":"2021-03-08T22:11:16.027Z","logger":"actorstate","caller":"actorstate/task.go:91","msg":"actor returned with error","height":564300,"actor":"fil/3/reward","address":"f02","error":"failed to extract parsed actor state: RPC client error: sendRequest fa
iled: websocket routine exiting"}
{"level":"error","ts":"2021-03-08T22:11:16.027Z","logger":"schedule","caller":"schedule/scheduler.go:133","msg":"task exited with failure","task":"Watcher","error":"index: notify tipset: context done: context deadline exceeded"}
{"level":"info","ts":"2021-03-08T22:11:16.027Z","logger":"schedule","caller":"schedule/scheduler.go:119","msg":"restarting task","task":"Watcher","delay":60}
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x105258a]

goroutine 183502 [running]:
[...]

PR #412, opened 2021-03-09
fix: guard concurrent accesses to node api

iand:
Following investigation by @placer14 we suspect the panic in #411 is due to concurrent modifications of the node field in the actor state task. When originally written this code was sequential but recent changes have introduced concurrent accesses to the field, namely when the Close method is called. I speculate this can happen when another task fails while an actor state task is still spawning goroutines. The task failure is detected by the indexer which closes all lenses used by the other running tasks, setting the node field to nil which is then passed to a new goroutine by the still-running task.  

This change guards accesses to the node field with mutexes in each of the tasks. The actor state task checks whether the api lens is available before processing and if it isn't returns a non-fatal error for each actor which will be logged in visor_processing_reports for that tipset+task+actor combination.

Also removed the node field from the messages task since it is not used.

Fixes https://github.com/filecoin-project/sentinel-visor/issues/411

### DIFF
--- a/chain/indexer.go
+++ b/chain/indexer.go
@@ -88,7 +88,7 @@ func NewTipSetIndexer(o lens.APIOpener, d model.Storage, window time.Duration, n
 		case BlocksTask:
 			tsi.processors[BlocksTask] = blocks.NewTask()
 		case MessagesTask:
-			tsi.messageProcessors[MessagesTask] = messages.NewTask(o)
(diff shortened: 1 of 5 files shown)


placer14 (approved):

placer14:
Thanks for putting this together, Ian.

frrist (approved):

PR #414, opened 2021-03-10
scheduler job control + watch, walk, and job cli

frrist:
### What
- Add the following commands:
  - `watch` start a watch job
  - `walk` start a walk job
  - `job start` start a job
  - `job stop` stop a job
  - `job list` list all jobs, their status, and config
- Modify Scheduler: controls jobs allowing them to be started and stopped independently.


storage/migrations/27_table_and_column_comments.go
```diff
@@ -181,11 +181,11 @@ func init() {
 	COMMENT ON COLUMN miner_infos.height IS 'Epoch at which this miner info was added/changed.';
 	COMMENT ON COLUMN miner_infos.miner_id IS 'Address of miner this info applies to.';
 	COMMENT ON COLUMN miner_infos.multi_addresses IS 'JSON array of multiaddrs at which this miner can be reached.';
-	COMMENT ON COLUMN miner_infos.new_worker IS 'Address of a new worker address that will become effective at worker_change_epoch.';
```
frrist, line 184: will revert.. editors rename shortcut was a little too enthusiastic

storage/sql_test.go
```diff
@@ -76,7 +76,7 @@ func TestModelUpsert(t *testing.T) {
 		MinerID:                 "minerID",
 		StateRoot:               "stateroot",
 		OwnerID:                 "owner",
-		WorkerID:                "worker",
```
frrist, line 79: ditto above

work/worker.go
```diff
@@ -0,0 +1,151 @@
...
+
+	w, ok := wm.Workers[id]
+	if !ok {
+		return xerrors.Errorf("starting worker ID: %d not found", id)
+	}
+	if w.Running() {
+		return xerrors.Errorf("starting worker ID: %d already running", id)
+	}
+	log.Infow("starting worker", "id", id)
+	w.Start(ctx)
```
frrist, line 74: I'd like to expose a workers ID to the worker (could allow workers to log independently of eachouther), but I can't figure out how to do so cleanly -- maybe its not important but open to suggestions

commands/lily/watch.go
```diff
@@ -17,21 +20,25 @@ var LilyWatchCmd = &cli.Command{
 	Name:  "watch",
 	Usage: "Start, Stop, and List watches",
 	Subcommands: []*cli.Command{
+		LilyWatchCreateCmd,
 		LilyWatchStartCmd,
+		LilyWatchStopCmd,
+		LilyWatchListCmd,
```
iand, line 26: Not sure about this API. It's logical but when writing the design doc I thought we would do something like:

```
visor watch <-- starts a watch
visor walk <-- starts a walk
visor monitor-reorg <- starts monitoring reorgs

visor jobs list <-- lists all the of the above jobs
visor jobs stop <-- stops any of the jobs
visor jobs status <-- gets status of any job
```

I think the main difference is that there is a single way to see what jobs your visor node is running and to stop them if you need to. The API you have here would lose that.

BUT I am not convinced either way is better than others at the moment. What do you think?

work/worker.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// AddWorker adds Worker `w` to WorkerManager's Worker list and returns its ID. Workers can be
+// started and stopped using their ID's.
+func (wm *WorkerManager) AddWorker(w Worker) int {
+	wm.WorkersMu.Lock()
+	defer func() {
+		wm.Workers[wm.WorkerID] = w
+		wm.WorkerID++
+		wm.WorkersMu.Unlock()
```
iand, line 55: I find this defer adds cognitive overhead. Why not simply defer the unlock and then add the worker to the map and increment the id outside the defer?

work/worker.go
```diff
@@ -0,0 +1,151 @@
...
+	}()
+	log.Infow("added worker", "id", wm.WorkerID)
+	return wm.WorkerID
+}
+
+// Start starts the worker with ID `id`, errors if the ID is not found, or id the worker us running.
```
iand, line 61: typos: **if** the worker **is** running

work/worker.go
iand: Could you set the worker ID in `AddWorker`. It it's just for logging then maybe give it a logger with a predefined tag

work/worker.go
```diff
@@ -0,0 +1,151 @@
...
+
+	w, ok := wm.Workers[id]
+	if !ok {
+		return xerrors.Errorf("stopping worker ID: %d not found", id)
+	}
+	if !w.Running() {
+		return xerrors.Errorf("stopping worker ID: %d is not running", id)
+	}
+	log.Infow("stopping worker", "id", id)
+	w.Stop(ctx)
```
iand, line 91: Should there be a timeout here? 

Should we allow `w.Stop` to return an error?

work/worker.go
```diff
@@ -0,0 +1,151 @@
...
+	tw.mu.Lock()
+	tw.running = true
+	tw.err = nil
+	tw.cancel = cancel
+	tw.mu.Unlock()
+}
+
+func (tw *TaskWorker) shutdown(err error) {
+	tw.mu.Lock()
+	tw.cancel()
```
iand, line 130: If `Stop` is called before `Start` then this will panic. Check `tw.running == true` first.

work/worker.go
```diff
@@ -0,0 +1,151 @@
...
+
+func NewWorkerManager(mctx helpers.MetricsCtx, lc fx.Lifecycle) *WorkerManager {
+	return &WorkerManager{
+		WorkerID: 1,
+		Workers:  make(map[int]Worker),
+	}
+}
+
+// WorkerManager controls Workers allowing them to be started and stopped independently of each other.
+type WorkerManager struct {
```
iand, line 25: Suggestion: can you combine `WorkerManager` and `schedule.Scheduler`. The advantage that Scheduler has is that it can restart tasks if they fail. It would need modifying to allow tasks to be added at any time and for them to be stopped by ID. Maybe add a status struct for each one that can hold a list of recent events such as "restarted at 11:03"

We can also get clearer about naming:

 - Task - a single capability of visor, such as extracting block data from a tipset
 - Job - a scheduled/managed set of tasks with configuration (walk is a Job, watch is a Job)
 - Scheduler - responsible for executing Jobs and ensuring they continue to run as specified

(I don't like xxxManager naming)

I think this means that you could rename `schedule.Task` to `schedule.Job` and combine your `Worker` type into Job so it gets the Start and Stop methods. Rename `schedule.TaskConfig` too. It has a name field, perhaps that could be used as the ID that you have for workers.

iand (changes requested):

commands/lily/watch.go
frrist: I'm in favor of your proposal -- it's also a smaller API surface, and easily extendable when/if more job types are added. I believe it will also align well with your suggestion below on [Task/Job/Scheduler](https://github.com/filecoin-project/sentinel-visor/pull/414#discussion_r596151173).

work/worker.go
frrist: > I think this means that you could rename schedule.Task to schedule.Job and combine your Worker type into Job 

Sounds good, I'll adapt this PR to do that and wire it up to watch, with walk coming in a follow-on.

> Maybe add a status struct for each one that can hold a list of recent events such as "restarted at 11:03"

I was thinking of adding Journaling to tasks -- this seems like a good starting point for that.

chain/indexer.go
```diff
@@ -351,6 +351,10 @@ func (t *TipSetIndexer) TipSet(ctx context.Context, ts *types.TipSet) error {
 				defer wg.Done()
 				ctx, _ = tag.New(ctx, tag.Upsert(metrics.TaskType, task))
 
+				// TODO(frrist) when this is the last item being persisted the scheduler will kill
+				// the job by canceling its context, since this operation is performed in a goroutine the
+				// scheduler has no way of knowing a persistence operation is still occurring and thus
+				// this operation fails due to context cancellation. Easy to reproduce using a walk job.
```
iand, line 357: Create an issue so we don't forget?

schedule/scheduler.go
```diff
@@ -14,159 +18,301 @@ import (
...
+	return nil
+}
 
-				// Is the context done?
-				select {
-				case <-ctx.Done():
-					return
-				default:
-				}
+func (s *Scheduler) Stop(id JobID) error {
```
iand, line 197: Rename this to StopJob so it's clearer that it is not stopping the scheduler.

schedule/scheduler.go
```diff
@@ -14,159 +18,301 @@ import (
...
+	job, ok := s.jobs[id]
```
iand, line 198: Need to guard access to s.jobs with a mutex

schedule/scheduler.go
```diff
@@ -14,159 +18,301 @@ import (
...
+			s.workerJobsRunning--
+			if s.workerJobsRunning == 0 {
+				log.Info("no worker jobs running")
 			}
+		}
+	}
+}
+
+func (s *Scheduler) Start(id JobID) error {
+	job, ok := s.jobs[id]
```
iand, line 184: Guard access to s.jobs with mutex

schedule/scheduler.go
```diff
@@ -14,159 +18,301 @@ import (
...
-					return
-				}
-				defer func() {
-					if err := tc.Locker.Unlock(ctx); err != nil {
-						if !errors.Is(err, context.Canceled) {
-							log.Errorw("failed to unlock task", "task", tc.Name, "error", err.Error())
-						}
-					}
-				}()
+	for _, tc := range s.jobs {
```
iand, line 146: Guard access to s.jobs with mutex

schedule/scheduler.go
```diff
@@ -14,159 +18,301 @@ import (
...
+		case <-s.workerJobComplete:
+			s.workerJobsRunning--
+			if s.workerJobsRunning == 0 {
+				log.Info("no worker jobs running")
 			}
+		}
+	}
+}
+
+func (s *Scheduler) Start(id JobID) error {
```
iand, line 183: Rename to StartJob

schedule/scheduler.go
```diff
@@ -14,159 +18,301 @@ import (
...
+	job, ok := s.jobs[id]
+	if !ok {
+		return xerrors.Errorf("starting worker ID: %d not found", id)
+	}
 
-			// Keep this task running forever
-			doneFirstRun := false
-			for {
+	if job.running {
```
iand, line 189: Needs a mutex

schedule/scheduler.go
```diff
@@ -14,159 +18,301 @@ import (
...
-				if doneFirstRun {
-					log.Infow("restarting task", "task", tc.Name, "delay", tc.RestartDelay)
-					if tc.RestartDelay > 0 {
-						time.Sleep(tc.RestartDelay)
-					}
-				} else {
-					log.Infow("running task", "task", tc.Name)
-					doneFirstRun = true
-				}
+	if !job.running {
```
iand, line 203: Needs a mutex - or put it in an `IsRunning()` method on job that guards access with mutex

schedule/scheduler.go
```diff
@@ -14,159 +18,301 @@ import (
...
+		return xerrors.Errorf("starting worker ID: %d already running", id)
+	}
+	job.log.Info("stopping job")
+	if job.cancel != nil {
```
iand, line 207: When is job.cancel ever set to nil? Maybe move changes to job.cancel and job.running onto the job in a `Stop()` method?

schedule/scheduler.go
```diff
@@ -14,159 +18,301 @@ import (
...
+
+func (s *Scheduler) execute(tc *JobConfig, complete chan struct{}) {
+	ctx, cancel := context.WithCancel(s.context)
+	tc.cancel = cancel
+	tc.running = true
+	// Report job is complete when this goroutine exits
+	defer func() {
+		complete <- struct{}{}
+		tc.running = false
+		tc.cancel()
```
iand, line 253: cancel can be nil? You checked for it earlier. 

Probably best to set to nil once the job has completed.

iand (commented):
Added some review comments. Will do some more on Monday or maybe over weekend if I am around.

schedule/scheduler.go
frrist: It can never be `nil`, all jobs are stated via the execute method and a cancelFunc is always assigned to a job at the beginning of the method. I've dropped the error check as it's not needed. I can't come up with a strong reason for setting cancel to nil, what do you think?

schedule/scheduler.go
```diff
@@ -14,159 +18,305 @@ import (
...
-						}
-					}
-				}()
+	s.jobsMu.Lock()
+	for _, tc := range s.jobs {
+		go s.execute(tc, s.scheduledJobComplete)
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
iand, line 145: This will leave the mutex locked

iand (commented):
Still some locking issues that need to be looked at

schedule/scheduler.go
frrist: I have addressed the comment [here](https://github.com/filecoin-project/sentinel-visor/pull/414/files#r599600777) by removing the locks previously taken before this loop. They are not required since jobs cannot be access concurrently at this point. Thoughts?

schedule/scheduler.go
frrist: see note here: https://github.com/filecoin-project/sentinel-visor/pull/414#discussion_r599755832

frrist:
Seeing some tests flake wrt database operations (should be unrelated to changes here) made this issue to track CI flakes: https://github.com/filecoin-project/sentinel-visor/issues/438

schedule/scheduler.go
iand: I'll need to re-review to follow the new logic.

schedule/scheduler.go
```diff
@@ -14,159 +18,309 @@ import (
...
-					log.Infow("task exited cleanly", "task", tc.Name)
-
-					if !tc.RestartOnCompletion {
-						// Exit the task
-						break
-					}
+func (s *Scheduler) execute(tc *JobConfig, complete chan struct{}) {
+	ctx, cancel := context.WithCancel(s.context)
+	tc.cancel = cancel
+	tc.running = true
```
iand, line 255: Need to look at the locking here. tc.running is written without a lock but up at line 242 it's also read by a function that could be in a different goroutine. `go test -race` confirms that there is a race here. I also points to tc.cancel being racy probably because you call it in StopJob.

iand (approved):
OK to merge to lily once the data races are fixed.

schedule/scheduler.go
frrist: Race's fixed in latest commit:
```
┬─[frrist@cypress:~/s/g/f/sentinel-visor]─[10:42:42 AM]─[G:frrist/watch-workers]
╰─>$ git rev-parse HEAD
0f83f196a152788501d6516f681c98d7a6629728

┬─[frrist@cypress:~/s/g/f/sentinel-visor]─[10:42:44 AM]─[G:frrist/watch-workers]
╰─>$ go test -count=1 -race ./schedule/
ok      github.com/filecoin-project/sentinel-visor/schedule     0.042s
```

PR #420, opened 2021-03-12
fix: embed genesis block in executable

### LINKED ISSUE #419: Embed genesis block in executable
### Description

Some lens types use Lotus functions that require access to the genesis block. This is present in the Lotus source but should be statically embedded into the Visor executable to make it readable without access to the Lotus source tree.

iand:
Fixes #419


iand:
Lint failure is fixed in #421

thattommyhall:
I can confirm this enables running sqlotus from the docker image

hsanjuan (approved):

PR #422, opened 2021-03-12
chore: reuse watcher in lily watch command

iand:
This change makes `chain.Watcher` independent of the lens opener interface so it can be reused by lily commands and the existing direct mode. It's a bit verbose at the moment (and really only saves ~6 lines of duplication) but it steers us toward a pattern where jobs started by lily are essentially the same as those started using the direct watch/walk commands. We should start considering those direct commands as single-job execution modes that terminate the visor process when complete, whereas lily is a multi-job mode with no termination.

After this change `chain.Watcher` no longer depends on the lens and opener abstraction.

This is also the start of reducing our dependency on the Lotus API. Using the API will be a restricted use-case where visor doesn't have direct access to a blockstore. We will rework the lens interface to support more efficient access patterns that assume a local blockstore and the Lotus API lens implementation will not support some of those patterns in the future.


lens/lily/impl.go
```diff
@@ -124,3 +130,96 @@ func setupDatabase(ctx context.Context, cfg *LilyDatabaseConfig) (*storage.Datab
...
+	h.mu.Lock()
+	if h.err != nil {
+		h.mu.Unlock()
+		return
+	}
+	h.err = err
+	if h.events == nil {
+		h.events = make(chan *chain.HeadEvent)
+	}
+	close(h.events)
```
frrist, line 172: This seems odd, why make the channel here if we're just going to close it immediately?

iand: It's so calling cancel always results in a closed channel. A nil channel will always block in a select, but a closed one will be chosen. This means that we can have a goroutine waiting on `HeadEvents()` and as soon as the notifier is cancelled that goroutine will be notified, and this will work even if the cancel happens before the wait.

commands/watch.go
```diff
@@ -95,3 +112,105 @@ func watch(cctx *cli.Context) error {
 	}
 	return nil
 }
+
+// LotusChainNotifier is a head event notifier that subscribes to a lens's ChainNotify method and adapts the
+// events received for use by a chain.Watcher
+// NOTE: this functionality will be probably folded into the Lotus API lens since other lenses will support more
+// direct methods of accessing new tipsets
```
frrist, line 119: Does this mean here: https://github.com/filecoin-project/sentinel-visor/blob/master/lens/lotus/api.go#L1

or somewhere different?

frrist (approved):

commands/watch.go
iand: Roughly there - I am still exploring

PR #427, opened 2021-03-19
Disable preloader by default (toggle with CLI flag), couple library updates

ribasushi:
This is a temporary thing to get @thattommyhall over the hump

Tangentially: @iand @hsanjuan can I haz sentinel-visor push perms so I can stop priv-forking?

### DIFF
--- a/lens/sqlrepo/repo.go
+++ b/lens/sqlrepo/repo.go
@@ -3,7 +3,6 @@ package sqlrepo
 import (
 	"context"
 	"fmt"
-	"os"
 
 	"github.com/urfave/cli/v2"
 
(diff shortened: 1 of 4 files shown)


hsanjuan:
I don't have admin

lens/sqlrepo/repo.go
```diff
@@ -21,6 +22,11 @@ func NewAPIOpener(c *cli.Context) (lens.APIOpener, lens.APICloser, error) {
 		LogCacheStatsOnUSR1:     true,
 	}
 
+	enablePreload := os.Getenv("LOTUS_CHAINSTORE_PRELOAD_RECENTS")
```
iand, line 25: Can you use `c.String("lens-postgres-preload")` so it's consistent with other usages here and then add it to the cli command. Trying to avoid lots of unduscoverable env variables - adding them to the cli command means the user can see them with `--help`

iand (changes requested):

iand:
I don't have privileges to add you I'm afraid

iand (approved):

ribasushi:
@iand @willscott I have forgotten to actually list the added option, one extra commit on top of what's already approved

ribasushi:
This was squash-merged as part of https://github.com/filecoin-project/sentinel-visor/pull/430

PR #432, opened 2021-03-22
merge master into lily

frrist:


iand:
Didn't we just do this? Why do I see the same commits like "vector benchmarking" that I thought we merged before?

frrist:
The changes look right, but the commits do seem a bit odd..

iand:
LGTM but perhaps wait for your recent fixes to be merged and include those too?

iand (approved):

frrist:
Yup, will be sure to include them here before merging into lily.

PR #435, opened 2021-03-23
chore: run vector tests in parallel

frrist:
Reduces vector test runtime from ~1.5min to ~25sec


frrist:
CI doesn't have enough resources to run these tests in parallel :disappointed:  Couple of options:
- Don't run in parallel; 1.5mins isn't _that_ long.
- Increase CI instance size
- Add a flag to run vector tests in its own step.:

PR #436, opened 2021-03-23
chore: increase linter timout in ci

frrist:
Its been timing out a lot lately.. :disappointed:


frrist:
Example of timeout while linting: https://github.com/filecoin-project/sentinel-visor/pull/433

iand (approved):

PR #445, opened 2021-03-26
chore: reorganise visor commands

iand:
This simplifies the cli subcommands so that the daemon mode is presented as the primary way of interacting with visor.

```
COMMANDS:
   daemon   Start a visor daemon process.
   job      Manage jobs being run by the daemon.
   migrate  Reports and verifies the current database schema version and latest available for migration. Use --to or --latest to perform a schema migration.
   run      Run a single job without starting a daemon.
   vector   Vector tooling for Visor.
   watch    Start a daemon job to watch the head of the filecoin blockchain.
   walk     Start a daemon job to walk a range of the filecoin blockchain.
```

The current non-daemon walk and watch commands are now subcommands of run. 

So now `visor watch` will attempt to start a watch job against a running daemon. To get the current behaviour you would use `visor run watch` which will start a watch in visor without starting a daemon.

I also removed environment variables for the commands that start daemon jobs since I don't think they will be useful in that context.


iand:
Branched from #444 so it includes that commit

frrist (approved):

iand:
Ugh. Can't get rid of that conflict

commands/daemon.go
```diff
@@ -26,58 +25,59 @@ import (
 	"github.com/filecoin-project/sentinel-visor/schedule"
 )
 
-var log = logging.Logger("lily-cli")
-
 type daemonOpts struct {
 	api            string
 	repo           string
-	bootstrap      bool
+	bootstrap      bool // TODO: is this necessary - do we want to run visor in this mode?
```
frrist, line 31: Running with bootstrap disabled will prevent Visor from connecting to peers (and therefore syncing the chain). It could prove useful in some future testing scenarios, but I'd be alright if we dropped this

PR #450, opened 2021-03-31
feat: Add --api parameter to daemon client commands

iand:


commands/job.go
```diff
@@ -22,60 +22,93 @@ var JobCmd = &cli.Command{
...
+	Usage: "start a job.",
 	Flags: []cli.Flag{
 		&cli.IntFlag{
 			Name:        "ID",
 			Usage:       "ID of job to start",
 			Required:    true,
 			Destination: &jobControlFlag.ID,
 		},
+		&cli.StringFlag{
+			Name:        "api",
```
frrist, line 42: An alternative to this is getting the apiAddr from the repo (we set it [here](https://github.com/filecoin-project/sentinel-visor/blob/lily/commands/daemon.go#L57)), similar to what's done here: 
https://github.com/filecoin-project/lotus/blob/c41777dcd2aea4e32405b7c1f548c7621cd97cd4/cli/util/api.go#L110-L133

frrist (approved):

commands/job.go
iand: That wouldn't work if the client was on a different computer though.

Issue #453, opened 2021-04-01
Move parameters from main to subcommands


iand:
### Description

Many parameters are currently specified on the `visor` command but are only relevant for non-daemon watches or walks. We should make them common parameters to the `visor run walk` and `visor run watch` commands. 

Instead of:

```
visor --api=blah --db=bleugh --lens=lotus run watch
```

we should have

```
visor run watch --api=blah --db=bleugh --lens=lotus
```

### Acceptance criteria

The following parameters should be usable from `visor run watch` and `visor run walk`

 - lens
 - repo
 - repo-read-only
 - api
 - db
 - db-pool-size
 - db-allow-upsert
 - allow-schema-migration
 - lens-cache-hint
 - name
 - lens-postgres-namespace
 - lens-postgres-preload-recents
 - lens-postgres-get-prefetch-depth

The following parameter should be usable from `visor migrate`

 - db

frrist:
closed via https://github.com/filecoin-project/sentinel-visor/pull/464

PR #464, opened 2021-04-14
feat: clean up command line flags for single-run jobs

iand:
Builds on #445 to better segregate flags so they only appear on relevant commands. 

The `visor run walk` and `visor run watch` commands now expect their database and lens flags to be specified after the subcommand, instead of after the visor command. This means that where previously we would use:

```
visor --db=foo --lens=lotus walk --from=1000 --to=1001
```

we will now use:

```
visor run walk --db=foo --lens=lotus --from=1000 --to=1001
```

The `migrate` command now also expects flags to be specified after the command.

Some consistency related changes:

 - the `--api` flag has been renamed to `--lens-lotus-api` since it is used by the lotus lens to specify the api that visor will connect to. This reduces confusion from the daemon-oriented `--api` flag which specifies the api address of visor's daemon.
 - the `--repo` flag has been renamed to `--lens-repo` since it is used by various lenses to specify the location of the data file or directory that visor will read from. This distinguishes it from the daemon-oriented `--repo` flag which specifies the path where the visor daemon should write its data.

### DIFF
--- a/commands/run.go
+++ b/commands/run.go
@@ -1,8 +1,23 @@
 package commands
 
 import (
+	"fmt"
+	"os"
+
 	"github.com/urfave/cli/v2"
+
+	"github.com/filecoin-project/sentinel-visor/version"
 )
+
+var defaultName = "visor"
+
+func init() {
+	defaultName := "visor_" + version.String()
+	hostname, err := os.Hostname()
+	if err == nil {
+		defaultName = fmt.Sprintf("%s_%s_%d", defaultName, hostname, os.Getpid())
+	}
+}
 
 var RunCmd = &cli.Command{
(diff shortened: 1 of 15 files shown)


commands/run.go
```diff
@@ -1,9 +1,24 @@
...
+
 	"github.com/urfave/cli/v2"
+
+	"github.com/filecoin-project/sentinel-visor/version"
 )
 
+var defaultName = "visor"
+
+func init() {
+	defaultName := "visor_" + version.String()
```
frrist, line 15: assignment should fix the linting error here.

frrist (approved):

frrist:
This closes #453, yea?

PR #476, opened 2021-04-20
feat: upgrade to actors v4

iand:
Merge after v0.6.0 is tagged

This is just a preliminary integration with the v4 types, will need another update once actors v4 is released


placer14 (approved):
I can't validate completeness of the unit tests, but what I can see looks good! 👍

frrist (approved):

PR #481, opened 2021-04-21
fix: docker builds to include hwloc dep

placer14:
Noticed CI was broken on master. This has a few improvements.

Fixes:
- fix dev docker build to include hwloc dep ([example](https://app.circleci.com/pipelines/github/filecoin-project/sentinel-visor/1728/workflows/91dba3b9-42a5-40fc-9ad9-6a564c15d5bc/jobs/3798))
- remove vector data from `make dep` target (and from docker images) and add vector data artifacts to `make clean`. Vector data now available as `make vector-setup`.
- remove `||` on `make testfull` hack

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,8 @@ FROM golang:1.15.2 as builder
 ENV GOROOT=/usr/local/go
 
 # Install deps for filecoin-project/filecoin-ffi
-RUN apt-get update && apt-get install -y jq mesa-opencl-icd ocl-icd-opencl-dev hwloc libhwloc-dev
+RUN apt-get update
+RUN apt-get install -y jq mesa-opencl-icd ocl-icd-opencl-dev hwloc libhwloc-dev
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -3,7 +3,8 @@ FROM golang:1.15.2 as builder
 ENV GOROOT=/usr/local/go
 
 # Install deps for filecoin-project/filecoin-ffi
-RUN apt-get update && apt-get install -y jq mesa-opencl-icd ocl-icd-opencl-dev hwloc libhwloc-dev
+RUN apt-get update
```
iand, line 6: I don't think we should split these lines, otherwise the update will be cached separately to the install. This caused a problem for me when adding hwloc since some of the cached apt repositories had changed and threw errors.

iand (approved):

Dockerfile
placer14: Good callout. 🤝

PR #483, opened 2021-04-21
chore: add circleci step to test docker build

iand:
Unfortunately makes builds a lot slower, but more reliable


iand:
Failure is expected since #481 fixes the dev docker build

placer14 (approved):

PR #487, opened 2021-04-21
fix(scheduler): deadlock on job specific lock in start

### LINKED ISSUE #454: Possible deadlock or hang in lily
### Describe the bug:

Running `visor watch` hung and also stopped the visor node from syncing

### Steps to Reproduce:

This is what I did:

 - started the daemon
 - started a watch
 - stopped the watch
 - started another watch with a storage
 - stopped the watch
 - started another watch with invalid storage name which fails as expected
 - started another watch with a storage

The command hung and so did the visor daemon. I waited about 10 minutes and then had to kill the daemon manually. Ctrl+C on the watch command didn't work.

frrist:
- fixes #454

### DIFF
--- a/schedule/scheduler.go
+++ b/schedule/scheduler.go
@@ -196,6 +196,7 @@ func (s *Scheduler) Run(ctx context.Context) error {
 func (s *Scheduler) StartJob(id JobID) error {
 	s.jobsMu.Lock()
 	defer s.jobsMu.Unlock()
+
 	job, ok := s.jobs[id]
 	if !ok {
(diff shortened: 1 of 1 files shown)


schedule/scheduler.go
```diff
@@ -205,6 +206,7 @@ func (s *Scheduler) StartJob(id JobID) error {
 	// clear any error messages if we are starting the job
 	job.errorMsg = ""
 	if job.running {
+		job.lk.Unlock()
```
iand, line 209: Could just defer after line 205. Not much else happens in this function so it won't be locked for long

iand (approved):

PR #489, opened 2021-04-22
perf: remove redundant tipset lookups

iand:
Minor performance optimization.

When extracting actor states we already have the full tipset and its parent available so we can avoid calling the chain store to look up the tipset from its key. All the Lotus APIs do this too so we could further optimize by implementing alternate calls that accept a full tipset - I noted some of these in the ActorStateAPI interface.

Also remove calls to StateGetActor (which did an additional load of the tipset from its key) since the actor info is already passed to each extractor.

This allows us to shrink the ActorStateAPI interface by removing the ChainGetTipSet method. I also removed the ChainHasObj and ChainReadObj methods which were unused.


frrist (approved):
Good stuff!

PR #491, opened 2021-04-27
fix(task): message task tracks unexecuted messages and their blocks

### LINKED ISSUE #426: Possible missing messages from blocks
### Describe the bug:
[Initial Slack Thread for context
](https://protocollabs.slack.com/archives/CT2265RFZ/p1616087045025000)

Message CID `bafy2bzaceac36ranzqs5jg66b4qlchhcufpxkiai6e7lc3l25cq2sftcl7ziu` appears in blocks:
`bafy2bzacecmca4wktk4vclyfyifem57v55hotbuaw4i3ojrpcrny675jggrkq` (Latest 238 Messages)
`bafy2bzacebmfagqhkl73syvvxlynk3tdbmdhh3yn2gwggulttjczasov345ne` (Latest 238 Messages)
`bafy2bzacebpix3zkybskexmpjyb5wfcx7q4pdnb5ijv7zwt4w3ybhxmewfbyi` (Latest 240 Messages)

Source FilFox: https://filscan.io/tipset/message-detail?cid=bafy2bzaceac36ranzqs5jg66b4qlchhcufpxkiai6e7lc3l25cq2sftcl7ziu

This messages is not found in Sentinel Production Monitoring and the aforementioned blocks have a different number of messages than FilFox:

`select count(message) from block_messages where block = 'bafy2bzacebpix3zkybskexmpjyb5wfcx7q4pdnb5ijv7zwt4w3ybhxmewfbyi' and height = '590662'`
Count: 239

`select count(message) from block_messages where block = 'bafy2bzacebmfagqhkl73syvvxlynk3tdbmdhh3yn2gwggulttjczasov345ne' and height = '590662'`
Count: 236

`select count(message) from block_messages where block = 'bafy2bzacecmca4wktk4vclyfyifem57v55hotbuaw4i3ojrpcrny675jggrkq' and height = '590662'`
Count: 236

### Visor Version: Sentinel Production Monitoring

frrist:
### What
Fixes #426 by modifying and renaming `GetExecutedMessagesForTipset` -> `GetExecutedAndBlockMessagesForTipset` such that it returns a list of blocks and their corresponding messages, called `BlockMessages`, in addition to the executed messages. The `MessageProcessor` consumes the `BlockMessages` and populates the `block_messages` model with all messages in each block regardless of whether or not the message was executed (has a receipt) (this is the fix for #426). The `MessageProcessor` also populates the `message` model with a deduplicated set of aforementioned messages -- this ensures all messages from `block_messages` have an entry in `messages` model. 

### Status
If this change gets a :+1: I will regenerate the test vectors since this PR breaks them by changing the data read by the message processor and the expected result.

### DIFF
--- a/chain/indexer.go
+++ b/chain/indexer.go
@@ -216,12 +216,12 @@ func (t *TipSetIndexer) TipSet(ctx context.Context, ts *types.TipSet) error {
 
 			// If we have message processors then extract the messages and receipts
 			if len(t.messageProcessors) > 0 {
-				emsgs, err := t.node.GetExecutedMessagesForTipset(ctx, child, parent)
+				emsgs, blkMsgs, err := t.node.GetExecutedAndBlockMessagesForTipset(ctx, child, parent)
 				if err == nil {
 					// Start all the message processors
 					for name, p := range t.messageProcessors {
 						inFlight++
-						go t.runMessageProcessor(tctx, p, name, child, parent, emsgs, results)
+						go t.runMessageProcessor(tctx, p, name, child, parent, emsgs, blkMsgs, results)
 					}
 				} else {
 					ll.Errorw("failed to extract messages", "error", err)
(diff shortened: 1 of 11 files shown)


iand (approved):
Cool

iand:
Can we hold merging this until after 0.6.2 since we need to test it in staging

hsanjuan:
@frrist I understand that with this we will now be ingesting all block messages, instead of just executed ones. Is there a way we are tracking which ones were actually executed?

I understand the normal case is that all of them executed. Sometimes a message and its replacement are included, in which case only one is executed. I don't know if there is another case when an included messages is not executed.

A different way of making the question: how could I graph, after this, the number of messages in a block that did not execute?

frrist:
> I understand that with this we will now be ingesting all block messages, instead of just executed ones.

Correct 👍, messages are only ever ignored (not executed) because an earlier block in the tipset contained a msg with the same sender+nonce -- as was the case in #426. Executed Messages will have receipts whereas unexecuted messages do not. 

> how could I graph, after this, the number of messages in a block that did not execute?

These would be messages without a receipt. Joining on the Receipts table could be expensive (or just annoying to do), we could add an `executed` column to the message table with `true/false` values? Thoughts?

hsanjuan:
> These would be messages without a receipt. Joining on the Receipts table could be expensive (or just annoying to do), we could add an `executed` column to the message table with `true/false` values? Thoughts?

In practice, we don't generally use messages JOIN receipts, and use derived_gas_values directly since it contains most of the info. Would non executed messages have entries in derived_gas_values? I'm guessing not so things should be fine.

hsanjuan:
This can be merged now @iand @frrist ?

iand:
OK to merge from a logistic point of view, but I think @frrist wanted to change something else in it, but I don't remember the details.

frrist:
Nope, all good on my end. :steam_locomotive:

PR #506, opened 2021-05-06
Replace lotus actor shims

frrist:
This PR use actor shim code-gen from lotus to generate actor shims specific to visor.


jaceppo (approved):

chain/actors/builtin/README.md
```diff
@@ -0,0 +1,29 @@
+# Actors
```
frrist, line 1: This file was copied over from lotus: https://github.com/filecoin-project/lotus/blob/master/chain/actors/builtin/README.md

Everything else in this PR was created via code gen.

chain/actors/agen/main.go
```diff
@@ -0,0 +1,19 @@
+package main
+
+// TODO uncomment when https://github.com/filecoin-project/lotus/pull/6301 merges
```
frrist, line 3: Filed https://github.com/filecoin-project/lotus/pull/6301 that allows Visor to import it here. Until it merges this will need to remain commented out, but that doesn't need to block merging this.

iand (commented):
Looks good in general But:

How can the shims be regenerated? What is the command. Can you add it to README and Makefile?

Can you add the following line to each template, as the first line in the output, leaving a line between it and the package declaration:

`// Code generated by: <instruction for generating>. DO NOT EDIT.`

(as per https://golang.org/s/generatedcode)

frrist:
Added note to actors README, a command in the Makefile, and a comment about code gen in the template files.

PR #512, opened 2021-05-12
chore: shorten indexhead-confidence option name on watch command

iand:


commands/watch.go
```diff
@@ -42,7 +42,7 @@ var WatchCmd = &cli.Command{
 	Usage: "Start a daemon job to watch the head of the filecoin blockchain.",
 	Flags: []cli.Flag{
 		&cli.IntFlag{
-			Name:        "indexhead-confidence",
+			Name:        "confidence",
```
placer14, line 45: You'll need to coordinate this change in the helm chart: https://github.com/filecoin-project/helm-charts/blob/master/charts/visor/templates/deployment.yaml#L90. Maybe include a conditional error that throws if the requested version of visor doesn't work with the new chart changes?

placer14 (approved):
LGTM, with thoughtful consideration around updating the helm chart and coordinating deployments to not accidentially use the wrong version of the chart with the wrong version of visor.

commands/watch.go
iand: This is just for the lily version of the command, the standard one is left unchanged

PR #513, opened 2021-05-13
feat: allow database urls to use environment variables in config

iand:
This allows the specification of connection information without embedding secrets in config files.


placer14 (approved):
I like the configuable envvar keys. LGTM 🤝

PR #521, opened 2021-05-20
fix: fail parsed message task for unknown actor methods

iand:


hsanjuan (approved):
I have a question. Can someone post messages to unknown actors/methods and these be included in blocks?

Is there perhaps a way of resolving this by breaking at compile time vs run time? I.e. checking that statediff actors version == ours

iand:
> I have a question. Can someone post messages to unknown actors/methods and these be included in blocks?
> 

In theory yes. This will only fail the extraction of the single message that could not be parsed. Other messages and other methods will be unaffected. There will be an error message recorded in the database for that epoch.

> Is there perhaps a way of resolving this by breaking at compile time vs run time? I.e. checking that statediff actors version == ours

Yes, I think so but I don't know yet how hard it would be to write a test. I feel it would need to enumerate all the actors and their methods and unless that is automatic it would need to be maintained. Which just moves the existing problem.

hsanjuan:
> In theory yes. This will only fail the extraction of the single message that could not be parsed. Other messages and other methods will be unaffected. There will be an error message recorded in the database for that epoch.

This means we are likely to encounter messages with unknown methods and this breaks message parsing for the whole epoch, which is also suboptimal because it is not a bug in our code per se.

iand:
> > In theory yes. This will only fail the extraction of the single message that could not be parsed. Other messages and other methods will be unaffected. There will be an error message recorded in the database for that epoch.
> 
> This means we are likely to encounter messages with unknown methods and this breaks message parsing for the whole epoch, which is also suboptimal because it is not a bug in our code per se.

Other messages in the epoch should parse fine. The tipset is marked as INFO which means at least one thing failed but any data extracted should still be persisted.

PR #525, opened 2021-05-21
Fix 495: limit what market_deal_states tracks.

hsanjuan:
See comment in code.

This is a very practical approach (avoid ingesting rows that mean little). If there is a reason why it would be minimally interesting to see when market-cron ticks and updates deals, or if we are to be purists ("ingest everything because the deal state does change") then we should not go ahead with this.

### DIFF
--- a/tasks/actorstate/market.go
+++ b/tasks/actorstate/market.go
@@ -238,6 +238,25 @@ func ExtractMarketDealStates(ctx context.Context, ec *MarketStateExtractionConte
 		idx++
 	}
 	for _, mod := range changes.Modified {
+		// The market actor cron updates deal's LastUpdatedEpoch on
+		// every run:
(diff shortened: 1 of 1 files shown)


frrist (approved):
This information was tracked in an anticipatory attempt to have data ready before it was requested. I'm good with dropping it now that we understand it is no longer needed/valuable at this granularity.

hsanjuan:
I wonder if we can track everything from deal_events and directly not use this table. I will check

frrist:
@hsanjuan is there anything blocking this from being merged?

hsanjuan:
> @hsanjuan is there anything blocking this from being merged?

Yes:

>  I will check

I haven't checked

iand:
@hsanjuan any update on this?

PR #527, opened 2021-05-21
feat: introduce schema v1

iand:
Starting with a dump of existing database schema, will refine 

Requires #518


iand:
Started a new branch

PR #529, opened 2021-05-26
polish(tracing): add more tracing calls to critial path

frrist:
Adds more tracing calls for finding performance bottlenecks.

### DIFF
--- a/chain/actors/builtin/miner/diff.go
+++ b/chain/actors/builtin/miner/diff.go
@@ -3,17 +3,23 @@ package miner
 import (
 	"context"
 
+	cbg "github.com/whyrusleeping/cbor-gen"
+	"go.opentelemetry.io/otel/api/global"
+
 	"github.com/filecoin-project/go-amt-ipld/v3"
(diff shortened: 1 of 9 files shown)


chain/actors/builtin/miner/diff.go
```diff
@@ -131,8 +145,8 @@ func DiffSectors(ctx context.Context, store adt.Store, pre, cur State) (*SectorC
 		return diffContainer.Results, nil
 	}
 	changes, err := diff.Amt(ctx, pres, curs, store, store, amt.UseTreeBitWidth(uint(preBw)))
-	if err != nil {
-		return nil, err
+	if span.IsRecording() {
```
iand, line 148: It looks like you've deleted the error checking code here

iand (approved):
LGTM if the error handling line is fixed

chain/actors/builtin/miner/diff.go
frrist: oooffff I did in a rebase, great catch!

PR #530, opened 2021-05-26
feat(statetree): implement optimized state tree diffing

frrist:
optimizes StateChangedActors by using the hamt diffign logic falling back to the lotus method when required, see code comments for specifics.

### DIFF
--- a/chain/indexer.go
+++ b/chain/indexer.go
@@ -1,15 +1,20 @@
 package chain
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
 	"sync"
 	"time"
 
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-hamt-ipld/v3"
 	"github.com/filecoin-project/lotus/chain/types"
 	sa0builtin "github.com/filecoin-project/specs-actors/actors/builtin"
 	sa2builtin "github.com/filecoin-project/specs-actors/v2/actors/builtin"
 	sa3builtin "github.com/filecoin-project/specs-actors/v3/actors/builtin"
 	sa4builtin "github.com/filecoin-project/specs-actors/v4/actors/builtin"
+	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
 	"go.opencensus.io/stats"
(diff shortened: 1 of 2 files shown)


chain/indexer.go
```diff
@@ -253,7 +258,7 @@ func (t *TipSetIndexer) TipSet(ctx context.Context, ts *types.TipSet) error {
 
 			// If we have actor processors then find actors that have changed state
 			if len(t.actorProcessors) > 0 {
-				changes, err := t.node.StateChangedActors(tctx, parent.ParentState(), child.ParentState())
+				changes, err := t.stateChangedActors(tctx, parent.ParentState(), child.ParentState())
```
frrist, line 261: From the tracing added in subsequent PR's I've observed this optimized call reducing the runtime of the method from ~17sec to ~3sec.

chain/indexer.go
```diff
@@ -408,6 +413,81 @@ func (t *TipSetIndexer) runProcessor(ctx context.Context, p TipSetProcessor, nam
...
+		if span.IsRecording() {
+			span.SetAttribute("diff", "fast")
+		}
+		changes, err := hamt.Diff(ctx, t.node.Store(), t.node.Store(), oldRoot, newRoot, hamt.UseTreeBitWidth(5), hamt.UseHashFunction(func(input []byte) []byte {
+			res := sha256.Sum256(input)
+			return res[:]
+		}))
+		if err != nil {
+			log.Errorw("failed to diff state tree efficiently, falling back to slow method", "error", err)
+		} else {
```
frrist, line 450: In hindsight, implementing a state tree shim and using here could make this less :spaghetti:.

chain/indexer.go
iand: Very nice

iand (approved):

chain/indexer.go
frrist: Probably a day or two worth of effort (including review). I would start by extracting a modified version of the [lotus StateTree](https://github.com/filecoin-project/lotus/blob/master/chain/state/statetree.go#L159) to visor and implement some code gen, similar to the actor-shims, to keep it up to date with the specs-actors actor-states version.

PR #536, opened 2021-06-07
fix: fall back to using config in repo directory if unspecified

iand:
I thought visor wasn't detecting the storages in the config file but it turned out it just couldn't find the config. It's fine if you specify it with --config but I think it's also useful to default it to looking in the root of the repo.


frrist (approved):
<3

PR #538, opened 2021-06-07
polish(log): add logging command and sane default log levels

frrist:
### Wat:
- Prefix all visor related logs with "visor-"
- Modify setup logs to set all logs with prefix "visor-" to level `info` and all other logs to level `error` (shows user what visor is doing while only allowing errors from the lotus daemon to come through)
- Copy logging commands over from lotus that allows loggers to be listed (`log list`) and logging systems to be controlled (`log set-level --system chain "error"`)


storage/null.go
```diff
@@ -13,11 +13,9 @@ type NullStorage struct {
...
 
 func (*NullStorage) PersistBatch(ctx context.Context, p ...model.Persistable) error {
-	log.Debugw("Not persisting data")
 	return nil
 }
 
 func (*NullStorage) PersistModel(ctx context.Context, m interface{}) error {
-	log.Debugw("Not persisting data")
 	return nil
 }
```
frrist, line 21: These are pretty spammy... I wish there was something lower than "debug", like "trace". I don't want to see these but I will add them back if there are strong feelings about keeping them.

iand:
So happy you did this. I had something hacked in to my branch to make the logs more useful. One thing to consider is whether logger names should be `visor/*` instead of `visor-*` - there are some existing loggers in visor or lotus that use the slash convention.

frrist:
I guess `visor/*` would be more consistent -- I'll make that change before merging. 

It would be pretty neat if the logging library (or maybe the linter?) could enforce this pattern, I think being consistent about this will be difficult for a developer to remember.

iand:
I prefer [logr](https://github.com/go-logr/logr) style logging which allows you to have subloggers etc. I maintain [one implementation](https://github.com/iand/logfmtr) and maybe I ought to write an IPFS logging bridge.

iand (approved):
LGTM with the change to use `/` as delimiter. Note we also have command line option for setting the log levels - I assume this overrides that when running as daemon?

PR #541, opened 2021-06-09
chore: remove dependency on statediff

iand:
github.com/filecoin-project/statediff has some dependencies on specific versions of ipld that are impeding upgrading to a newer version of lotus. Also it required us to use a temporary shim in a branch of go-hamt-ipld/v2 which doesn't exist for go-hamt-ipld/v3 which we need for our new diffing logic.

We currently only use statediff for message parsing so this PR imports the parts we need and removes the dependency.


iand:
Not sure why ffi broke in the docker container

frrist (approved):
Assuming the bulk of these changes are copy-pasted from statediff? LGMT :+1:

PR #544, opened 2021-06-10
chore: update to latest lotus

iand:
Note that this doesn't contain actors v5 updates yet.

Upgrades the minimum version of Go to 1.16 since the genesis block is now embedded using the native `go:embed` directive.


iand:
Squashed commits to help rebasing the v1 schema changes

frrist (approved):
👍 LGTM

PR #554, opened 2021-06-17
chore: add support for v5 actors

iand:
Update to actors v5-rc3. The actor methods and states should not change between rc and release so there should not be any difficulty on final upgrade.

Update to latest statediff which I have updated to v5 too ([see here](https://github.com/filecoin-project/statediff/pull/225)). We only use statediff for generating message parameter type descriptions.

We should be good to parse the new miner messages and the signature change for ComputeDataCommitment.

I reworked a lot of places where we were importing codes from each actors version to use our generated actor shims and added some helper methods to the templates.

(Ideally I would like to merge the statediff message parameter generation and the actor shim generation into a single package but this is not the best time to do that.)


chain/actors/agen/generator/gen.go
```diff
@@ -83,7 +84,7 @@ func generateAdapters() error {
 				return err
 			}
 
-			if err := ioutil.WriteFile(filepath.Join(actDir, fmt.Sprintf("%s.go", act)), b.Bytes(), 0666); err != nil {
+			if err := ioutil.WriteFile(filepath.Join(actDir, fmt.Sprintf("%s.go", act)), b.Bytes(), 0o666); err != nil {
```
frrist, line 87: curious what the difference is here?

frrist (approved):
Looks good! Can we add https://github.com/filecoin-project/sentinel-visor/pull/555 in with this?

chain/actors/agen/generator/gen.go
iand: It's a consistent style thing from gofumpt or gofmt.

PR #566, opened 2021-06-22
fix: give head notifier a buffer

iand:
Converting head notifier channel into a buffered channel decouples the event subsystem from the indexer. Since this change I haven't seen the node fall out of sync.

### DIFF
--- a/chain/indexer.go
+++ b/chain/indexer.go
@@ -502,7 +502,7 @@ func (t *TipSetIndexer) stateChangedActors(ctx context.Context, old, new cid.Cid
 			return out, nil
 		}
 	}
-	log.Debugf("using slow state diff")
+	log.Debug("using slow state diff")
 	return t.node.StateChangedActors(ctx, old, new)
 }
 
--- a/chain/indexer.go
+++ b/chain/indexer.go
@@ -456,6 +456,7 @@ func (t *TipSetIndexer) stateChangedActors(ctx context.Context, old, new cid.Cid
 		if span.IsRecording() {
 			span.SetAttribute("diff", "fast")
 		}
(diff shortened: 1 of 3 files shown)


chain/indexer.go
```diff
@@ -460,6 +469,7 @@ func (t *TipSetIndexer) stateChangedActors(ctx context.Context, old, new cid.Cid
 			return out, nil
 		}
 	}
+	log.Debugf("using slow state diff")
```
frrist, line 472: should just be Debug, not sure if go-log will complain about the lack of print directives.

chain/indexer.go
```diff
@@ -560,10 +580,12 @@ func (t *TipSetIndexer) Close() error {
...
 type TaskResult struct {
-	Task   string
-	Error  error
-	Report *visormodel.ProcessingReport
-	Data   model.Persistable
+	Task        string
+	Error       error
+	Report      *visormodel.ProcessingReport
+	Data        model.Persistable
+	StartedAt   time.Time
```
frrist, line 587: <3

commands/setup.go
```diff
@@ -141,7 +141,7 @@ func jaegerConfigFromCliContext(cctx *cli.Context) (*jaegerConfig, error) {
 func setupLogging(cctx *cli.Context) error {
 	// logging.SetAllLoggers(logging.LevelError)
 	ll := cctx.String("log-level")
-	if err := logging.SetLogLevelRegex("visor/*", ll); err != nil {
+	if err := logging.SetLogLevelRegex(".*", ll); err != nil {
 		return xerrors.Errorf("set log level: %w", err)
```
frrist, line 145: You can use `SetLogLevel` instead and drop the regex.

frrist (approved):

PR #573, opened 2021-06-22
feat(bf): decode message bitfields to json

### LINKED ISSUE #569: Decode bitfields
### Description

Some message params and states have bitfield types. This particularly affects ProveCommitAggregate message params.

In our JSONb serializations, we are just storing the bitfields bytes (which are an RLE+ serialization of a bitfield I think) and it is difficult to work with this.

Bitfields are serialized in JSON as array of run-lengths which is also hard to work with. Ideally, what we'd want is an array with the list of elements in the set (https://github.com/filecoin-project/go-bitfield/blob/master/rle/rleplus.go#L71).

Without having this list we cannot, for example, know how many sectors are being proved on a batch commit.

frrist:
- closes #569

### DIFF
--- a/tasks/messages/fcjson/marshal.go
+++ b/tasks/messages/fcjson/marshal.go
@@ -3,9 +3,9 @@ package fcjson
 import (
 	"bytes"
 	"encoding/base64"
-	"encoding/hex"
 	"fmt"
 	"math/big"
+	"strconv"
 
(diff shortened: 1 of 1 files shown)


iand (commented):
A test would be good

hsanjuan:
I am good with the output format.

iand (approved):

PR #574, opened 2021-06-23
Implement ActorState Extractor Filtering

frrist:
Work in progress to allow the extractor tasks to only extract the specified models. Here is the new task string format, it will remain backward compatible:
```
visor walk --tasks="actorstatesminer:miner_info+miner_fee_debt,actorstatespower:chain_power,blocks:block_header,messages" --storage=Database2
```
This command will cause visor to:
- run miner extractor such that it only extracts the `miner_info` and `miner_fee_debt` models (implemented in miner task)
- run power extractor such that it only extracts the  `chain_power` model. (not implemented in task, but the option is there)
- run block extractor such that it only extracts the `block_header` model. (not implemented in task, but the option is there)
- tune the full message extraction with all models.( not implemented in task, but the option is there)


tasks/actorstate/miner.go
```diff
@@ -132,23 +129,58 @@ func NewMinerStateExtractionContext(ctx context.Context, a ActorInfo, node Actor
 		CurrActor: &a.Actor,
 		CurrState: curState,
 		CurrTs:    a.TipSet,
+		Address:   a.Address,
+		Store:     node.Store(),
+		API:       node,
+		cache:     NewDiffCache(),
```
frrist, line 135: This should allow us to keep nearly the same performance we had before the refactor. This caches the results of Diffs for a single extraction.

chain/indexer.go
```diff
@@ -79,6 +68,67 @@ func AddressFilterOpt(f *AddressFilter) TipSetIndexerOpt {
...
+func ModelsForTopLevelTask(t string) (model.PersistableList, error) {
+	return modelreg.ModelRegistry.ModelsForTask(t)
+}
+
+func ParseTaskString(name string) (string, model.PersistableList, error) {
+	if strings.Contains(name, ":") {
+		tokens := strings.Split(name, ":")
+		if len(tokens) != 2 {
+			// TODO don't do this
+			panic("invalid task format? you fucked up, stop doing that")
```
iand, line 84: please don't forget this todo

iand:
I am not a fan of the new task command line syntax. Mainly that it doesn't even attempt to keep backwards compatibility with existing visor. I'd prefer it if the existing format just continued to work so when people upgrade they only need to change their command lines if they are using new features.

Simplistically that could be done by keeping comma as the separator between main tasks and using another character to delimit the subtasks

So instead of

`actorstatesminer:miner_info,miner_fee_debt|actorstatespower:power_actor_claims`

it would be

`actorstatesminer:miner_info+miner_fee_debt,actorstatespower:power_actor_claims`

However I still find it really verbose (it already was  too verbose). Can we shorten the subtasks since they only have meaning in the context of the parent?

`actorstatesminer:info+feedebt,actorstatespower:claims`

tasks/actorstate/miner.go
```diff
@@ -46,48 +72,19 @@ func (m StorageMinerExtractor) Extract(ctx context.Context, a ActorInfo, node Ac
...
-	preCommitModel, sectorModel, sectorDealsModel, sectorEventsModel, err := ExtractMinerSectorData(ctx, ec, a, node)
-	if err != nil {
-		return nil, xerrors.Errorf("extracting miner sector changes: %w", err)
-	}
-
-	posts, err := ExtractMinerPoSts(ctx, &a, ec, node)
-	if err != nil {
-		return nil, xerrors.Errorf("extracting miner posts: %v", err)
+	var out model.PersistableList
+	for _, m := range a.Models {
```
iand, line 76: Can we make this concurrent? So each extraction is in a separate goroutine. Does the diffcache support that?

frrist:
Re: verbosity/compatibility of the task string 

Agreed, will adopt the pattern you've described above. Then, as we discussed synchronously, I will file a follow on PR allowing visor tasks to be specified statically in the configuration file. Following that another PR to adjust miner sub-task schedules allowing them to run in parallel with special attention to the diff caching.

tasks/actorstate/miner.go
frrist: Will add concurrency in a follow-on. Since these tasks never ran concurrently, to begin with, I expect any performance impact resulting from this change to be minimal.

model/genesis/task.go
```diff
@@ -1,125 +0,0 @@
-package genesis
```
frrist, line 1: dead code, not used.

model/registry/register.go
```diff
@@ -0,0 +1,52 @@
...
+	Models map[string]model.PersistableList
+}
+
+func NewRegistry() *Registry {
+	return &Registry{
+		Models: make(map[string]model.PersistableList),
+	}
+}
+
+func (r *Registry) Register(t string, m model.Persistable) {
```
frrist, line 18: how models register themselves as being produced from task names `t`.

model/actors/miner/lockedfunds.go
```diff
@@ -11,6 +12,10 @@ import (
 	"github.com/filecoin-project/sentinel-visor/model"
 )
 
+func init() {
+	registry.ModelRegistry.Register(registry.ActorStatesMinerTask, &MinerLockedFund{})
```
frrist, line 16: each model must register itself as follows.

model/registry/register.go
```diff
@@ -0,0 +1,52 @@
...
+	models, found := r.Models[t]
+	if !found {
+		return nil, xerrors.Errorf("no models for task: %s", string(t))
+	}
+	return models, nil
+}
+
+var ModelRegistry = NewRegistry()
+
+const (
```
frrist, line 40: task names live here now primally due to circular dependencies.

tasks/actorstate/miner.go
```diff
@@ -31,6 +33,31 @@ func init() {
 	}
 }
 
+func ModelExtractors(m model.Persistable) (func(ctx context.Context, ectx *MinerStateExtractionContext) (model.Persistable, error), error) {
```
frrist, line 36: In a subsequent PR this switch can be replaced with a map by forcing models to register themselves with a task: https://github.com/filecoin-project/sentinel-visor/blob/frrist/tasksv2/tasks/actorstate/miner/tasks/extractors/currentdeadlineinfo.go#L14 is a WIP

tasks/actorstate/task.go
```diff
@@ -29,12 +29,15 @@ type Task struct {
 	closer lens.APICloser
 
 	extracterMap ActorExtractorMap
+
+	models model.PersistableList
```
frrist, line 33: ActorState tasks are now passed a list of models to run extraction for during processing.

frrist:
This is rotting, going to close and revisit at a later time if we think we still need it. Currently the miner task is able to average under 30s

PR #577, opened 2021-06-23
feat: add observe command with peer agents task

iand:
This is an initial approach to adding tasks that observe the state of the network rather than the chain. We can iterate on naming.

Adds:
 - a new command `observe` that runs a set of network observing tasks at specified intervals
 - an example task `peeragents` that counts number of each reported agent string among the node's peers
 - a new model and table `observed_peer_agents`

Usage:

The command only works with a visor daemon. Start the daemon then run:

```
visor observe --tasks=peeragents --interval=300s
```

Every 300s the task will write rows to the `observed_peer_agents` table

Notes:

We could add support for using a remote Lotus API lens. Car and SQL lenses would need to be no-ops or fail support somehow.

TODO:

 - [ ] add a migration for the new table. Probably should extract from https://github.com/filecoin-project/sentinel-visor/pull/539 first

### DIFF
--- a/commands/watch.go
+++ b/commands/watch.go
@@ -115,7 +115,7 @@ var WatchCmd = &cli.Command{
 		if err != nil {
 			return err
 		}
-		if _, err := fmt.Fprintf(os.Stdout, "Created Watch Job: %d", watchID); err != nil {
+		if _, err := fmt.Fprintf(os.Stdout, "Created Watch Job: %d\n", watchID); err != nil {
 			return err
 		}
 		return nil
--- a/model/observed/peeragent.go
+++ b/model/observed/peeragent.go
@@ -1,0 +1,55 @@
+package observed
+
+import (
+	"context"
+	"time"
+
+	"go.opencensus.io/tag"
+	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/trace"
+	"go.opentelemetry.io/otel/label"
+
+	"github.com/filecoin-project/sentinel-visor/metrics"
+	"github.com/filecoin-project/sentinel-visor/model"
+)
+
+type PeerAgent struct {
+	//lint:ignore U1000 tableName is a convention used by go-pg
+	tableName struct{} `pg:"observed_peer_agents"`
+
+	// ObservedAt is the time the observation was made
+	ObservedAt time.Time `pg:",notnull"`
+
+	// RawAgent is the raw peer agent string
+	RawAgent string `pg:",notnull"`
+
+	// NormalizedAgent is a parsed version of peer agent string, stripping out patch versions
+	NormalizedAgent string `pg:",notnull"`
+
+	// Count is the number of peers with the associated agent
+	Count int64 `pg:",use_zero,notnull"`
+}
+
+func (p *PeerAgent) Persist(ctx context.Context, s model.StorageBatch, version model.Version) error {
+	ctx, _ = tag.New(ctx, tag.Upsert(metrics.Table, "observed_peer_agents"))
+	stop := metrics.Timer(ctx, metrics.PersistDuration)
+	defer stop()
+
+	return s.PersistModel(ctx, p)
+}
+
+type PeerAgentList []*PeerAgent
+
+func (l PeerAgentList) Persist(ctx context.Context, s model.StorageBatch, version model.Version) error {
+	if len(l) == 0 {
+		return nil
+	}
+	ctx, span := global.Tracer("").Start(ctx, "PeerAgentList.Persist", trace.WithAttributes(label.Int("count", len(l))))
+	defer span.End()
+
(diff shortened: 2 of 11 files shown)


model/observed/peeragent.go
```diff
@@ -0,0 +1,55 @@
...
+	ObservedAt time.Time `pg:",notnull"`
+
+	// RawAgent is the raw peer agent string
+	RawAgent string `pg:",notnull"`
+
+	// NormalizedAgent is a parsed version of peer agent string, stripping out patch versions
+	NormalizedAgent string `pg:",notnull"`
+
+	// Count is the number of peers with the associated agent
+	Count int64 `pg:",use_zero,notnull"`
```
frrist, line 30: I wonder if there is a way to make these code comments table comments on the model via go-pg :thinking:

tasks/observe/peeragents/peeragents.go
```diff
@@ -0,0 +1,82 @@
...
+	for _, peer := range peers {
+		agent, err := t.api.NetAgentVersion(ctx, peer.ID)
+		if err != nil {
+			log.Debugw("failed to get agent version", "error", err)
+			continue
+		}
+		agents[agent]++
+	}
+
+	var l observed.PeerAgentList
```
frrist, line 51: probably not a big deal, but the size of this list is known ahead of time. i.e.
```go
l := make(observed.PeerAgentList, len(agents))
```

frrist (approved):
Nice!

model/observed/peeragent.go
iand: I don't think so but we could/should be generating our schema from these structs and we could do some magic with turning Go comments into postgres comments

frrist:
`observe` doesn't bother me too much, perhaps a name change could be punted until more tasks are added and a better name becomes obvious?

But, If I were to guess what tasks we _might_ add later: `blocksub` (collects blocks the node gets via gossipsub), `msgsub` (same, but for messages), and `mpoolsub` come to mind. This leads me to think `monitor` would be a good choice.

iand:
The more I think about it, the more I like `survey`, as in surveying the environment around the chain: network, peers, incoming blocks and messages. 

I think we should reserve `monitor` for something that is more pro-active, such as threshold watching and/or raising some kind of alert (if we ever want to do that - it's conceivable that we might)

iand:
This overlaps with the new drone spec so we need to decide where it will end up being implemented

commands/watch.go
```diff
@@ -115,7 +115,7 @@ var WatchCmd = &cli.Command{
 		if err != nil {
 			return err
 		}
-		if _, err := fmt.Fprintf(os.Stdout, "Created Watch Job: %d", watchID); err != nil {
+		if _, err := fmt.Fprintf(os.Stdout, "Created Watch Job: %d\n", watchID); err != nil {
```
placer14, line 118: Maybe pull these fixes into master anyway?

placer14 (approved):
Sorry for the late review here. I think this looks great from an impl perspective (regardless of where it lands). For the subcommand name, maybe something like you're taking a measurement from a specific specimen periodically... so I think of `sample` or `glimpse` (at the result of). I'm also happy w `survey`.

schemas/v1/4_surveyed_peer_agents.go
```diff
@@ -0,0 +1,23 @@
...
+	-- Growth: N/A
+	-- ----------------------------------------------------------------
+
+	CREATE TABLE {{ .SchemaName | default "public"}}.surveyed_peer_agents (
+		observed_at       timestamp with time zone NOT NULL,
+		raw_agent         text NOT NULL,
+		normalized_agent  text NOT NULL,
+    	count             bigint NOT NULL
+	);
+	ALTER TABLE ONLY {{ .SchemaName | default "public"}}.surveyed_peer_agents ADD CONSTRAINT surveyed_peer_agents_pkey PRIMARY KEY (observed_at, raw_agent);
```
placer14, line 21: Should we add comments to the schema?

placer14: You might also consider including comments for `gapfill_reports` and `chain_consensus`. 🙏

placer14 (approved):
Let's get it out there and see how it does!

network/surveyer.go
```diff
@@ -0,0 +1,178 @@
...
+
+		llt.Debugw("task report", "time", res.CompletedAt.Sub(res.StartedAt))
+
+		startPersist := time.Now()
+		if err := s.storage.PersistBatch(ctx, res.Data); err != nil {
+			stats.Record(ctx, metrics.PersistFailure.M(1))
+			llt.Errorw("persistence failed", "error", err)
+		} else {
+			llt.Debugw("task data persisted", "time", time.Since(startPersist))
+		}
```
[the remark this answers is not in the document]
iand: I think logging is fine at moment. I don't expect any processing report mechanism since we're not processing tipsets. We're just recording the state of the network at some point in time.

PR #539, opened 2021-06-08
feat(veri): implement verifreg diff methods and update shims

frrist:
- define task for tracking verifreg actor
- define actor shim with codegen and diffing logic for verifreg actor
- define sql migration for new task tables

PR #578, opened 2021-06-23
feat: add v1 schema migration capability

iand:
Needed by #577 and #539

Follows the rough design I outlined in #539 

 - Migrations are tuples consisting of a patch version and some templated sql
 - Each migration lives in its own file just for convenience. The migration tuple is added to a list of "patches" in an init function (similar to what we already do)
 - Change the exported 'Patches' variable in each version package to be a function that accepts the schema config, signature: func GetPatches(cfg schemas.Config) (*migration.Collection, error)
 - For schema v0 we simply return the go-pg collection
 - For schema v1 we execute all the patch sql, use that to make a migration.Migration for each patch and pass them in when creating a new go-pg collection,
 - In storage package, change collectionForVersion to use GetPatches. It will need to have access to the config.

Complete but not thoroughly tested yet


frrist:
Rebased this onto master. Will base #577 and #539 onto this and get them working.

iand:
Planning to merge this after #594

PR #580, opened 2021-06-24
manifest the vector failure: WIP

frrist:
Unfortunately, a lot broke here:
- The process for fetching vectors was broken in the Makefile as it won't detect new vectors being added to the manifest
- The Validate method needs to return an error to indicate failures
- The messages, init_actor, and chain economics vectors broke due to minor changes in their models scheme (the addition of heights and a new field added to the chain economics vector)

Im regenerating the new vectors and fixing the Makefile to detect them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,12 +66,8 @@ build: deps visor
 deps: $(BUILD_DEPS)
 
 .PHONY: vector-setup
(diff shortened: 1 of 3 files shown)


frrist:
This is tracked via #603 will revisit after higher priority items land

PR #583, opened 2021-06-25
fix: avoid indexer blocking head events

iand:
During a watch, skip tipset processing if the indexer is taking longer than one epoch to process. We need to avoid blocking the stream of incoming tipsets otherwise we will cause the node to fall behind the chain while it waits for us to catch up (which may never happen if we consistently take too long)

### DIFF
--- a/chain/indexer.go
+++ b/chain/indexer.go
@@ -63,7 +63,8 @@ type TipSetIndexer struct {
 	messageProcessors map[string]MessageProcessor
 	actorProcessors   map[string]ActorProcessor
 	name              string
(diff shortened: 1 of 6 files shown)


frrist (approved):
Generally looks good.

It seems unlikely that persisting the skipped processing reports could fail due to db congestion, yeah?

iand:
> Generally looks good.
> 
> It seems unlikely that persisting the skipped processing reports could fail due to db congestion, yeah?

Could happen but handling that would make it even more complex so I defered until we need it.

PR #584, opened 2021-06-28
fix: increase API cache size

iand:
Increase API cache size and remove restart delay if indexing takes too long. This reduces choppiness in the first few epochs when starting up visor against mainnet.


frrist (approved):
👍

PR #598, opened 2021-07-13
feat: GapFind and GapFill task implementation

frrist:
Testing the Gap Fill logic is blocked on https://github.com/filecoin-project/sentinel-visor/issues/655. (Need a vector to run the TipSet indexer, Gap Finding logic is well testing however.)

Here is how I've been using this locally:
- Setup Environment (I assume your node is somewhat synced and has a storage config):
 ```
 make calibnet && make dockerdown && make dockerup && sleep 2 && ./visor migrate --latest --db="postgres://postgres:password@localhost:5432/postgres?sslmode=disable" && ./visor daemon
 ```

- Walk the chain with _some_ tasks enable (do this as many times as you'd like, finding should disregard multiple reports at the same epoch):
```
./visor walk --storage=Database1 --tasks=blocks,messages,implicitmessage,msapprovals,actorstatesraw -from=0 -to=200
```

- Run the Gap Find operation (`--to` can go beyond the visor processing report highest height, will label everything as gap) :
```
./visor gap find --storage=Database1 --from=0 --to=200
```

- Check if it found any gaps, lets try the miner first:
```
select count(1)
from visor_gap_reports
where task = 'actorstatesminer' and status = 'GAP'
-> 187
```

- Tell it to gap fill miner tasks:
```
./visor gap fill --storage=Database1 --from=0 --to=200 --tasks=actorstatesminer
```

- Ensure they were all filled:
```
select count(1)
from visor_gap_reports
where task = 'actorstatesminer' and status = 'GAP'
-> 0
```

- Now some other actors:
```
./visor gap fill --storage=Database1 --from=0 --to=200 --tasks=actorstatespower,actorstatesreward
```

- Now everything remaining:
```
./visor gap fill --storage=Database1 --from=0 --to=200
```

- How many gaps left? Repeat the find query and see that there are none found.
```
select count(1)
from visor_gap_reports
where status = 'GAP'
-> 0
```

### DIFF
--- a/chain/gap/fill.go
+++ b/chain/gap/fill.go
@@ -58,7 +58,7 @@ func (g *GapFiller) Run(ctx context.Context) error {
 		}
 		fillLog.Infow("fill success", "height", height, "remaining", len(gaps)-idx)
 
-		if err := g.setGapsFilled(ctx, height); err != nil {
+		if err := g.setGapsFilled(ctx, height, gaps[height]...); err != nil {
 			return err
 		}
 		idx += 1
@@ -113,10 +113,11 @@ func (g *GapFiller) queryGaps(ctx context.Context) ([]*visor.GapReport, error) {
 }
 
 // mark all gaps at height as filled.
-func (g *GapFiller) setGapsFilled(ctx context.Context, height int64) error {
+func (g *GapFiller) setGapsFilled(ctx context.Context, height int64, tasks ...string) error {
 	if _, err := g.DB.AsORM().ModelContext(ctx, &visor.GapReport{}).
 		Set("status = 'FILLED'").
 		Where("height = ?", height).
+		Where("task = ANY (?)", pg.Array(tasks)).
 		Update(); err != nil {
 		return err
 	}
--- a/chain/gap/fill.go
+++ b/chain/gap/fill.go
@@ -1,0 +1,124 @@
+package gap
+
+import (
+	"context"
+	"sort"
+
+	"github.com/filecoin-project/sentinel-visor/chain"
+	"github.com/filecoin-project/sentinel-visor/lens"
+	"github.com/filecoin-project/sentinel-visor/model/visor"
+	"github.com/filecoin-project/sentinel-visor/storage"
+	"github.com/go-pg/pg/v10"
+	logging "github.com/ipfs/go-log/v2"
+	"golang.org/x/xerrors"
+)
+
+var log = logging.Logger("visor/gap")
+
+type GapFiller struct {
+	DB                   *storage.Database
+	opener               lens.APIOpener
+	name                 string
+	minHeight, maxHeight uint64
+	tasks                []string
+}
+
+func NewGapFiller(o lens.APIOpener, db *storage.Database, name string, maxHeight, minHeight uint64, tasks []string) *GapFiller {
+	return &GapFiller{
+		DB:        db,
+		opener:    o,
+		name:      name,
+		maxHeight: maxHeight,
+		minHeight: minHeight,
+		tasks:     tasks,
+	}
+}
+
+func (g *GapFiller) Run(ctx context.Context) error {
+	gaps, heights, err := g.consolidateGaps(ctx)
+	if err != nil {
+		return err
+	}
+	fillLog := log.With("type", "fill")
+	fillLog.Infow("run", "count", len(gaps))
+
+	idx := 0
+	for _, height := range heights {
+		indexer, err := chain.NewTipSetIndexer(g.opener, g.DB, 0, g.name, gaps[height])
+		if err != nil {
+			return err
+		}
+
+		// walk a single height at a time since there is no guarantee neighboring heights share the same missing tasks.
+		if err := chain.NewWalker(indexer, g.opener, height, height).Run(ctx); err != nil {
+			log.Errorw("fill failed", "height", height, "error", err.Error())
+			// TODO we could add an error to the gap report in a follow on if needed, but the actualy error should
+			// exist in the processing report if this fails.
+			continue
+		}
+		fillLog.Infow("fill success", "height", height, "remaining", len(gaps)-idx)
+
+		if err := g.setGapsFilled(ctx, height); err != nil {
+			return err
+		}
+		idx += 1
+	}
+	return nil
+}
+
+// returns a map of heights to missing tasks, and a list of heights to iterate the map in order with.
+func (g *GapFiller) consolidateGaps(ctx context.Context) (map[int64][]string, []int64, error) {
+	gaps, err := g.queryGaps(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+	// used to walk gaps in order, should help optimize some caching.
+	heights := make([]int64, 0, len(gaps))
+	out := make(map[int64][]string)
+	for _, gap := range gaps {
+		if _, ok := out[gap.Height]; !ok {
+			heights = append(heights, gap.Height)
+		}
+		out[gap.Height] = append(out[gap.Height], gap.Task)
+	}
+	sort.Slice(heights, func(i, j int) bool {
+		return heights[i] < heights[j]
+	})
+	return out, heights, nil
+}
+
+func (g *GapFiller) queryGaps(ctx context.Context) ([]*visor.GapReport, error) {
+	var out []*visor.GapReport
+	if len(g.tasks) != 0 {
+		if err := g.DB.AsORM().ModelContext(ctx, &out).
+			Order("height desc").
+			Where("status = ?", "GAP").
+			Where("task = ANY (?)", pg.Array(g.tasks)).
+			Where("height >= ?", g.minHeight).
+			Where("height <= ?", g.maxHeight).
+			Select(); err != nil {
+			return nil, xerrors.Errorf("querying gap reports: %w", err)
+		}
+	} else {
+		if err := g.DB.AsORM().ModelContext(ctx, &out).
+			Order("height desc").
+			Where("status = ?", "GAP").
+			Where("height >= ?", g.minHeight).
+			Where("height <= ?", g.maxHeight).
+			Select(); err != nil {
+			return nil, xerrors.Errorf("querying gap reports: %w", err)
+		}
+	}
+	return out, nil
+}
+
+// mark all gaps at height as filled.
+func (g *GapFiller) setGapsFilled(ctx context.Context, height int64) error {
+	if _, err := g.DB.AsORM().ModelContext(ctx, &visor.GapReport{}).
+		Set("status = 'FILLED'").
+		Where("height = ?", height).
+		Update(); err != nil {
+		return err
+	}
+	return nil
+}
--- a/chain/gap/find.go
+++ b/chain/gap/find.go
@@ -1,0 +1,264 @@
+package gap
+
+import (
+	"context"
+	"time"
+
+	"github.com/deckarep/golang-set"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/sentinel-visor/chain"
+	"github.com/filecoin-project/sentinel-visor/lens"
+	"github.com/filecoin-project/sentinel-visor/model/visor"
+	"github.com/filecoin-project/sentinel-visor/storage"
+	"golang.org/x/xerrors"
+)
+
+type GapIndexer struct {
+	DB                   *storage.Database
+	opener               lens.APIOpener
+	name                 string
+	minHeight, maxHeight uint64
+}
+
+var TaskSet mapset.Set
+
+func init() {
+	TaskSet = mapset.NewSet()
+	for _, t := range chain.AllTasks {
+		TaskSet.Add(t)
+	}
+}
+
+func NewGapIndexer(o lens.APIOpener, db *storage.Database, name string, maxHeight, minHeight uint64) *GapIndexer {
+	return &GapIndexer{
+		DB:        db,
+		opener:    o,
+		name:      name,
+		maxHeight: maxHeight,
+		minHeight: minHeight,
+	}
+}
+
+func (g *GapIndexer) Run(ctx context.Context) error {
+	startTime := time.Now()
+	node, closer, err := g.opener.Open(ctx)
+	if err != nil {
+		return xerrors.Errorf("open lens: %w", err)
+	}
+	defer func() {
+		closer()
+	}()
+
+	head, err := node.ChainHead(ctx)
+	if err != nil {
+		return err
+	}
+	maxHeight := g.maxHeight
+	if uint64(head.Height()) < g.maxHeight {
+		return xerrors.Errorf("cannot look for gaps beyond chain head height %d", head.Height())
+	}
+
+	findLog := log.With("type", "find")
+
+	// looks for incomplete epochs. An incomplete epoch has some, but not all tasks in the processing report table.
+	taskGaps, err := g.findTaskEpochGaps(ctx, maxHeight, g.minHeight, chain.AllTasks...)
+	if err != nil {
+		return xerrors.Errorf("finding task epoch gaps: %w", err)
+	}
+	findLog.Infow("found gaps in tasks", "count", len(taskGaps))
+
+	// looks for missing epochs and null rounds. A missing epoch is a non-null-round height missing from the processing report table
+	heightGaps, nulls, err := g.findEpochGapsAndNullRounds(ctx, node, maxHeight, g.minHeight)
+	if err != nil {
+		return xerrors.Errorf("finding epoch gaps: %w", err)
+	}
+	findLog.Infow("found gaps in epochs", "count", len(heightGaps))
+	findLog.Infow("found null rounds", "count", len(nulls))
+
+	// looks for entriest in the visor processing report table that have been skipped.
+	skipGaps, err := g.findEpochSkips(ctx, maxHeight, g.minHeight)
+	if err != nil {
+		return xerrors.Errorf("detecting skipped gaps: %w", err)
+	}
+	findLog.Infow("found skipped epochs", "count", len(skipGaps))
+
+	var nullRounds visor.ProcessingReportList
+	for _, epoch := range nulls {
+		nullRounds = append(nullRounds, &visor.ProcessingReport{
+			Height:            int64(epoch),
+			StateRoot:         "NULL_ROUND", // let gap fill add the correct state root for this epoch when it runs the consensus task.
+			Reporter:          g.name,
+			Task:              "gap_find",
+			StartedAt:         startTime,
+			CompletedAt:       time.Now(),
+			Status:            visor.ProcessingStatusInfo,
+			StatusInformation: visor.ProcessingStatusInformationNullRound, // the gap finding logic uses this value to exclude null rounds from gap report.
+		})
+	}
+
+	return g.DB.PersistBatch(ctx, skipGaps, heightGaps, taskGaps, nullRounds)
+}
+
+type GapIndexerLens interface {
+	ChainGetTipSetByHeight(ctx context.Context, epoch abi.ChainEpoch, tsk types.TipSetKey) (*types.TipSet, error)
+}
+
+func (g *GapIndexer) findEpochSkips(ctx context.Context, max, min uint64) (visor.GapReportList, error) {
+	log.Debug("finding skipped epochs")
+	reportTime := time.Now()
+
+	var skippedReports []visor.ProcessingReport
+	if err := g.DB.AsORM().ModelContext(ctx, &skippedReports).
+		Order("height desc").
+		Where("status = ?", visor.ProcessingStatusSkip).
+		Where("height >= ?", min).
+		Where("height <= ?", max).
+		Select(); err != nil {
+		return nil, xerrors.Errorf("query processing report skips: %w", err)
+	}
+	log.Debugw("executed find skipped epoch query", "count", len(skippedReports))
+
+	gapReport := make([]*visor.GapReport, len(skippedReports))
+	for idx, r := range skippedReports {
+		gapReport[idx] = &visor.GapReport{
+			Height:     r.Height,
+			Task:       r.Task,
+			Status:     "GAP",
+			Reporter:   g.name,
+			ReportedAt: reportTime,
+		}
+	}
+	return gapReport, nil
+}
+
+func (g *GapIndexer) findEpochGapsAndNullRounds(ctx context.Context, node GapIndexerLens, max, min uint64) (visor.GapReportList, []abi.ChainEpoch, error) {
+	log.Debug("finding epoch gaps and null rounds")
+	reportTime := time.Now()
+
+	var nullRounds []abi.ChainEpoch
+	var missingHeights []uint64
+	res, err := g.DB.AsORM().QueryContext(
+		ctx,
+		&missingHeights,
+		`
+		SELECT s.i AS missing_epoch
+		FROM generate_series(?, ?) s(i)
+		WHERE NOT EXISTS (SELECT 1 FROM visor_processing_reports WHERE height = s.i);
+		`,
+		min, max)
+	if err != nil {
+		return nil, nil, err
+	}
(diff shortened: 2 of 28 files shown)


chain/gap/fill.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (g *GapFiller) updateGaps(ctx context.Context, gaps visor.GapReportList) error {
+	if len(gaps) == 0 {
+		return nil
+	}
+	_, err := g.DB.AsORM().Model(&gaps).WherePK().Update()
+	return err
+}
+
+func tipSetKeyFromString(tss string) (types.TipSetKey, error) {
```
frrist, line 97: didn't end up needing this but let it here since I may circle back.

chain/gap/find.go
```diff
@@ -0,0 +1,147 @@
...
+			return nil, xerrors.Errorf("getting tipset by height %d: %w", gh, err)
+		}
+		if tsgap.Height() == gh {
+			for _, task := range chain.AllTasks {
+				gapReport = append(gapReport, &visor.GapReport{
+					Height:     int64(tsgap.Height()),
+					TipSet:     tsgap.Key().String(),
+					Task:       task,
+					Status:     "GAP",
+					Reporter:   "gapIndexer", // TODO does this really need a name?
```
iand, line 109: The reporter should let us identify which instance found the gap if we are running multiple

chain/gap/find.go
```diff
@@ -0,0 +1,147 @@
...
+	return gapReport, nil
+}
+
+func (g *GapIndexer) detectProcessingGaps(ctx context.Context, node lens.API) (visor.GapReportList, error) {
+	reportTime := time.Now()
+	// TODO unsure how big these lists will be, will likely want some sort of pagination here with limits and a loop over it
+
+	// build list of all reports ordered by height, scan these to find gaps in epochs
+	var reportHeights []visor.ProcessingReport
+	if err := g.DB.AsORM().Model(&reportHeights).
```
iand, line 87: This is a lot of data. Soon to be a million rows.

chain/gap/find.go
```diff
@@ -0,0 +1,147 @@
...
+	node, closer, err := g.opener.Open(ctx)
+	if err != nil {
+		return xerrors.Errorf("open lens: %w", err)
+	}
+	defer func() {
+		closer()
+	}()
+
+	findLog := log.With("type", "find")
+	heightGaps, err := g.detectProcessingGaps(ctx, node)
```
iand, line 39: I think it would be useful to identify and explicitly label null rounds. It would stop us repeatedly checking them to see if they need to be gap filled. Also it would be a useful record of the chain history. 

There are a few ways we could represent the data:
 1. use visor_processing_reports with a new status of "NULL"
 2. create a new `null_rounds` table that simply lists the heights that were determined to be null rounds
 3. create a new `height_blocks` table that contains the number of blocks at each height

I'm pretty sure that null rounds are affected by reorgs (e.g. your node happens to be following a fork with a bunch of null rounds but then discovers a better fork with a different distribution of null rounds). That implies that we need to continually re-check for null rounds until finality has been reached.... but then we should really be reprocessing tipsets after a reorg anyway.

We could perhaps handle this with a new "consensus chain" task that follows the chain 900 epochs behind the head and builds a table consisting of height, stateroot and block cids/tipset key (empty for a null round).

chain/gap/find.go
```diff
@@ -0,0 +1,147 @@
...
+			Task:       r.Task,
+			Status:     "GAP",
+			Reporter:   "gapIndexer", // TODO does this really need a name?
+			ReportedAt: reportTime,
+		}
+	}
+	return gapReport, nil
+}
+
+func (g *GapIndexer) detectProcessingGaps(ctx context.Context, node lens.API) (visor.GapReportList, error) {
```
iand, line 81: What if we have a gap for one task but not another?

chain/gap/find.go
```diff
@@ -0,0 +1,147 @@
...
+	if err := g.DB.AsORM().Model(&reportHeights).
+		Order("height desc").
+		DistinctOn("height").
+		Select(); err != nil {
+		return nil, xerrors.Errorf("query processing report heights: %w", err)
+	}
+
+	gapReport := make([]*visor.GapReport, 0, len(reportHeights))
+	// walk the possible gaps and query lotus to determine if gap was a null round or missed epoch.
+	for _, gap := range findEpochGaps(reportHeights) {
```
iand, line 96: Is a gap simply missing data or will it take into account taks that failed due to errors. These will have a report in the table.

iand:
Is there a need for both tasks? Can they be combined so we fill gaps as we find them?

chain/gap/find.go
whyrusleeping: seems like you could select reports where an adjacently numbered report doesnt exist.

frrist: I think a lot could be improved here:
1. Configurable pagination based on height.
2. Clever SQL, e.g. `select * from visor_processing_reports where task=<task> and (select count(1) from visor_processing_reports where height = rep.height + 1) = 0;` 

will adjust.

chain/gap/find.go
frrist: Perhaps a combination of the above. I'd like to incorporate (1) into this PR since it will make gap finding more efficient.
I've filed https://github.com/filecoin-project/sentinel-visor/issues/611 for creating a canonical task for tracking null rounds.

chain/gap/find.go
frrist: Yeah great point (thought I had a TODO for this edge case). I think https://github.com/filecoin-project/sentinel-visor/pull/598#discussion_r681203074 should address this.

chain/gap/find.go
frrist: The method `detectSkippedTipSets` will look for gaps due to skips. I'm not 100% how errors should be handled since this could represent an error in our processing logic or something unknown that needs to be investigated. I'd like to punt handling errors to a follow-on.

frrist:
> Is there a need for both tasks? Can they be combined so we fill gaps as we find them?

I have a hunch the gap-finding task will run relatively quickly and expect we'll only need a single instance on some cron job. In contrast, I expect the gap-filling task to take longer to run and expect we may want to only gap-fill certain task types, or run multiple gap-filling tasks in parallel. The pattern I am attempting to implement here involves the gap-finding task creating items for multiple gap-filling tasks to consume.

iand:
Note: might be better to refer to these as jobs not tasks. They should appear when `visor job list` is run and it's less confusing to users if we keep the word "task" to mean a subcomponent of a job.

chain/gap/find.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	return g.DB.PersistBatch(ctx, skipGaps, heightGaps, taskGaps, nullRounds)
+}
+
+type GapIndexerLens interface {
+	ChainGetTipSetByHeight(ctx context.Context, epoch abi.ChainEpoch, tsk types.TipSetKey) (*types.TipSet, error)
+}
+
+func (g *GapIndexer) findEpochSkips(ctx context.Context, max, min uint64) (visor.GapReportList, error) {
```
iand, line 103: Can you split this into a separate type and introduce a `GapFinder` interface? I'd like to be able to reuse some of this logic by passing in a different implementation (reading from csv dump for example).

chain/gap/fill.go
```diff
@@ -0,0 +1,101 @@
...
+	fillLog.Infow("run", "count", len(gaps))
+
+	for idx, gap := range gaps {
+		// TODO we could optimize here by collecting all gaps at same height and launching a single instance of the walker for them
+		indexer, err := chain.NewTipSetIndexer(g.opener, g.DB, 0, g.name, []string{gap.Task})
+		if err != nil {
+			gap.Status = err.Error()
+			log.Errorw("fill failed", "height", gap.Height, "error", err.Error())
+		} else {
+			walker := chain.NewWalker(indexer, g.opener, gap.Height, gap.Height)
```
iand, line 51: Is this only ever going to deal with single height gaps? What if we are missing 10 epochs in a row?

Also, many tasks require 2 tipsets to operate to perform the diff (we could simply fetch parent if we haven't walked it but currently we don't do that)

frrist: > Is this only ever going to deal with single height gaps? What if we are missing 10 epochs in a row?

I have a TODO above that calls this out. An enhancement to this logic would find the longest contiguous gap and walk that. I haven't written it yet as the code works fine without the optimization.

> Also, many tasks require 2 tipsets to operate to perform the diff (we could simply fetch parent if we haven't walked it but currently we don't do that)

I think this happens implicitly via the WalkChain method, but maybe I am misunderstanding the point you are making:
First call fails to find a parent and set `lastTipSet`
https://github.com/filecoin-project/sentinel-visor/blob/ba4c266e345c7f0018bd609c566d19324874e587/chain/walker.go#L75-L78

Second call has a parent and:
https://github.com/filecoin-project/sentinel-visor/blob/ba4c266e345c7f0018bd609c566d19324874e587/chain/walker.go#L93-L96

Are you stating the indexer could fetch the parent if it's null?

iand: Yes if parent  == nil we could fetch it quite easily and cheaply.

frrist: Addressed via: https://github.com/filecoin-project/sentinel-visor/pull/653

chain/gap/find.go
frrist: Addressed. via fancy sql

chain/gap/find.go
frrist: I've updated comments on the methods describing gaps.

frrist:
> Note: might be better to refer to these as jobs not tasks. They should appear when visor job list is run and it's less confusing to users if we keep the word "task" to mean a subcomponent of a job.

These are in fact jobs, they will show up in the `job list` command:
```
{
                "ID": 2,
                "Name": "find_1629338958",
                "Type": "",
                "Error": "",
                "Tasks": null,
                "Running": false,
                "RestartOnFailure": false,
                "RestartOnCompletion": false,
                "RestartDelay": 0,
                "Params": null,
                "StartedAt": "2021-08-19T02:09:18.506351009Z",
                "EndedAt": "2021-08-19T02:09:18.571265185Z"
        },
        {
                "ID": 3,
                "Name": "fill_1629338965",
                "Type": "",
                "Error": "",
                "Tasks": [
                        "actorstatesraw",
                        "actorstatespower",
                        "actorstatesreward",
                        "actorstatesminer",
                        "actorstatesinit",
                        "actorstatesmarket",
                        "actorstatesmultisig",
                        "blocks",
                        "messages",
                        "chaineconomics",
                        "msapprovals",
                        "implicitmessage",
                        "consensus"
                ],
                "Running": false,
                "RestartOnFailure": false,
                "RestartOnCompletion": false,
                "RestartDelay": 0,
                "Params": null,
                "StartedAt": "2021-08-19T02:09:25.713230155Z",
                "EndedAt": "2021-08-19T02:09:39.261150894Z"
        },
```
Maybe I am missing params?

chain/gap/find.go
frrist: Yes, what would you like it to look like? Maybe:
```go
type GapFinder interface {
    Find(ctx context.Context, to, from uint64, tasks ...string) (visor.GapReportList, []abi.ChainEpoch, error)
}
``` 

The returned array `[]abi.ChainEpoch` containing null rounds

chain/gap/find.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+type TaskHeight struct {
+	Task   string
+	Height uint64
+	Status string
+}
+
+// TODO rather than use the length of `tasks` to determine where gaps are, use the contents to look for
+// gaps in specific task. Forrest' SQL-Foo isn't good enough for this yet.
```
frrist, line 189: I'd love to get this addressed, just ran out of ideas for how to write it. Being able to only report gaps to the gaps_report_table for specific types of tasks will be useful for our monitoring instances since they don't run with everything enabled. Alternatively filtering could be performed below.

chain/gap/find.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	maxHeight := g.maxHeight
+	if uint64(head.Height()) < g.maxHeight {
+		return xerrors.Errorf("cannot look for gaps beyond chain head height %d", head.Height())
+	}
+
+	findLog := log.With("type", "find")
+
+	// looks for incomplete epochs. An incomplete epoch has some, but not all tasks in the processing report table.
+	taskGaps, err := g.findTaskEpochGaps(ctx, maxHeight, g.minHeight, chain.AllTasks...)
```
frrist, line 65: TODO for myself tomorrow, don't hard code this, make it a param to gap find.

chain/gap/find.go
```diff
@@ -0,0 +1,264 @@
...
+	var out visor.GapReportList
+	// returns a list of tasks and heights for all incomplete heights
+	// and incomplete height is a height with less than len(tasks) entries, the tasks returned
+	// are the completed tasks for the height, we can diff them against all known tasks to find the
+	// missing ones.
+	res, err := g.DB.AsORM().QueryContext(
+		ctx,
+		&result,
+		`
+select height, task
```
frrist, line 203: I ran this against mainnet on range 900_000 to 1_000_000 and it took about 3 seconds fwiw.

lens/lotus/api.go
```diff
@@ -37,6 +38,10 @@ type APIWrapper struct {
 	store adt.Store
 }
 
+func (aw *APIWrapper) ChainGetTipSetAfterHeight(ctx context.Context, epoch abi.ChainEpoch, key types.TipSetKey) (*types.TipSet, error) {
+	panic("implement me")
```
frrist, line 42: were killing all these soon, panic seems fine.

chain/walker.go
```diff
@@ -54,11 +54,10 @@ func (c *Walker) Run(ctx context.Context) error {
 		return xerrors.Errorf("cannot walk history, chain head (%d) is earlier than minimum height (%d)", int64(ts.Height()), c.minHeight)
 	}
 
-	// TODO: use ChainGetTipSetAfterHeight to handle null rounds properly (deferring until we have removed/simplified lenses)
 	// Start at maxHeight+1 so that the tipset at maxHeight becomes the parent for any tasks that need to make a diff between two tipsets.
 	// A walk where min==max must still process two tipsets to be sure of extracting data.
 	if int64(ts.Height()) > c.maxHeight+1 {
-		ts, err = node.ChainGetTipSetByHeight(ctx, abi.ChainEpoch(c.maxHeight+1), types.EmptyTSK)
+		ts, err = node.ChainGetTipSetAfterHeight(ctx, abi.ChainEpoch(c.maxHeight+1), types.EmptyTSK)
```
frrist, line 60: this needed to be addressed inorder for this all to work.

lens/lotus/api.go
iand: It does mean that walks against a lotus api will be broken in master. We ought to have a tagged release before this happens.

Alternatively we could fudge it by returning ChainGetTipSetByHeight(height+1). That is no worse than what we are currently doing.

chain/gap/find.go
```diff
@@ -0,0 +1,264 @@
...
+		`,
+		min, max)
+	if err != nil {
+		return nil, nil, err
+	}
+	log.Debugw("executed find epoch gap query", "count", res.RowsReturned())
+
+	gapReport := make([]*visor.GapReport, 0, len(missingHeights))
+	// walk the possible gaps and query lotus to determine if gap was a null round or missed epoch.
+	for _, gap := range missingHeights {
```
iand, line 157: Probably want to check whether context is done here so we can shut down nicely.

chain/gap/find.go
iand: Could you weave in clauses like `where task in('messsages','blocks')` and pass in the number of tasks to compare with?

commands/gap.go
```diff
@@ -0,0 +1,200 @@
...
+			Storage:             gapFlags.storage,
+			Name:                fillName,
+			Tasks:               tasks,
+			MaxHeight:           gapFlags.to,
+			MinHeight:           gapFlags.from,
+		})
+		if err != nil {
+			return err
+		}
+		if _, err := fmt.Fprintf(os.Stdout, "Created Gap Job: %d", gapFindID); err != nil {
```
iand, line 119: Missing newline

commands/gap.go
```diff
@@ -0,0 +1,200 @@
...
+			RestartDelay:        0,
+			Storage:             gapFlags.storage,
+			Name:                findName,
+			MaxHeight:           gapFlags.to,
+			MinHeight:           gapFlags.from,
+		})
+		if err != nil {
+			return err
+		}
+		if _, err := fmt.Fprintf(os.Stdout, "Created Gap Job: %d", gapFindID); err != nil {
```
iand, line 195: newline

commands/gap.go
```diff
@@ -0,0 +1,200 @@
...
+		},
+		&cli.StringFlag{
+			Name:        "storage",
+			Usage:       "Name of storage that results will be written to.",
+			Value:       "",
+			Destination: &gapFlags.storage,
+		},
+		&cli.StringFlag{
+			Name:        "name",
+			Usage:       "name of instance performing find",
```
iand, line 152: Walk and Watch have usage `"Name of job for easy identification later."`

commands/setup.go
```diff
@@ -139,11 +139,15 @@ func jaegerConfigFromCliContext(cctx *cli.Context) (*jaegerConfig, error) {
...
 func setupLogging(cctx *cli.Context) error {
-	ll := cctx.String("log-level")
-	if err := logging.SetLogLevel("*", ll); err != nil {
+	//ll := cctx.String("log-level")
+	if err := logging.SetLogLevel("*", "error"); err != nil {
 		return xerrors.Errorf("set log level: %w", err)
 	}
 
+	if err := logging.SetLogLevelRegex("visor/*", "debug"); err != nil {
+		panic(err)
```
iand, line 148: Panic doesn't seem right here. Why not return an error which will be reported to user running command.

commands/setup.go
```diff
@@ -139,11 +139,15 @@ func jaegerConfigFromCliContext(cctx *cli.Context) (*jaegerConfig, error) {
...
 
 func setupLogging(cctx *cli.Context) error {
-	ll := cctx.String("log-level")
-	if err := logging.SetLogLevel("*", ll); err != nil {
+	//ll := cctx.String("log-level")
+	if err := logging.SetLogLevel("*", "error"); err != nil {
 		return xerrors.Errorf("set log level: %w", err)
 	}
 
+	if err := logging.SetLogLevelRegex("visor/*", "debug"); err != nil {
```
iand, line 147: Shouldn't this be `visor/.*` since it's a regex? And why hardcode the level?

commands/setup.go
```diff
@@ -139,11 +139,15 @@ func jaegerConfigFromCliContext(cctx *cli.Context) (*jaegerConfig, error) {
 }
 
 func setupLogging(cctx *cli.Context) error {
-	ll := cctx.String("log-level")
-	if err := logging.SetLogLevel("*", ll); err != nil {
+	//ll := cctx.String("log-level")
+	if err := logging.SetLogLevel("*", "error"); err != nil {
```
iand, line 143: Remove hard coded level

commands/setup.go
```diff
@@ -159,6 +163,9 @@ func setupLogging(cctx *cli.Context) error {
 	}
 
 	log.Infof("Visor version:%s", version.String())
+	if err := logging.SetLogLevelRegex("visor/*", "info"); err != nil {
```
iand, line 166: Same as above

lens/lily/impl.go
```diff
@@ -142,6 +153,59 @@ func (m *LilyNodeAPI) LilyWalk(_ context.Context, cfg *LilyWalkConfig) (schedule
...
+		JobName: cfg.Name,
+	}
+
+	// create a database connection for this watch, ensure its pingable, and run migrations if needed/configured to.
+	db, err := m.StorageCatalog.ConnectAsDatabase(ctx, cfg.Storage, md)
+	if err != nil {
+		return schedule.InvalidJobID, err
+	}
+
+	// TODO add params like walk and watch
```
iand, line 170: Let's do this TODO. from/to and storage seem the best params to add

lens/lily/impl.go
```diff
@@ -142,6 +153,59 @@ func (m *LilyNodeAPI) LilyWalk(_ context.Context, cfg *LilyWalkConfig) (schedule
...
```
iand, line 196: Do the TODO

lens/lotusrepo/repo.go
```diff
@@ -111,6 +112,10 @@ type RepoAPI struct {
 	cacheSize int
 }
 
+func (ra *RepoAPI) ChainGetTipSetAfterHeight(ctx context.Context, epoch abi.ChainEpoch, key types.TipSetKey) (*types.TipSet, error) {
+	panic("implement me")
```
iand, line 116: See my comment elsewhere. We could fudge it by calling ChainGetTipSetByHeight(height+1). This repo lens is still in current use for backfilling so we can't not support it right now.

storage/catalog.go
```diff
@@ -115,3 +116,18 @@ type StorageWithMetadata interface {
...
+// a Database storage.
+func (c *Catalog) ConnectAsDatabase(ctx context.Context, name string, md Metadata) (*Database, error) {
+	strg, err := c.Connect(ctx, name, md)
+	if err != nil {
+		return nil, err
+	}
+
+	db, ok := strg.(*Database)
+	if !ok {
+		return nil, xerrors.Errorf("connecting to storage as database. unexpected type: %T", strg)
```
iand, line 130: Improve this error message since it's going to be shown to a user. Maybe `storage type is unsupported`. Log the type if you think it is useful.

chain/gap/find.go
iand: On second thoughts I don't think it adds much. We can simply replicate logic if needed - it's pretty short.

chain/gap/find.go
frrist: I think I could do something like that -- tried here recently but broke things. I'm going to punt that to a follow-on once #655 is addressed and I can automate testing for this with vectors.

storage/catalog.go
frrist: done.

lens/lotusrepo/repo.go
frrist: I just reimplemented what I did for the lily impl, should work correctly for null rounds now as well.

lens/lily/impl.go
frrist: done.

lens/lily/impl.go
frrist: done.

commands/setup.go
frrist: I removed all my logging changes.

commands/gap.go
frrist: done.

commands/gap.go
frrist: done here and in fill

chain/gap/fill.go
```diff
@@ -0,0 +1,125 @@
...
+func (g *GapFiller) Run(ctx context.Context) error {
+	gaps, heights, err := g.consolidateGaps(ctx)
+	if err != nil {
+		return err
+	}
+	fillLog := log.With("type", "fill")
+	fillLog.Infow("run", "count", len(gaps))
+
+	idx := 0
+	for _, height := range heights {
```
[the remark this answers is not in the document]
frrist: done.

chain/gap/find.go
frrist: done.

chain/gap/fill.go
```diff
@@ -0,0 +1,125 @@
+package gap
```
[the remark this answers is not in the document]
frrist: I wanted to give it its own logger and still refer to it as `log`. Move into the chain pkg as you've suggested.

frrist:
@iand I've addressed all feedback and added filtering for finding gaps -- meaning we can pick which task(s) we look for gaps in.

iand (approved):

iand:
Looking forward to getting this running

PR #599, opened 2021-07-14
feat: add metrics for tipset cache used by watches

iand:
Adds metrics that allows monitoring of the depth of the tipset cache and identification of reorgs that extend beyond the size of the cache. Also fixes the `watch_height` metric that was not being reported for lily. `watch_height` records the height of the most recent tipset event which may be an apply or a revert. Reorgs should be visible as a rapid reduction in this metric followed by a corresponding increase back to the expected epoch height.


frrist (approved):
:+1:

placer14 (approved):

PR #606, opened 2021-07-29
fix: export watch_height metric

iand:
This is not the first time we have forgotten to export a view by not adding it to the DefaultViews slice. So I made it harder to get wrong by defining the views directly in the slice.


placer14 (approved):

iand:
Lint failure fixed in https://github.com/filecoin-project/sentinel-visor/pull/607

PR #618, opened 2021-08-05
fix(task): only process message executions for tipsets on same chain

frrist:


chain/indexer.go
```diff
@@ -285,6 +285,36 @@ func (t *TipSetIndexer) TipSet(ctx context.Context, ts *types.TipSet) error {
 						}
 					}
 				}
+
+				// If we have messages execution processors then extract internal messages
+				if len(t.messageExecutionProcessors) > 0 {
```
frrist, line 290: I must have messed this up while rebasing. Fixed now.

iand (approved):

PR #630, opened 2021-08-10
polish: include window and storage in job params output

iand:
Also refactor how the params are surfaced to avoid hardcoding known jobs types in a type switch. Changed params map to have string values to make it simpler for clients to format result as a list or similar.


iand:
Not sure why test build failed.

frrist (approved):
love this

Issue #593, opened 2021-07-01
Date of genesis epoch is hard coded for mainnet


iand:
### Describe the bug:

The date of the genesis epoch is hardcoded here https://github.com/filecoin-project/sentinel-visor/blob/master/commands/util.go#L10 

This is used to estimate the current epoch which is currently only used for setting the default value of the `to` option for the `walk` and `vector` commands.

PR #631, opened 2021-08-10
fix: replace current epoch estimation with new function

iand:
Fixes https://github.com/filecoin-project/sentinel-visor/issues/616 and #593


frrist (approved):

frrist:
My approval is sticky, but any chance https://github.com/filecoin-project/sentinel-visor/issues/593 could be included in the fix here?

PR #648, opened 2021-08-16
fix: avoid missing block when miner loads receipt

### LINKED ISSUE #647: Block not found when extracting miner posts
### Describe the bug:

The `actorstatesminer` task occasionally fails with the following error: `failed to extract parsed actor state: extracting miner posts: failed to load message: blockstore: block not found`

For example:

[{"Code":"bafkqaetgnfwc6njpon2g64tbm5sw22lomvza","Name":"fil/5/storageminer","Head":"bafy2bzacearue7akcwpowzmbj3nf3x6qwwqiotnsgcgegcptkgiwkxzw3skfs","Address":"f0729412","Error":"failed to extract parsed actor state: extracting miner posts: failed to load message: blockstore: block not found"}

Adding more error context reveals that the error is returned from a call to `StateGetReceipt` which the extracter is using to check whether the message sent to the miner actor was executed

### Steps to Reproduce:

`./visor walk --from 1006966 --to=1006967 --tasks=actorstatesminer` will reveal 3 instances of the error

### Visor Version: master c2c680a628a5aa5130a052bb798b398751efe8bc

This is likely a different failure to the one reported in https://github.com/filecoin-project/sentinel-visor/issues/562

iand:
Fixes #647 

The miner extracter called `ChainGetParentMessages` and `StateGetReceipt` but occasionally the latter would return a block not found error instead of nil when there is no receipt for a message. Fix the error by using our pre-extracted list of messages instead. This has the side efffect of speeding up the miner task (from 30s -> 20s plus ~5s overhead for extracting messages which are needed for other tasks anyway)


iand:
I need to rebuild vectors if you're ok with these changes.

I did some before and after testing on my node over the suggested range in the linked issue. With this fix we return 3 more rows in miner_locked_funds and miner_fee_debts corresponding to the 3 miners that were failing. Plus 28125 new rows in miner_sector_posts for the same 3 miners which must be from other window post messages in the same tipset.

chain/indexer.go
```diff
@@ -819,6 +836,6 @@ type MessageExecutionProcessor interface {
 type ActorProcessor interface {
 	// ProcessActor processes a set of actors. If error is non-nil then the processor encountered a fatal error.
 	// Any data returned must be accompanied by a processing report.
-	ProcessActors(ctx context.Context, ts *types.TipSet, pts *types.TipSet, actors map[string]types.Actor) (model.Persistable, *visormodel.ProcessingReport, error)
+	ProcessActors(ctx context.Context, ts *types.TipSet, pts *types.TipSet, actors map[string]types.Actor, emsgs []*lens.ExecutedMessage) (model.Persistable, *visormodel.ProcessingReport, error)
```
frrist, line 839: I _almost_ rather just special case the miner actor with its own processor. Does any other actor processor use the `ExecutedMessage` list?

frrist (approved):
Good with the changes as they exist now (love the perf improvement!). I'd also be alright with special casing the miner processor.

chain/indexer.go
iand: No it's the only one. It would be cleaner if we somehow made the messages available through the `lens.Interface`.

The actor extractors themselves now use almost none of the lens interface: just the methods on [ActorStateAPI](https://github.com/filecoin-project/sentinel-visor/blob/iand-missing-block/tasks/actorstate/actorstate.go#L31) which is now just 3 methods with optimization potential plus a way to get a store.

I think there's an opportunity to really simplify this and pass in a type that contains the executed messages and provides access for loading actor state.

frrist: > It would be cleaner if we somehow made the messages available through the lens.Interface

Agreed. I'd like to rework that interface a bit, or at least reduce the number of implementations (to just lily) as well.

I also suspect the `StateMinerPower` method call to be slow as it loads the power actor every time, although maybe there is already some caching implemented there.

iand:
Vectors are highly broken which is tracked in https://github.com/filecoin-project/sentinel-visor/issues/603.

PR #650, opened 2021-08-17
fix: record correct height for messages and block_messages

iand:
Commit 6f9f9998e6094d35cbb81796d5f4259b9b594971 introduced a bug that caused messages and block_messages rows to be written using the height of the receipt tipset instead of the tipset containing the message. This change fixes it by using the height directly associated with the message and/or block.

### DIFF
--- a/tasks/messages/message.go
+++ b/tasks/messages/message.go
@@ -181,7 +181,7 @@ func (p *Task) ProcessMessages(ctx context.Context, ts *types.TipSet, pts *types
 		}
 
 		rcpt := &messagemodel.Receipt{
-			Height:    int64(m.Height + 1),
+			Height:    int64(ts.Height()),
 			Message:   m.Cid.String(),
 			StateRoot: ts.ParentState().String(),
 			Idx:       int(m.Index),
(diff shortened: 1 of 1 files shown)


tasks/messages/message.go
```diff
@@ -181,7 +181,7 @@ func (p *Task) ProcessMessages(ctx context.Context, ts *types.TipSet, pts *types
 		}
 
 		rcpt := &messagemodel.Receipt{
-			Height:    int64(ts.Height()), // this is the child height
+			Height:    int64(m.Height + 1),
```
frrist, line 184: I'm not sure this is right. My concern is:
- Message included in block at height N
- N+1 is a null round
- Message is applied at height N+2

iand: Hmm. I could just leave it as ts.Height, since we know that is the next tipset

frrist: I think that should work -- since we can be certain the receipt was included in the `ts`.

frrist (approved):
Approved assuming https://github.com/filecoin-project/sentinel-visor/pull/650#discussion_r690518776 change will be made

frrist:
I suspect we need a similar change for msapprovals as well.

frrist:
And possibly implicit messages as well.

PR #640, opened 2021-08-12
fix(reports): visor processing reports correspond to tipset passed to indexer

frrist:
This change ensures visor processing report entries correspond to the tipset and state root passed to the indexer. Without this change there exists an annoying edge case in gap filling (#598): e.g. If we find a gap for message task at height 9 and instruct the indexer to index the tipset at height 9 the message processors will end up processing the tipset at height 8.

PR #653, opened 2021-08-17
fix: consistent processing report heights

iand:
Four changes:

 1. actor tasks use parent height for processing report
 2. reverse direction of consensus task processing so that processing report includes one for parent height
 3. move single tipset processors into the indexer's parent/child loop and ensure they use the parent tipset only
 4. start a walk at the tipset after upper bound height specified by `--to` so that the upper bound has a child tipset for context

Fixes https://github.com/filecoin-project/sentinel-visor/issues/652

### DIFF
--- a/chain/indexer.go
+++ b/chain/indexer.go
@@ -183,192 +183,189 @@ func (t *TipSetIndexer) TipSet(ctx context.Context, ts *types.TipSet) error {
 	// A map to gather the persistable outputs from each task
 	taskOutputs := make(map[string]model.PersistableList, len(t.processors)+len(t.actorProcessors))
 
-	// Run each tipset processing task concurrently
-	for name, p := range t.processors {
-		inFlight++
-		go t.runProcessor(tctx, p, name, ts, results)
+	// The parent is the primary tipset that tasks act upon. The child adds additional context such as outcomes of
+	// message execution.
+	var parent, child *types.TipSet
+	if t.lastTipSet != nil {
+		if t.lastTipSet.Height() > ts.Height() {
+			// last tipset seen was the child
+			child = t.lastTipSet
+			parent = ts
+		} else if t.lastTipSet.Height() < ts.Height() {
+			// last tipset seen was the parent
+			child = ts
+			parent = t.lastTipSet
+		} else {
+			log.Errorw("out of order tipsets", "height", ts.Height(), "last_height", t.lastTipSet.Height())
+		}
 	}
 
-	// Run each actor or message processing task concurrently if we have any and we've seen a previous tipset to compare with
-	if len(t.actorProcessors) > 0 || len(t.messageProcessors) > 0 || len(t.messageExecutionProcessors) > 0 || len(t.consensusProcessor) > 0 {
+	// If no parent tipset available then we need to wait for another tipset. It's likely we received the last or first tipset
+	// in a batch.
+	if parent != nil {
 
-		// Actor processors perform a diff between two tipsets so we need to keep track of parent and child
-		var parent, child *types.TipSet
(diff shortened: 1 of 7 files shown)


chain/indexer.go
```diff
@@ -183,192 +183,189 @@ func (t *TipSetIndexer) TipSet(ctx context.Context, ts *types.TipSet) error {
...
-				t.closer = closer
+	// If no parent tipset available then we need to wait for another tipset. It's likely we received the last or first tipset
+	// in a batch.
+	if parent != nil {
+
+		// Run each tipset processing task concurrently
+		for name, p := range t.processors {
+			inFlight++
+			go t.runProcessor(tctx, p, name, parent, results)
+		}
```
iand, line 211: This large diff is simply because I moved lines 186-190 here as lines 207-211

chain/walker.go
```diff
@@ -90,6 +92,10 @@ func (c *Walker) WalkChain(ctx context.Context, node lens.API, ts *types.TipSet)
 			return xerrors.Errorf("get tipset: %w", err)
 		}
 
+		if int64(ts.Height()) < c.minHeight {
```
iand, line 95: Same change as in the last commit on https://github.com/filecoin-project/sentinel-visor/pull/640 and it has the same problem: the consensus task will not detect null rounds if the lower bound of the walk happens to be a null round. Will raise an issue to fix this.

iand:
Results of running walks over 1 and 3 tipset ranges: https://gist.github.com/iand/68f712e2ea2cb53b7223f468e999c2de

chain/indexer.go
```diff
@@ -183,192 +183,189 @@ func (t *TipSetIndexer) TipSet(ctx context.Context, ts *types.TipSet) error {
...
-			} else if t.lastTipSet.Height() < ts.Height() {
-				// last tipset seen was the parent
-				child = ts
-				parent = t.lastTipSet
-			} else {
-				log.Errorw("out of order tipsets", "height", ts.Height(), "last_height", t.lastTipSet.Height())
-			}
+	// The parent is the primary tipset that tasks act upon. The child adds additional context such as outcomes of
+	// message execution.
+	var parent, child *types.TipSet
```
frrist, line 188: What do you think of renaming these?
parent -> current
child -> next

chain/walker.go
```diff
@@ -78,7 +80,7 @@ func (c *Walker) WalkChain(ctx context.Context, node lens.API, ts *types.TipSet)
 	}
```
frrist, line 80: The call above on line 76/78 is basically a noop. All it does now is set the indexer's last tipset to the provided tipset, right?

tasks/actorstate/task.go
```diff
@@ -61,8 +61,8 @@ func (t *Task) ProcessActors(ctx context.Context, ts *types.TipSet, pts *types.T
 	log.Debugw("processing actor state changes", "height", ts.Height(), "parent_height", pts.Height())
 
 	report := &visormodel.ProcessingReport{
-		Height:    int64(ts.Height()),
-		StateRoot: ts.ParentState().String(),
+		Height:    int64(pts.Height()),
+		StateRoot: pts.ParentState().String(),
```
frrist, line 65: Just noting that:
- block processor doesn't need this change since it only acts on a single tipset.
- message processor already uses parent tipset.
- message execution processor uses parent tipset.

frrist (approved):
Looks great, will close #640 in favor of this.

chain/indexer.go
iand: Love it. I'll make that change.

iand: Will also adjust the logging to include this

PR #670, opened 2021-08-18
fix: better error message for "unknown parameters for b" error

iand:
See https://github.com/filecoin-project/sentinel-visor/issues/660 for context. Makes the reported error message more accurate but does not fix the underlying problem.

PR is a little noisy because I renamed ts/pts arguments to next/current for clarity.


iand:
Closing in favor of https://github.com/filecoin-project/sentinel-visor/pull/671

Issue #678, opened 2021-08-25
lily should create its own `lily` repo rather than a `lotus` repo.


frrist:
### Description
Upon intalizing lily the following logs (and actions) are show (performed):
```
╰─>$ ./visor init --import-snapshot https://fil-chain-snapshots-fallback.s3.amazonaws.com/mainnet/complete_chain_with_finality_stateroots_latest.car
2021-08-24T16:58:30.792-0700    INFO    visor/commands  commands/init.go:57     lotus repo: /home/frrist/.lotus
2021-08-24T16:58:30.792-0700    INFO    repo    repo/fsrepo.go:122      Initializing repo at '/home/frrist/.lotus'
```

Instead of creating a `lotus` repo in the home directory, a `lily` repo should be created.

### Acceptance criteria
change the repo file path.

### Where to begin
in the daemon somewhere, we could use a default flag value for the `--repo` flag.

PR #681, opened 2021-08-25
fix: config file is written with working defaults

frrist:
### What changes?
`visor init` will write the config file with `[Storage]` values by default. Previously this never worked.
The default config file will work once uncommented. Previously this didn't work.
`visor daemon` (with or without the config flag) will not write the `[Storage]` values.
`visor daemon --default-config=true` (defaults false) will overwrite the config file with `[Storage]` values included.

### DIFF
--- a/commands/daemon.go
+++ b/commands/daemon.go
@@ -3,7 +3,6 @@ package commands
 import (
 	"context"
 	"io/ioutil"
-	"path/filepath"
 
 	paramfetch "github.com/filecoin-project/go-paramfetch"
 	lotusbuild "github.com/filecoin-project/lotus/build"
@@ -57,10 +56,11 @@ var clientAPIFlagSet = []cli.Flag{
 }
 
 type daemonOpts struct {
-	repo      string
-	bootstrap bool // TODO: is this necessary - do we want to run visor in this mode?
-	config    string
-	genesis   string
+	repo          string
+	bootstrap     bool
+	config        string
+	commentConfig bool
+	genesis       string
 }
 
 var daemonFlags daemonOpts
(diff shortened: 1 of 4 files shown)


commands/daemon.go
```diff
@@ -127,19 +127,25 @@ Note that jobs are not persisted between restarts of the daemon. See
...
 			EnvVars:     []string{"VISOR_CONFIG"},
+			Value:       "~/.lotus/config.toml",
 			Destination: &daemonFlags.config,
 		},
+		&cli.BoolFlag{
+			Name:        "default-config",
+			Usage:       "If true the daemon config file will be overwritten with default configuration values.",
+			Value:       false,
+			Destination: &daemonFlags.commentConfig,
+		},
```
placer14, line 148: nit: I'd personally prefer it if bootstrapping were done in `lily init` to keep concerns tidy, unless there's a good reason to have it here.

iand:
>     * run



> I guess I just find it offputting that:
> 
>     * If config file isn't created (via `init --config`) before the daemon command is run, the daemon will create a config file missing all the `[Storage]` bits.

Init doesn't write working storage bits - they are commented examples to help document what is needed. Running without init could do the same.

> 
>     * if the `--config` flag isn't provided to `init` it will create a config anyways, but without the `[Storage]` bits.
> 

It could.



> 
> This isn't how I would expect this to work if I had worked with lotus or ipfs before.

Surprising because most of it is copied from Lotus!

commands/daemon.go
frrist: Agreed. I think the `bootstrap` value would need to be persisted somewhere since it's used when constructing the node:
https://github.com/filecoin-project/lily/blob/2eb534750326b5dead21e94e90259260b32a94ec/commands/daemon.go#L255-L258

frrist:
I've filed the following to track the work discussed here: 
https://github.com/filecoin-project/lily/issues/693
https://github.com/filecoin-project/lily/issues/694
https://github.com/filecoin-project/lily/issues/695
https://github.com/filecoin-project/lily/issues/678
going to close this out and address them individually.

PR #688, opened 2021-08-27
docs: update change log v0.8.0

frrist:
waiting on:
~https://github.com/filecoin-project/lily/pull/689 & https://github.com/filecoin-project/lily/pull/681 & #696~


placer14 (approved):

iand (commented):
Suggest you write a sentence or two noting that this is quite a major change for visor: renaming and removal of all lenses so it no longer works with a separate Lotus node or a car file.

PR #690, opened 2021-08-28
Fix migrate command

f8-ptrk:
the `--db` flag is a part of `lily migrate` - not lily

the command as is fails. 

Visor version:v0.7.7+16-gb048a53


frrist (approved):
thanks @f8-ptrk :heart:

PR #707, opened 2021-09-10
Fix partial migration check

### LINKED ISSUE #706: Migration fails from v1.0 to v1.3
### Describe the bug:

When upgrading database created w visor prior to v0.7.5 and attempting to run `migrate --latest` w v0.7.7 or later results in the following error output:

```
[21:53:56] $ ./visor migrate --db postgres://postgres:password@localhost:5432/postgres?sslmode=disable --latest
2021-09-09T21:53:58.915-0400    INFO    visor/commands  commands/setup.go:161   Visor version:v0.7.7
2021-09-09T21:53:58.957-0400    INFO    visor/storage   storage/migrate.go:194  current database schema is version 1.0
2021-09-09T21:53:58.961-0400    INFO    visor/storage   storage/migrate.go:230  creating base schema for major version 1
creating base schema: ERROR #42710 type "miner_sector_event_type" already exists
```

Note: This is specific to running a partial migration. Running a migration from a clean DB works fine:

```
[21:51:42] $ ./visor migrate --db postgres://postgres:password@localhost:5432/postgres?sslmode=disable --latest
2021-09-09T21:51:51.827-0400    INFO    visor/commands  commands/setup.go:161   Visor version:v0.7.7
2021-09-09T21:51:51.861-0400    INFO    visor/storage   storage/migrate.go:194  current database schema is version 0.0
2021-09-09T21:51:51.887-0400    INFO    visor/storage   storage/migrate.go:230  creating base schema for major version 1
2021-09-09T21:51:52.305-0400    INFO    visor/storage   storage/migrate.go:267  running schema migration from version 1.0 to version 1.3
2021-09-09T21:51:52.377-0400    INFO    visor/storage   storage/migrate.go:275  current database schema is now version 1.3
```

### Steps to Reproduce:

```sh
# remove old DB state
$ docker volume rm visor_timescaledb
$ docker-compose up timescaledb
$ git checkout v0.7.5
$ make visor
$ ./visor migrate --db postgres://postgres:password@localhost:5432/postgres?sslmode=disable --latest
$ git checkout v0.7.7 # also fails for v0.8.0
$ make visor
$ ./visor migrate --db postgres://postgres:password@localhost:5432/postgres?sslmode=disable --latest
# observe error
```

placer14:
When running migrations over a partially completed schema, the base schema creation is being attempted again due to an incorrect check.

~I've also moved a defer closer to the lock it grabs to avoid held locks if base creation fails. This is probably never a concern as the schema wouldn't be shared without the base, but it would be annoying for the lock to be held from a prior failure and then deadlock waiting to apply patches. LMK if I'm misreading this.~ Removed.

fixes #706

### DIFF
--- a/storage/migrate.go
+++ b/storage/migrate.go
@@ -219,7 +219,7 @@ func (d *Database) MigrateSchemaTo(ctx context.Context, target model.Version) er
 	}
 
 	// Check if we need to create the base schema
-	if dbVersion.Patch == 0 {
+	if !initialized {
 		log.Infof("creating base schema for major version %d", target.Major)
 
 		base, err := baseForVersion(target, d.SchemaConfig())


storage/migrate.go
```diff
@@ -214,12 +214,20 @@ func (d *Database) MigrateSchemaTo(ctx context.Context, target model.Version) er
...
+		}
+	}()
+
 	if err := initDatabaseSchema(ctx, db, d.SchemaConfig()); err != nil {
 		return xerrors.Errorf("initializing schema version tables: %w", err)
 	}
 
 	// Check if we need to create the base schema
-	if dbVersion.Patch == 0 {
+	if dbVersion.Major == 0 {
```
iand, line 230: This is not correct. Major is the version of the schema in use. Each major version will have its own base schema on which patches are applied.

iand (changes requested):

iand:
If you are seeing the base schema being created again then something else must be wrong. Are you specifying the correct schema name when running lily? Are the 2 version tables present in  that schema (`gopg_migrations` and `visor_version`) and what do they contain?  Perhaps visor version table has been renamed - I have not checked.

iand:
Removed the hold I placed on this in my review since I am OOO after today and don't want to block. The intent is still there though - the if clause is correct as stands.

frrist:
Contents of tables:
visor_version:
```
+-----+
|major|
+-----+
|1    |
+-----+
```
gopg_migrations:
```
+--+-------+--------------------------+
|id|version|created_at                |
+--+-------+--------------------------+
|1 |1      |2021-09-10 12:33:16.824088|
|2 |2      |2021-09-10 12:33:17.162100|
|3 |3      |2021-09-10 12:33:17.514487|
+--+-------+--------------------------+

```

iand:
Current version of that schema is 1.3 according to those tables.

storage/migrate.go
frrist: I followed the reproduction steps in #706 and observed the base schema to be erroneously reapplied, this is the 🐛. I believe the issue here is using a lack of patch versions to determine if the database is initialized, which is wrong since the base schema can exist without patch versions (as it does going from v0.7.5 -> beyond). I think the fix here is the following:
```suggestion
if !initialized {
```

`initialized` is true iff the Database has the version tables **and** they are populated, meaning the base schema has been applied.

frrist (approved):

Issue #705, opened 2021-09-09
Indexer off-by-one error when used by Watcher


frrist:
### Describe the bug:
Relates to https://github.com/filecoin-project/lily/issues/704.
The problem boils down to how the Watcher causes skips to occur in the indexer. Its easiest to describe with an example. But it's also worth pointing out this behavior is reflected in the database: 
![image](https://user-images.githubusercontent.com/6546409/132745864-8851e927-fb03-422c-acf3-e2635a13077e.png)

### Steps to Reproduce:
Indexer processed tipset N-1
  - current: null
  - next: null
  
Indexer is currently indexing tipset N
 - current: N-1
 - next: N
 
Watcher receives TipSet N+1 but the indexer isn't ready therefore tipset N+1 is skipped
- current: N-1
- next: N

Indexer receives tipset N+2
- current: N
- next: N+2

N+2 is also skipped since it's parents are unknown to indexer: https://github.com/filecoin-project/lily/blob/dd9645bd5562fa91e2c78089cabffa9ce60f8ccb/chain/indexer.go#L243

Issue #704, opened 2021-09-09
Watcher does not respect Window config


frrist:
### Describe the bug:
If the Watcher's `indexSlot` is unavailable when a new TipSet is ready for indexing, the Watcher will Skip the TipSet currently being processed by the indexer. This means indexing timeouts are triggered by the arrivals of a new TipSet instead of the Window config value used for timeouts.
https://github.com/filecoin-project/lily/blob/ba4c266e345c7f0018bd609c566d19324874e587/chain/watcher.go#L111-L132

PR #710, opened 2021-09-10
fix: watcher respects window timeout value

frrist:
First pass at addressing #704 and #705


chain/indexer.go
```diff
@@ -744,10 +768,18 @@ func (t *TipSetIndexer) SkipTipSet(ctx context.Context, ts *types.TipSet, reason
 		reports = append(reports, t.buildSkippedTipsetReport(ts, name, timestamp, reason))
 	}
 
+	for name := range t.messageExecutionProcessors {
```
frrist, line 771: These were missing, adding them for completeness

chain/watcher.go
```diff
@@ -108,27 +110,12 @@ func (c *Watcher) maybeIndexTipSet(ctx context.Context, ts *types.TipSet) error
...
-		// Indexing slot was available which means we can continue.
-		go func() {
-			// Clear the slot when we have completed indexing
-			defer func() {
-				<-c.indexSlot
-			}()
-
+	default:
+		c.wp.Submit(func() {
 			if err := c.obs.TipSet(ctx, ts); err != nil {
```
frrist, line 115: The window value passed to the indexer will cause this method to exit when the timeout occurs.

chain/watcher.go
```diff
@@ -108,27 +110,12 @@ func (c *Watcher) maybeIndexTipSet(ctx context.Context, ts *types.TipSet) error
...
 				log.Errorw("failed to index tipset", "error", err, "height", ts.Height())
 			}
-		}()
-	default:
-		// The indexer is taking longer than one epoch to process. We need to avoid blocking the stream of incoming
-		// tipsets otherwise we will cause the node to fall behind the chain while it waits for us to catch up
-		// (which may never happen if we consistently take too long)
-		log.Errorw("skipping tipset since indexer is not ready", "height", ts.Height())
-		stats.Record(ctx, metrics.TipSetSkip.M(1))
-		if err := c.obs.SkipTipSet(ctx, ts, "indexer not ready"); err != nil {
```
frrist, line 129: Calling SkipTipSet from outside the indexer probably isn't a good idea since it can cause off by one error as outlined in #705. Skip could also set `lastTipSet` when called.

PR #729, opened 2021-10-06
chore(docker): Include jq in docker images; Fix dev entrypoint

placer14:


Dockerfile
```diff
@@ -40,5 +40,9 @@ COPY --from=builder /usr/lib/x86_64-linux-gnu/libhwloc.so* /lib/
 COPY --from=builder /usr/lib/x86_64-linux-gnu/libnuma.so* /lib/
 COPY --from=builder /usr/lib/x86_64-linux-gnu/libltdl.so* /lib/
 
+RUN apt-get update
+RUN apt-get install -y --no-install-recommends \
```
frrist, line 44: Looks like this is already being done above on line 19.

Dockerfile.dev
```diff
@@ -25,12 +25,16 @@ WORKDIR /go/src/github.com/filecoin-project/lily
...
-RUN go mod download -x
+RUN go mod download
 RUN make $LILY_NETWORK_TARGET
 RUN cp ./lily /usr/bin/
 
 # build/docker/dev_entrypoint.tpl
 # partial for completing a dev lily dockerfile
 
+RUN apt-get update
+RUN apt-get install -y --no-install-recommends \
```
frrist, line 36: ditto

Dockerfile
placer14: This one is needed because of the new image being created. Line 19 installs jq inside `builder` but not the final image.

frrist: Alternatively, the `jq` bin could be copied into the new image.

placer14: I believe `jq` is written in C and it's possible that it dynamically links in other libs which are expected to be in the environment. But I'll try it out.

PR #734, opened 2021-10-08
update to lotus 1.13

frrist:
gets us ready for lotus 1.13 when it lands
TODO:
- [x] update to specs-actors v6 when cut
- [x] update to 1.13 (instead of RC) lotus release when cut
- [x] validate lily can sync with the calibration network

### DIFF
--- a/commands/setup.go
+++ b/commands/setup.go
@@ -13,6 +13,7 @@ import (
 	metricsprom "github.com/ipfs/go-metrics-prometheus"
 	_ "github.com/lib/pq"
 	prom "github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/collectors"
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/zpages"
 	"go.opentelemetry.io/otel/exporters/trace/jaeger"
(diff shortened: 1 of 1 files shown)


chain/actors/builtin/miner/diff.go
```diff
@@ -4,7 +4,8 @@ import (
 	"context"
 
 	cbg "github.com/whyrusleeping/cbor-gen"
-	"go.opentelemetry.io/otel/api/global"
```
frrist, line 7: the bulk of changes in PR are these: upgrading from v0.12 to v0.20 of opentelemetry

iand: Wonder if we could do these changes first in a separate PR?

frrist: Good idea: https://github.com/filecoin-project/lily/pull/749

commands/daemon.go
```diff
@@ -238,7 +238,7 @@ Note that jobs are not persisted between restarts of the daemon. See
 
 			node.Override(new(dtypes.Bootstrapper), isBootstrapper),
 			node.Override(new(dtypes.ShutdownChan), shutdown),
-			node.Online(),
+			node.Base(),
```
placer14, line 241: What is the functional difference between the old and new injection?

commands/lily.go
```diff
@@ -43,6 +43,7 @@ func LilyNodeAPIOption(out *lily.LilyAPI, fopts ...node.Option) node.Option {
 	resAPI := &lily.LilyNodeAPI{}
 	opts := node.Options(
 		node.WithRepoType(repo.FullNode),
+		node.WithEnableLibp2pNode(true),
```
placer14, line 46: Any reason this is now required?

commands/daemon.go
frrist: Before: https://github.com/filecoin-project/lotus/blob/v1.11.0/node/builder.go#L457-L473
After: https://github.com/filecoin-project/lotus/blob/v1.13.0/node/builder.go#L223-L235
A name change and a different way of enabling the libp2p bits.

PR #739, opened 2021-10-12
feat: implement TipSetCache warming

frrist:


chain/tipset.go
```diff
@@ -157,6 +163,32 @@ func (c *TipSetCache) Reset() {
 	c.len = 0
 }
 
+// Warm fills the TipSetCache with confidence tipsets so that subsequent calls to Add return a tipset.
+func (c *TipSetCache) Warm(ctx context.Context, head *types.TipSet, fn func(ctx context.Context, tsk types.TipSetKey) (*types.TipSet, error)) error {
```
placer14, line 167: nit: A slightly better arg name?

```suggestion
func (c *TipSetCache) Warm(ctx context.Context, head *types.TipSet, tipsetGetter func(ctx context.Context, tsk types.TipSetKey) (*types.TipSet, error)) error {
```

chain/tipset.go
```diff
@@ -157,6 +163,32 @@ func (c *TipSetCache) Reset() {
...
+	cur := head
+	tss := make([]*types.TipSet, 0, c.Confidence())
+	for i := 0; i < c.Confidence(); i++ {
+		cur, err := fn(ctx, cur.Parents())
+		if err != nil {
+			return err
+		}
+		tss = append(tss, cur)
+		if cur.Height() == 0 {
```
iand, line 176: For safety you should probably make this check at the start of the loop. head may be genesis

iand (approved):

PR #748, opened 2021-10-18
fix: daemon api uses full flag value

frrist:
makes all commands work as expect with the API flag set as opposed to the daemon flag only expecting a port and other commands expecting a full multiaddress.

Changes doesn't interfere with docs: https://github.com/filecoin-project/sentinel/blob/master/docs/lily/operation.md#api-endpoint


placer14:
Let's do pass over the docs (internal help and sentinel docs page) to make sure we don't need to update anything.

iand (approved):

PR #749, opened 2021-10-19
deps: update otel v0.20.0 and pg/v10 v10.10.6

frrist:


placer14:
@frrist Did you happen to check whether otel tracing worked before this change?

PR #750, opened 2021-10-19
cli: implement chain set head

frrist:
useful for unwedging a node, copied from lotus


iand (approved):
Approved but I hate the cli ux. 3 different ways to set the head split across 2 flags and an argument.

PR #753, opened 2021-10-21
Revert "feat: add observe command with peer agents task (#577)"

frrist:
This reverts commit 41e4fcd7d0fb0577ea14dd24d3ddc7d960b46ad0.

I mistakenly merged this before verifying there was a schema migration written. :facepalm:

PR #754, opened 2021-10-21
fix: add missing observed_peer_agent schema

frrist:
Can either take https://github.com/filecoin-project/lily/pull/753 and reapply the change, or merge this in and call it good.


placer14:
I'm okay with reverting now. I have some changes I want to put up and cut a release next week too. So we can add it then?

PR #755, opened 2021-10-21
feat: add survey command with peer agents task

iand:
This is a local copy of my branch that was corrupted in #577, rebased onto master


lens/lily/impl.go
```diff
@@ -326,6 +327,34 @@ func (m *LilyNodeAPI) Shutdown(ctx context.Context) error {
...
+	// instantiate a new surveyer.
+	surv, err := network.NewSurveyer(m, strg, cfg.Interval, cfg.Name, cfg.Tasks)
+	if err != nil {
+		return nil, err
+	}
+
+	res := m.Scheduler.Submit(&schedule.JobConfig{
+		Name:                cfg.Name,
+		Tasks:               cfg.Tasks,
+		Job:                 surv,
```
frrist, line 349: `cfg.Interval` should be listed as a param.

frrist (approved):
Sorry about the corruption I caused, thanks for refile-ing :pray:

lens/lily/impl.go
frrist: addressed this via bc803e9

PR #756, opened 2021-10-21
chore: update message parsing to use latest statediff generator

iand:
Draft until https://github.com/filecoin-project/statediff/pull/265 is merged (need to update support/tools/go.mod to reference the final commit)


frrist (approved):
Looks like most of this was code gen?

Issue #767, opened 2021-11-18
gapfind vs consensus NULL_ROUND


placer14:
I'm seeing both `gap_find` and `consensus` reports showing NULL_ROUND and wonder the following:
- Are these NULL_ROUNDs equivalent?
- If so, can we converge their tasks to have the same?

Motivations
- Having different representations for the same thing is a smell.
- I don't think gap_find should touch both tables. It would be a better separation of concerns for consensus to just do that and let gapfind leave a hint in gap reports for the consensus task to check when/if it runs.

Aside: The only obvious difference is that `gap_find` does not have a state_root which would be valuable. So if we make gap_find say `consensus` instead, I'd want state_root to also be populated.

iand:
By reports I assume you mean table?

What do you mean by converging the tasks? For clarity on terminology: 

 - `consensus` is a task used by a `walk` or `watch` job
 - `find` is a job that is used to find gaps that should be filled

If no `walk` or `watch` job has run the `consensus` task over a range of epochs then there will be missing data some of which may be null rounds. Therefore the `find` job needs to examine the chain across those missing epochs to determine whether each height should have data (a gap) or not (a null round).

Additionally, when the chain forks there may be different null rounds on each branch and a `watch` with low confidence could record either or both of these.

On that basis I think we need to keep gap finder's concept of null rounds distinct from consensus.

iand:
A follow up thought. I question why gap finding writes to `visor_processing_reports` at all. It could write the null round information to `visor_gap_reports` instead.

placer14:
My main concern is that gapfind can insert a single `NULL_ROUND` into processing reports with the task as `gap_find`. This is inconsistent with `gap_find`'s other behavior and creates a footgun in the representation of data in processing reports (having to require a special condition for the `gap_find` case is going to be surprising to other users not intimately aware that processing reports might have a job name (`gap find`) instead of only task names).

My preference would be that:
- gap find leaves a hint in the gap reports and
- tasks (regardless of gap/walk/watch job) produce `NULL_ROUND` rows in processing reports for themselves

frrist:
via https://github.com/filecoin-project/lily/pull/884. This behavior is no longer present

PR #769, opened 2021-11-18
Simplify gapfind hole detection

### LINKED ISSUE #768: gapfind produces incorrect set of heights,tasks to fill
Trying to reconcile an internal "lily data completeness" query, I believe I've identified some cases where gapfind does not detect holes and incorrectly produces holes which don't need filling.

Within `chain/find.go` you execute `GapIndexer.findTaskEpochGaps` w the following query (which I've refactored without adjusting the behavior for clarity with comments included inline indicating the logic bugs):

```sql
with
all_heights_and_tasks as (
	select height, task
    from visor_processing_reports
    where height between ? and ?
	group by 1, 2
)
, incomplete_heights as (
	select height, count(task)
	from all_heights_and_tasks
	group by 1
        -- this incorrectly handles NULL_ROUND heights which
        -- only have a single task (gap_find or consensus)
        -- PS: 14 comes from len(AllTasks) in the calling go function
	having count(task) != 14
)
, incomplete_heights_and_their_completed_tasks as (
	select ih.height, pr.task, pr.status_information
	from incomplete_heights ih
	left join visor_processing_reports pr using (height)

	-- this condition excludes NULL_ROUND tasks
	-- which are completed and should be included
	where pr.status_information != 'NULL_ROUND'

	-- what is this filter for?
	-- this will accidentally include rows `ih`
	-- which don't have a joined row on `pr` (which
	-- happens when there's no task history on those heights)
	-- creating heights with no completed tasks at all
	or pr.status_information is null
	group by 1, 2, 3
)
select height, task
from incomplete_heights_and_their_completed_tasks
order by height desc
;
```

placer14:
- removed bugs around edges cases
- this includes the diffing logic within the query which was previously happening after the query in go
- this logic appears to be complete according to spot checks in both directions (ensure this query points at heights that actually contain gaps as well as gaps manually perceived in visor reports which appear in the results of the query)
- gap find no longer writes NULL_ROUNDs to the processing reports table (this separates the concerns of fill and find more clearly, removes special cases for easier querying going forward, and had the side-effect of making test logic simpler)
- this speeds up execution (casual observations suggest around 10x improvement)

fixes #768

### DIFF
--- a/chain/find.go
+++ b/chain/find.go
@@ -258,9 +258,9 @@ all_heights_and_tasks_in_range as (
 ,
 
 -- all heights from processing reports which were
--- recorded (by gap_fill or consensus) that it is
--- a null round with no data to index.
--- then take cross product of these heights w tasks
+-- recorded (by gap_fill or consensus) and are
+-- null rounds with no data to index.
+-- take cross product of these heights w tasks
 null_round_heights_and_tasks_in_range as (
 	select pr.height, t.task
 	from visor_processing_reports pr
@@ -274,22 +274,22 @@ null_round_heights_and_tasks_in_range as (
 -- all heights and tasks which need to be filled
 all_incomplete_heights_and_tasks as (
 
-	select height, task
 		-- starting from the set of all heights and tasks
 		-- in our range
-    from all_heights_and_tasks_in_range
+		select height, task
+		from all_heights_and_tasks_in_range
 
-    -- remove all heights and tasks which have at least one OK
-    except
-    select height, task
-    from visor_processing_reports
-    where status = ?3
+		-- remove all heights and tasks which have at least one OK
+		except
+		select height, task
+		from visor_processing_reports
+		where status = ?3
 		and height between ?0 and ?1
 
-    -- remove the null rounds by height and task
-    except
-    select height, task
-    from null_round_heights_and_tasks_in_range
+		-- remove the null rounds by height and task
+		except
+		select height, task
+		from null_round_heights_and_tasks_in_range
 )
 
 -- ordering for tidy persistence
--- a/chain/find.go
+++ b/chain/find.go
@@ -48,7 +48,7 @@ func NewGapIndexer(node lens.API, db *storage.Database, name string, minHeight, 
 }
 
 func (g *GapIndexer) Run(ctx context.Context) error {
-	startTime := time.Now()
+	log.With("type", "find")
 
 	head, err := g.node.ChainHead(ctx)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


chain/find.go
```diff
@@ -197,90 +199,132 @@ type TaskHeight struct {
...
+
+-- generate all heights in range
+interesting_heights as (
+	select *
+	from generate_series(
+		?0,
+		?1
+	)
+	as x(height)
+)
```
iand, line 226: :heavy_check_mark:

chain/find.go
```diff
@@ -197,90 +199,132 @@ type TaskHeight struct {
...
+		-- ('actorstatesmultisig'),
+		-- ('actorstatesverifreg'),
+		-- ('blocks'),
+		-- ('messages'),
+		-- ('chaineconomics'),
+		-- ('msapprovals'),
+		-- ('implicitmessage'),
+		-- ('consensus')
+	) as x(task)
+)
```
iand, line 249: :heavy_check_mark:

chain/find.go
```diff
@@ -197,90 +199,132 @@ type TaskHeight struct {
...
+-- a null round with no data to index.
+-- then take cross product of these heights w tasks
+null_round_heights_and_tasks_in_range as (
+	select pr.height, t.task
+	from visor_processing_reports pr
+	cross join interesting_tasks t
+	where pr.status_information = ?3
+	and pr.height between ?0 and ?1
+	group by 1, 2
+)
```
iand, line 271: :heavy_check_mark:

chain/find.go
```diff
@@ -197,90 +199,132 @@ type TaskHeight struct {
...
+    select height, task
+    from visor_processing_reports
+    where status = ?3
+		and height between ?0 and ?1
+
+    -- remove the null rounds by height and task
+    except
+    select height, task
+    from null_round_heights_and_tasks_in_range
+)
```
iand, line 293: :heavy_check_mark:

chain/find.go
```diff
@@ -197,90 +199,132 @@ type TaskHeight struct {
...
+	) as x(task)
+)
+,
+
+-- cross product of heights and tasks
+all_heights_and_tasks_in_range as (
+	select h.height, t.task
+	from interesting_heights h
+	cross join interesting_tasks t
+)
```
iand, line 257: :heavy_check_mark:

iand (approved):
LGTM. The :heavy_check_mark:  are just where I was thinking through each stage of the query

placer14:
I've been poking at this query and it seems to be pretty solid for the expected behavior we're looking for.
- I've ensured that there are no rows from the original query (as well as our internal "completeness" report's query) which are missed by the new one
- I've done some spot checking on the false negatives missed by the original query and included in the new one to ensure they are accurate

I've also fixed some tests which were incorrectly ignoring the null round height (which should have 14 missing tasks each being returned by the find function).

placer14:
I found [this bug](https://github.com/filecoin-project/lily/issues/773) and the above query fixes does resolve this case. Will make sure I add a unit test for this. 👏

chain/find.go
```diff
@@ -56,231 +58,158 @@ func (g *GapIndexer) Run(ctx context.Context) error {
...
-	gapReport := make([]*visor.GapReport, 0, len(missingHeights))
-	// walk the possible gaps and query lotus to determine if gap was a null round or missed epoch.
-	for _, gap := range missingHeights {
-		select {
-		case <-ctx.Done():
-			return nil, nil, ctx.Err()
-		default:
-		}
-		gh := abi.ChainEpoch(gap)
-		tsgap, err := node.ChainGetTipSetByHeight(ctx, gh, types.EmptyTSK)
```
frrist, line 171: This changes the behavior of gap find by causing it to label undetected null rounds as gaps. I think it will be better to address this as a separate change taking into consideration the conversation here: https://github.com/filecoin-project/lily/issues/767

chain/find_test.go
```diff
@@ -8,15 +8,12 @@ import (
 	"testing"
 	"time"
 
-	"github.com/filecoin-project/go-address"
```
frrist, line 11: These changes should be removed and addressed separately in a different PR that fixes https://github.com/filecoin-project/lily/issues/767

frrist (commented):
Good stuff. 
I think this PR will be most impactful if it focuses on improving the find query, fixing https://github.com/filecoin-project/lily/issues/773, and adding a unit test for it (see https://github.com/filecoin-project/lily/pull/774).

placer14:
This PR has served it's purpose. Closing and breaking into smaller pieces.

PR #774, opened 2021-11-23
fix(find): find only consideres lack of OK's when finding gaps

### LINKED ISSUE #773: Gaps reported for epochs,tasks which already have an OK
After doing some exploration in our local DB instance, we see many cases where a `gap find` produces a GAP report for epoch,tasks already have OK processing records. (Specifically, the OK existed and still produced a GAP report after the fact.) Notably, all of these observed rows also contain rows with SKIPs, which may be related.

Here are some records I found in our production DB to illustrate (query at the end): https://gist.github.com/placer14/df45f16be350ae7c3223584fe10a6039

frrist:
- closes #773 and adds a test.

https://github.com/filecoin-project/lily/pull/769 address this bug, but is slightly larger in scope. We can take this as is or incorporate into #769

### DIFF
--- a/chain/find.go
+++ b/chain/find.go
@@ -45,6 +45,30 @@ func NewGapIndexer(node lens.API, db *storage.Database, name string, minHeight, 
 	}
 }
 
+func (g *GapIndexer) findGapsAndNullRounds(ctx context.Context) (visor.GapReportList, []abi.ChainEpoch, error) {
+	findLog := log.With("type", "find")
+
+	var gaps []*visor.GapReport
+	// looks for incomplete epochs. An incomplete epoch has some, but not all tasks in the processing report table.
+	taskGaps, err := g.findTaskEpochGaps(ctx)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


chain/find.go
```diff
@@ -45,40 +45,45 @@ func NewGapIndexer(node lens.API, db *storage.Database, name string, minHeight,
...
-
-	head, err := g.node.ChainHead(ctx)
-	if err != nil {
-		return err
-	}
-	if uint64(head.Height()) < g.maxHeight {
-		return xerrors.Errorf("cannot look for gaps beyond chain head height %d", head.Height())
-	}
-
+func (g *GapIndexer) findGapsAndNullRounds(ctx context.Context) (visor.GapReportList, []abi.ChainEpoch, error) {
```
placer14, line 48: I don't think you meant to change this from `Run` to `findGapsAndNullRounds` did you?

placer14: Ugh, I see it now. Sry, thx.

placer14:
Superseded by #778 which fixes the behavior in refactoring. New unit test is included in #778. Closing.

PR #778, opened 2021-11-24
chore(find): Refactor GapIndexer.findMissingTasksAndEpochs

### LINKED ISSUE #773: Gaps reported for epochs,tasks which already have an OK

placer14:
Does not change the behavior of the tests, only tightens the expectations to only the subjects being tested.

Refactor happens to fix #773 as well, so this test is included.

### DIFF
--- a/chain/find.go
+++ b/chain/find.go
@@ -2,14 +2,14 @@ package chain
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"time"
 
 	"github.com/deckarep/golang-set"
-	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/lily/lens"
 	"github.com/filecoin-project/lily/model/visor"
 	"github.com/filecoin-project/lily/storage"
-	"github.com/filecoin-project/lotus/chain/types"
 	"golang.org/x/xerrors"
 )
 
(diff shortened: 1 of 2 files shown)


chain/find.go
```diff
@@ -58,229 +56,159 @@ func (g *GapIndexer) Run(ctx context.Context) error {
...
-	gapReport := make([]*visor.GapReport, 0, len(missingHeights))
-	// walk the possible gaps and query lotus to determine if gap was a null round or missed epoch.
-	for _, gap := range missingHeights {
-		select {
-		case <-ctx.Done():
-			return nil, nil, ctx.Err()
-		default:
-		}
-		gh := abi.ChainEpoch(gap)
-		tsgap, err := node.ChainGetTipSetByHeight(ctx, gh, types.EmptyTSK)
```
frrist, line 171: This shouldn't be removed as doing so changes the behavior of the implementation. Gap find should probably continue to be aware of null rounds vs. gaps. A solution is outlined here: https://github.com/filecoin-project/lily/issues/767#issuecomment-972916153. I'd be fine with that landing as a part of this work, or as a follow-on.

frrist: As a part of this we should also label gaps resulting from undetected null rounds as `NULLROUND-GAP` or similar.

placer14:
This work will be re-applied along with the TBD plan for #767. Closing as it is no longer needed.

PR #779, opened 2021-11-24
feat(find): Include ERROR gaps in find gap reports

### LINKED ISSUE #775: fill/find behavior around ERRORing epoch-task reports
There is currently an implicit behavior in `find` to omit reports which have resulted in an `ERROR` status. The rationale being that these will likely always produce the same `ERROR`ing result and we should save lily from performing the same work again with no benefit. However, I’ve noticed that there are processing records with `ERROR`s which also show as `FILLED` in the gap reports. This suggests that some errors could be transient and the premise that performing the same work again will result in the same outcome are not true.

Further, `ERROR` reports are effectively gaps in dataset and `find` should not treat these reports any differently from other types of gaps. If the cause of an `ERROR` is eventually resolved, using `find` to identify that epoch-task as a candidate to later `fill` is required to fix the bug in our data (and not just in the software).

placer14:
fixes #775
includes test from #774

(includes changes in #778 as they are dependent)

### DIFF
--- a/chain/find.go
+++ b/chain/find.go
@@ -165,14 +165,6 @@ all_incomplete_heights_and_tasks as (
 	-- remove the null rounds by height and task
 	select height, task
 	from null_round_heights_and_tasks_in_range
-
-		except
-
-	-- remove all errors (to match the original behavior)
-	select height, task
-	from visor_processing_reports
-	where status = ?4
-	and height between ?0 and ?1
 )
 
 -- ordering for tidy persistence
--- a/chain/find.go
+++ b/chain/find.go
@@ -181,7 +181,6 @@ order by 1 desc
 		g.maxHeight,
 		visor.ProcessingStatusInformationNullRound,
 		visor.ProcessingStatusOK,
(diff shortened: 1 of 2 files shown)


frrist:
Can you base this on https://github.com/filecoin-project/lily/pull/778 to make reviewing easier, please?

placer14:
Not unless we remember to merge commit that PR instead of squash commit (which will produce a new SHA and won't merge cleanly).

frrist:
This is a follow on to #778, right?

frrist (commented):
good stuff, I think the only thing missing is a flag on the the gap fill command to control whether or not error-gaps are filled, since we will only want to run those fills when the bugs are fixed. eg
`gap fill --fill-errors=true --tasks=messages --from=100 -- to=200`

placer14:
This is no longer needed as it is being included in https://github.com/filecoin-project/lily/pull/791. (Albeit without the switch that @frrist wanted. We can add this switch separately.) Closing this.

PR #785, opened 2021-11-30
feat(job): add job waiting option

### LINKED ISSUE #723: feat: blocking job wait for scripting (lily wait job --id N)
### Description

I was trying to automate the successive execution of multiple jobs. I ended up with a brittle bash/jq script which depends on the job list output not changing. Would be really useful if we could script something like `lily job wait --id N` which would block until the job completed. Even better if it spits out the job status once it's done so things like `Error` could be inspected.

frrist:
- closes #723
- adds a new test `Job Submit and wait`
- refactors the scheduler test, can add as a separate change if prefers.

### DIFF
--- a/schedule/scheduler.go
+++ b/schedule/scheduler.go
@@ -26,6 +26,7 @@ type Job interface {
 	// retry the job so implemententions must ensure that Run resets any
 	// necessary state.
 	Run(context.Context) error
+	Done() <-chan struct{}
 }
 
 type JobConfig struct {
@@ -247,12 +248,9 @@ func (s *Scheduler) Run(ctx context.Context) error {
 }
 
 func (s *Scheduler) StartJob(id JobID) error {
-	s.jobsMu.Lock()
-	defer s.jobsMu.Unlock()
-
-	job, ok := s.jobs[id]
-	if !ok {
-		return xerrors.Errorf("starting worker ID: %d not found", id)
+	job, err := s.getJob(id)
+	if err != nil {
+		return xerrors.Errorf("start job: %w", err)
 	}
 
 	job.lk.Lock()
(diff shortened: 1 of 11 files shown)


schedule/scheduler.go
```diff
@@ -290,6 +291,40 @@ func (s *Scheduler) StopJob(id JobID) error {
 	return nil
 }
 
+func (s *Scheduler) WaitJob(id JobID) (*JobListResult, error) {
+	s.jobsMu.Lock()
+	job, ok := s.jobs[id]
+	if !ok {
+		return nil, xerrors.Errorf("wait job ID: %d not found", id)
```
iand, line 298: Unlock `jobsMu` before return. Can just move the existing `Unlock` to just before the `if`

schedule/scheduler.go
```diff
@@ -290,6 +291,40 @@ func (s *Scheduler) StopJob(id JobID) error {
...
+	// wait on the job to complete
+	<-job.Job.Done()
+
+	// refetch the job to get latest results (EndedAt and Running will have changed)
+	s.jobsMu.Lock()
+	job, ok = s.jobs[id]
+	if !ok {
+		return nil, xerrors.Errorf("wait job ID: %d not found", id)
+	}
+	s.jobsMu.Unlock()
```
iand, line 311: Same here, make sure jobsMu is unlocked before return

iand (changes requested):

schedule/scheduler.go
frrist: whoops, done.

schedule/scheduler.go
frrist: done.

iand (approved):

PR #789, opened 2021-12-01
chore: consolidate logger names

iand:
This consolidates logger names so we have fewer to manage. Since Lotus is very chatty we often set all logging to error level and selectively lower the level to info or debug for loggers of interest (for example: `--log-level=error --log-level-named=lily/chain:debug,lily/storage:debug,lily/schedule:debug,lens/util:debug,chain:info,badgerbs:info`). There were several loggers that logged only in one or two places but had their own obscure composite name (e.g. `lily/lens/lily`)

I've followed the general principle of naming loggers as `lily/` followed by the top level package name. This hopefully will be somewhat robust against refactoring inner packages over time.

The main exception is `lily/actors` which is found in the `chain/actors` package, but I think we ought to move that to its own top level package. I'm leaving that as a separate exercise.

Fixed a couple of logger names that were not prefixed with `lily`

We only have these loggers now:

- lily/main
- lily/chain
- lily/actors
- lily/commands
- lily/config
- lily/lens
- lily/network
- lily/storage
- lily/tasks

### DIFF
--- a/chain/actors/builtin/init/diff.go
+++ b/chain/actors/builtin/init/diff.go
@@ -15,7 +15,7 @@ import (
 	typegen "github.com/whyrusleeping/cbor-gen"
 )
 
-var log = logging.Logger("lily/actor/init")
+var log = logging.Logger("lily/actors")
 
 type AddressMapChanges struct {
 	Added    []AddressPair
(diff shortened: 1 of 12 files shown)


iand:
This change may affect deployment configurations

placer14:
Do you know if lily will complain about setting log levels for loggers which don't exist?

placer14 (approved):

iand:
> Do you know if lily will complain about setting log levels for loggers which don't exist?

Yes it will fail to start

frrist (approved):

iand:
Test failure looks unrelated.

PR #790, opened 2021-12-01
fix(tests): Ensembleminimal requires mocker proofs

frrist:
Really weird and concerning this ever passed in CI but not locally.

This appears to be related to M1 macbooks and proof parameters


placer14 (approved):
This appears to work on my hardware.

PR #794, opened 2021-12-03
improve tracing data and include lotus tracing 

frrist:
Allows lily to send traces produced by lotus and improves the quality of lily's tracing data. Below is an example of a walk trace. The trace include the spans implemented in lily as well as the spans from lotus (e.g. `vm.Invoke`)
Screen Shot 2021-12-03 at 10 35 37 AM


frrist:
Just saw https://github.com/filecoin-project/lotus/pull/7646 is landing soon, this will be greater for further tracing through the stores.

chain/indexer.go
```diff
@@ -158,10 +160,13 @@ func NewTipSetIndexer(node lens.API, d model.Storage, window time.Duration, name
 
 // TipSet is called when a new tipset has been discovered
 func (t *TipSetIndexer) TipSet(ctx context.Context, ts *types.TipSet) error {
-	ctx, span := otel.Tracer("").Start(ctx, "Indexer.TipSet")
-	if span.IsRecording() {
-		span.SetAttributes(attribute.String("tipset", ts.String()), attribute.Int64("height", int64(ts.Height())))
-	}
+	ctx, span := otel.Tracer("").Start(ctx, "TipSetIndexer.TipSet", trace.WithAttributes(
```
iand, line 163: Can we optimize this to avoid allocations for the attributes? `ts.String()` allocates for example. The original code used `span.IsRecording` to do this. We should aim to keep tracing impact very low when it is disabled.

chain/indexer.go
```diff
@@ -626,6 +646,9 @@ func (t *TipSetIndexer) stateChangedActors(ctx context.Context, old, new cid.Cid
 }
 
 func (t *TipSetIndexer) runMessageProcessor(ctx context.Context, p MessageProcessor, name string, ts, pts *types.TipSet, emsgs []*lens.ExecutedMessage, blkMsgs []*lens.BlockMessages, results chan *TaskResult) {
+	ctx, span := otel.Tracer("").Start(ctx, fmt.Sprintf("TipSetIndexer.%s", name))
```
iand, line 649: I think it could be worth making this name more specific such as `"TipSetIndexer.MessageProcessor.%s"` (and the same for each of the run processor functions below)

main.go
```diff
@@ -91,39 +91,24 @@ func main() {
 				Destination: &commands.VisorLogFlags.LogLevelNamed,
 			},
 			&cli.BoolFlag{
-				Name:        "tracing",
-				EnvVars:     []string{"LILY_TRACING"},
+				Name:        "jaeger-tracing",
```
iand, line 94: Add way to set these options using environment variables

iand (approved):
Looks ok with some tweaks.

chain/indexer.go
frrist: done.

main.go
```diff
@@ -91,39 +91,28 @@ func main() {
 				Destination: &commands.VisorLogFlags.LogLevelNamed,
 			},
 			&cli.BoolFlag{
-				Name:        "tracing",
```
frrist, line 94: @placer14 heads up, this changes the flags for setting up tracing a little bit. Details of the change are in this file.

PR #811, opened 2022-01-13
Generic in memory lily testing via vectors

frrist:
### What
Revive vector tests - this PR adds the logic for testing lily via test vectors!
#### Add new vectors:
Add a vector to `build/test-vectors/vectors.json`, see `build/test-vectors/README.md` for instruction on how to do this

#### Execute vectors
Right now we only have vectors from the calibration network. Mainnet vectors can be added later with no modification to the test suite. 
When executing `make itest-calibration` for the follow occurs:
1. Vectors are downloaded (based on`vector.json`) to `/var/tmp/lily-test-vectors/<network_name>`. If the vector files are cached their contents is simply validated.
2. Ensure the expected vectors are present on the filesystem, failing the test suite otherwise.
3. Connect to timescale instance, migrating to the latest version if required.
4. Construct a full in-memory lily node, configured to talk to storage from the previous step
5. Execute a walk over each vector, the range of the walk is configured via the `to` and `from` fields in `vectors.json`.
6. Extract state from the vector using lotus api methods.
7. Compare data returned by lotus API with contents of storage, passing if equivalent, failing otherwise.

#### How do I run this?
`make dockerup && make itest-calibnet`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,6 +156,44 @@ jobs:
           file: /tmp/artifacts/coverage.out
       - store_artifacts:
           path: lily
+  integration-test:
+    resource_class: large
+    docker:
+      - image: cimg/go:1.16.6
+      - image: timescale/timescaledb:2.5.0-pg13
+        environment:
+          POSTGRES_PASSWORD: password
+    steps:
+      - checkout
+      - prepare
+      - run: # dep for DB wait script
+          name: install dockerize
+          command: wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && sudo tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
+          environment:
+            DOCKERIZE_VERSION: v0.3.0
+      - run: sudo apt-get update
+      - run: make deps
+      - run: make build
+      - run:
+          name: waiting for db
+          command: dockerize -wait tcp://localhost:5432 -timeout 1m
+      - run:
+          name: installing schema
+          command: ./lily migrate --latest
+          environment:
+            LILY_DB: postgres://postgres:password@localhost:5432/postgres?sslmode=disable
+      - restore_cache:
+          name: restore test vectors cache
+          key: lily-test-vectors-{{ checksum "./build/test-vectors/vectors.json" }}
+          paths:
+            - /var/tmp/lily-test-vectors/
+      - run:
+          command: make itest-calibnet
+      - save_cache:
+          name: save test vectors cache
+          key: lily-test-vectors-{{ checksum "./build/test-vectors/vectors.json" }}
+          paths:
+            - /var/tmp/lily-test-vectors/
   lint: &lint
     description: |
       Run golangci-lint.
@@ -252,6 +290,7 @@ workflows:
       - mod-tidy-check
       - lint-all
       - test
+      - integration-test
       - test-docker-mainnet
       - test-docker-mainnet-dev
       - test-docker-calibnet
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,7 +188,7 @@ jobs:
           paths:
             - /var/tmp/lily-test-vectors/
       - run:
-          command: make itest-calibnet
+          command: make itest
       - save_cache:
           name: save test vectors cache
           key: lily-test-vectors-{{ checksum "./build/test-vectors/vectors.json" }}
--- a/build/test-vectors/README.md
+++ b/build/test-vectors/README.md
@@ -1,13 +1,18 @@
-# Test Vectors
-This directory contains the test vector manuscript.
-Add new test vectors by appending them to `vectors.json`.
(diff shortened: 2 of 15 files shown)


.circleci/config.yml
```diff
@@ -156,6 +156,44 @@ jobs:
           file: /tmp/artifacts/coverage.out
       - store_artifacts:
           path: lily
+  integration-test:
```
frrist, line 159: new integration test job in CI.

build/test-vectors/README.md
```diff
@@ -0,0 +1,13 @@
+# Test Vectors
```
frrist, line 1: How to add new vectors

build/test-vectors/vectors.json
```diff
@@ -0,0 +1,16 @@
+{
```
frrist, line 1: the test vectors, all pinned on estuary.tech

commands/util/import.go
```diff
@@ -21,6 +22,72 @@ import (
 	"gopkg.in/cheggaaa/pb.v1"
 )
 
+// used for test vectors only
+func ImportFromFsFile(ctx context.Context, r repo.Repo, fs fs.File, snapshot bool) (err error) {
```
frrist, line 26: Helper to import a chain from an open file, used in vector testing.

docker-compose.yml
```diff
@@ -3,7 +3,7 @@ version: '3.5'
 services:
   timescaledb:
     container_name: timescaledb
-    image: timescale/timescaledb:2.4.1-pg13
+    image: timescale/timescaledb:2.5.0-pg13
```
frrist, line 6: We forgot to update this last time we updated CI, resulting in different behavior for tests. This change fixes the behavior differences.

itests/builder.go
```diff
@@ -0,0 +1,440 @@
+package itests
```
frrist, line 1: lily node builder that constructs a node for walking a test vector.

itests/fetch/fetch.go
```diff
@@ -0,0 +1,248 @@
...
+
+	"go.uber.org/multierr"
+	"golang.org/x/xerrors"
+	"gopkg.in/cheggaaa/pb.v1"
+
+	fslock "github.com/ipfs/go-fs-lock"
+	logging "github.com/ipfs/go-log/v2"
+)
+
+// Ported from github.com/filecoin-project/go-paramfetch
```
frrist, line 25: Mostly copied from here with a few modifications.

itests/init.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+	// this check ensures all the vectors in fetch.VectorFile were downloaded and will be run in the testing suite.
+	if len(MainnetTestVectors) != mainVectorCount {
+		log.Fatal("Failed to download all expected test vectors for mainnet")
+	}
+	if len(CalibnetTestVectors) != calibVectorCount {
+		log.Fatal("Failed to download all expected test vectors for calibnet")
+	}
+}
```
frrist, line 69: Ensure the vectors are downloaded and validated (vectors match digest) before running the tests, fail tests if vectors don't exist.

itests/vector_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/filecoin-project/lily/chain"
+	tstorage "github.com/filecoin-project/lily/storage/testing"
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/stretchr/testify/require"
+	"path/filepath"
+	"testing"
+	"time"
+)
+
+func TestCalibrationVector(t *testing.T) {
```
frrist, line 17: Where the tests run.

iand (approved):
General approach looks good. I noted some unnecessary complexity.

Missing validators for actor states?

Is the database truncated between runs. On CI I assume we get a fresh database but locally we may not.

I would double all timeouts at least if you want reliable runs on CI. Also to speed up you could implement a waiting pattern using the Until function in lily's wait package. Basically poll with an interval tens of milliseconds until a context deadline is reached.

frrist:
> Missing validators for actor states?

They are [here](https://github.com/filecoin-project/lily/pull/811/files#diff-9348bbe37364660bacb0e01ac0ef67b30e59d797a6ddbc50470365c8c9e4d903R111-R161
)

> Is the database truncated between runs. On CI I assume we get a fresh database but locally we may not.

Yes the `VectorWalkValidatorBuilder` does so in its [`Build()`](https://github.com/filecoin-project/lily/pull/811/files#diff-8503ab45683ce201bcaa420033c83db672927c0b114b8c20bb4daee9a58c827fR120-R128) method; [a new one is built](https://github.com/filecoin-project/lily/pull/811/files#diff-7eff7ef017c0412b112860e68cae3fbf59c83b7bd87a15cdf5d92b043773f876R36) for each test vector, meaning the database is fresh for each run.

> I would double all timeouts at least if you want reliable runs on CI. 

👍 will do before merging.

> Also to speed up you could implement a waiting pattern using the Until function in lily's wait package. Basically poll with an interval tens of milliseconds until a context deadline is reached.

Would this be different than the waiting for job completion that is performed [here](https://github.com/filecoin-project/lily/pull/811/files#diff-8503ab45683ce201bcaa420033c83db672927c0b114b8c20bb4daee9a58c827fR174)?

frrist:
Looks like the docker failures may be related to a CI outage: https://status.circleci.com/

Issue #819, opened 2022-01-25
Message Parse Param Changes


frrist:
### Description
With https://github.com/filecoin-project/lily/pull/799 landed the values of the parsed_messages params row have changed slightly. Here are few examples of the changes:
| Method Name | Previous | Current |
|---------------|----------|---------|
| Propose | {"To": "f1gojwyi2jcvptacdaopoh6as2whvwv2i67j2gqbq", "Value": "3149869747391647641", "Method": 0, "Params": ""} | {"To": "f1gojwyi2jcvptacdaopoh6as2whvwv2i67j2gqbq", "Value": "3149869747391647641", "Method": 0, "Params": null} |
| TerminatedSectors | {"Terminations": [{"Sectors": {"rle": "[14184,1]", "_type": "bitfield", "elemcount": "1"}, "Deadline": 6, "Partition": 0}]} | {"Terminations": [{"Sectors": [14184, 1], "Deadline": 6, "Partition": 0}]} |
| ProveCommitSectors | {"Proof": "<omitted_for_clarity>", "SectorNumber": 287854} | {"Proof": "<omitted_for_clarity>", "SectorNumber": 287854}|
[...]

PR #821, opened 2022-01-25
implement wrapper type for custom message param json marshalling

frrist:


lens/util/repo.go
```diff
@@ -340,3 +346,118 @@ func MakeGetActorCodeFunc(ctx context.Context, store adt.Store, next, current *t
...
 	}, nil
 }
+
+// wrapper type for overloading json marshal methods
+type paramWraperType struct {
+	obj     interface{}
+	replace map[reflect.Type]func(interface{}) interface{}
+}
+
+func (wt *paramWraperType) MarshalJSON() ([]byte, error) {
```
iand, line 356: It would be a lot clearer if you made this a first class function:

```Go
type marshaler func(interface{}) ([]byte, error)

func MarshalWithOverrides(v interface{}, overrides map[reflect.Type]marshaler) []byte, error {
  ...
}
```
The function then uses paramWraperType as an internal implementation detail.

The marshaler returns an error so you can replace the panics with error returns in your bitfield marshaler. 

Then you can call it like this:

```Go
var bitfieldCountMarshaler = func(v interface{}) ([]byte, error) {
   ...
}

b, err := MarshalWithOverrides(b, map[reflect.Type]marshaler{
   reflect.TypeOf(bitfield.BitField{}): bitfieldCountMarshaler,
})


```

iand (commented):
While I think this would work it's feels like it's more complex than it needs to be.

How about this alternate approach. Assuming that all message params are simple structs then you could use reflection to convert the param struct to an equivalent map (like is done on line 380 in this PR). If the value of a param field is a bitfield then you wrap it in a struct with its own custom marshaler. Then you just marshal the map. 

This approach relies on message params being simple structs without nested maps of other structs, which is valid as far as I can tell.

iand (approved):

lens/util/repo.go
```diff
@@ -340,3 +344,125 @@ func MakeGetActorCodeFunc(ctx context.Context, store adt.Store, next, current *t
...
+	}
+	count, err := rle.Count()
+	if err != nil {
+		return nil, err
+	}
+
+	var ret = struct {
+		Count uint64
+		RLE   []uint64
+	}{}
```
placer14, line 444: Given:

> [Old](https://github.com/filecoin-project/lily/issues/819#issue-1113354802):
`{"Recoveries": [{"Sectors": {"rle": "[10816,31,1407,1,2,23,1,4,1,1,1405,1,3,26,1,1,1,1,521,1,3,915,58,1,2,1343]", "_type": "bitfield", "elemcount": "2349"}, "Deadline": 6, "Partition": 0}]}`

I would have made this struct match the existing schema:

```go
var ret = struct {
  Count uint64 `json:elemcount`
  RLE []uint64 `json:rle`
  Type string `json:_type`
}{ Type: "bitfield" }
```

placer14, line 444: Given:

> [Old](https://github.com/filecoin-project/lily/issues/819#issue-1113354802):
`{"Recoveries": [{"Sectors": {"rle": "[10816,31,1407,1,2,23,1,4,1,1,1405,1,3,26,1,1,1,1,521,1,3,915,58,1,2,1343]", "_type": "bitfield", "elemcount": "2349"}, "Deadline": 6, "Partition": 0}]}`

I would have made this struct match the existing schema:

```go
var ret = struct {
  Count uint64 `json:elemcount`
  RLE []uint64 `json:rle`
  Type string `json:_type`
}{ Type: "bitfield" }
```

PR #828, opened 2022-01-31
fix(docker): Make dirty docker builds clean

placer14:
Lily has items committed to the repo which are also listed in `.dockerignore`. When docker sends the context during build, it is missing files which git state interprets as "dirty".

Removing `.dockerignore` fixes this problem. There are also some `.gitignore` cleanup.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,0 @@
-/.circleci
-/.github
-/build/*
-/extern/filecoin-ffi
-/Dockerfile*
-/docker-compose*
-/sentinel-visor
--- a/.gitignore
+++ b/.gitignore
@@ -12,9 +12,7 @@
 *.out
 
 # Dependency directories (remove the comment below to include it)
-# vendor/
 .idea
-visor
 !model/visor
 /*lily
 


iand:
Wouldn't it be better to align dockerignore and gitignore so nothing commited is excluded from the docker image? We don't need to include stray binaries and other artifacts in the build image. Even though they don't affect the final image copying them slows down the build,

placer14:
> be better to align dockerignore and gitignore so nothing commited is excluded from the docker image

Not quite sure what you mean. Majority of the items in dockerignore are committed into the repo. Any lines on dockerignore which remove items committed within the repo will cause the gitversion to appear as dirty. Removing the dockerignore will achieve exactly what you describe (nothing committed in the repo is excluded from the docker image).

The other option we could do here is to generate the git version outside of the docker build process while the repo is still "clean". Perhaps this would be better than conflating our build payload and our build versioning together?

iand:
> > be better to align dockerignore and gitignore so nothing commited is excluded from the docker image
> 
> Not quite sure what you mean. Majority of the items in dockerignore are committed into the repo. Any lines on dockerignore which remove items committed within the repo will cause the gitversion to appear as dirty. Removing the dockerignore will achieve exactly what you describe (nothing committed in the repo is excluded from the docker image).
> 
> The other option we could do here is to generate the git version outside of the docker build process while the repo is still "clean". Perhaps this would be better than conflating our build payload and our build versioning together?

I think your following comment makes this redundant but just in case: I mean we should keep dockerignore and ensure that it doesn't ignore anything that is committed (by aligning it with similar rules to those in gitignore)

placer14:
My point is that we've already committed items that are also enumerated in gitignore. For example, `./build` contains supporting files for the build process which deserve to be committed (albeit, not necessarily there).

frrist (approved):
looks good, please merge at will once @iand's comments are addressed.

placer14:
Superseded by https://github.com/filecoin-project/lily/pull/835. Closing.

PR #833, opened 2022-02-02
chore: add InternalParsedMessage to schema compatibility checklist

iand:


storage/sql.go
```diff
@@ -57,6 +57,7 @@ var models = []interface{}{
 	(*messages.MessageGasEconomy)(nil),
 	(*messages.ParsedMessage)(nil),
 	(*messages.InternalMessage)(nil),
+	(*messages.InternalParsedMessage)(nil),
```
frrist, line 60: Thank you. I wonder if there is a less error-prone way to build this list.

frrist (approved):

PR #835, opened 2022-02-03
fix(docker): Docker image build versions are no longer "dirty"

placer14:
Building the lily binary within the docker context makes the build
version always "dirty". This is because dockerignore omits files
from the build context which are committed into the git repo. When
these files are missing, git considers that repo "dirty".

This change allows the Makefile to use LILY_VERSION to set the build
version of the lily binary. This value used as an envvar within the
docker build environment which overrides the default ("dirty")
LILY_VERSION with the version perceived by git before dockerignore
filters/removes files from the build.


iand (approved):

frrist:
Ship it!

frrist (approved):

PR #848, opened 2022-02-09
fix(docker): Update docker base image for prod; Fix libc mismatch

### LINKED ISSUE #847: docker-calibnet builds result in lily binary which cannot run
### Describe the bug:

When docker images are built, the resulting lily binary within cannot run.
```
root@ed43517b0505:/# lily
lily: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by lily)
```

### Steps to Reproduce:

1. Run `make clean docker-calibnet`
2. Run `docker run -it --rm --entrypoint bash filecoin/lily:latest`
3. You are now within the docker container's shell, run `lily` and observe the above error.

placer14:
Fixes #847

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ commands:
     steps:
       - run:
           name: Run lily within the docker image
-          command: docker run --rm filecoin/lily:latest
+          command: docker run --rm filecoin/lily:latest --version
   publish-docker-from-branch-production:
     steps:
       - run:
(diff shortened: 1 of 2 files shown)


placer14:
I added a test here to ensure docker images are executable. Once they are confirmed to properly fail, I'll apply the fix commit and confirm they turn green.

placer14:
💥

placer14:
💥

frrist (commented):
It's not clear to me how this fixes and libc mismatch, is it related to `bookworm-curl`?

Also, go ahead and target `master` with this fix and I'll rebaase the rc branch onto it once it's in.

placer14:
My best guess is that the go1.17 environment uses a newer libc than what `buildpack-deps:buster-curl` has be default. `:bookworm-curl` has a later version of libc to match what go1.17 provides.

PR #849, opened 2022-02-09
feat: implement taskAPI

frrist:
### What
This PR implements the `taskAPI` which is passed to and used by the indexer and its tasks.
- The `taskAPI` presents a smaller interface for easier testing of tasks, it's simpler to mock than than the entire LilyAPI.
- The `StateChangedActors` method has been removed from the indexer and now lives in the `taskAPI`. I think this is a lot cleaner.
- The `GetExecutedAndBlockMessagesForTipset` and `GetMessageExecutionsForTipSet` method have been memoized, this allows them to be called within the Processors instead of called in the indexer with results passed to the processors. The goal here is to allow Processors to call these methods separatly to increase readability and extensabiliy of the code without sacrificing performance.

### Where is this going?
Part 1: https://github.com/filecoin-project/lily/pull/850
Part 2: https://github.com/filecoin-project/lily/pull/851
Part 3: https://github.com/filecoin-project/lily/pull/854


chain/indexer.go
```diff
@@ -261,14 +256,7 @@ func (t *TipSetIndexer) TipSet(ctx context.Context, ts *types.TipSet) error {
...
 					changesStart := time.Now()
-					var err error
-					var changes map[string]lens.ActorStateChange
-					// special case, we want to extract all actor states from the genesis block.
-					if current.Height() == 0 {
-						changes, err = t.getGenesisActors(ctx)
-					} else {
-						changes, err = t.stateChangedActors(tctx, current.ParentState(), next.ParentState())
-					}
+					changes, err := t.node.StateChangedActors(ctx, t.node.Store(), current, next)
```
placer14, line 260: Is the genesis case being handled somehow that I'm not seeing?

tasks/actorstate/actorstate_test.go
```diff
@@ -2,6 +2,7 @@ package actorstate_test
 
 import (
 	"context"
+	"github.com/filecoin-project/lily/chain/actors/adt"
```
placer14: nit: I prefer to see aliases on packages which are imported with a name other than their path.

```
	adt2 "github.com/filecoin-project/lily/chain/actors/adt"
```

placer14 (approved):
This looks good so far...

iand (approved):
LGTM once other reviewer comments resolved.

Some cosmetic/stylistic remarks: 

I don't like creating another new package with a single file, especially since it's under the lens package (which is just calling out to be renamed something relevant). It explodes the number of random packages to keep track of and loses some cohesion.

I think the main interface has the wrong name. `TaskAPI` sounds like the API that a task should implement but it's actually an API used by tasks to extract/lookup chain data. The word API is redundant too since every interface is an API. This interface represents some kind of source of chain activity data.

I think it would be cleaner to move the interface closer to the things that require it. I would move it to the `tasks` package and rename it to something like `DataSource`. Although I'm not 100% sold on that name, it does allow us to refer to it as `tasks.DataSource` which is more descriptive of its purpose.

Then I suggest also moving the implementation of the interface closer to where it is actually passed to the tasks. Currently that is in the `chain` package. Remove the `Impl` suffix which is a 20th Century Java-ism :wink:

frrist (commented):
> I don't like creating another new package with a single file, especially since it's under the lens package (which is just calling out to be renamed something relevant). It explodes the number of random packages to keep track of and loses some cohesion.

Agreed, my thinking here was getting the implementation right first, then following up with package name refactors, making review easier.


> I think it would be cleaner to move the interface closer to the things that require it. I would move it to the tasks package and rename it to something like DataSource

👍 I'm sold on the name for now, renaming later is easy.

> Then I suggest also moving the implementation of the interface closer to where it is actually passed to the tasks

iirc I didn't do that because of some circular dep error. I'll take a second look and land if its not too much 🔴 🟢

tasks/actorstate/actorstate_test.go
frrist: There are two adt packages here, one from specs-actors which is required for initializing adt strcutures, and a lily specific adt which is required for satisfying the MockAPI interface. I will rename the specs-actors adt package to `sa_adt`

PR #850, opened 2022-02-09
setup for indexer refacor

frrist:
### What
This PR targets the `indexer-refactor` branch.
This PR aims to keep indexer behavior consistent with the current while landing some refactor work to improve code extensibility.
- embed `GetExecutedAndBlockMessagesForTipset` in processors, preventing unnecessary (and expensive!) calls to the method when its output isn't used. And example of this would be running all actor tasks excluding the miner actor. Prior changes [here](https://github.com/filecoin-project/lily/pull/849/files#diff-88c7f968b3518b2c701d948daf6e103fdc0b57c3525a8e24f19645ae5c1d74c6R96-R115) ensure multiple calls to this method are only executed once, returning a cached value for subsequent calls. This change also makes the indexer's logic more concise.
- embed `GetMessageExecutions` in processors, taking advantage of the memoized method pattern used in previous commit. This has the added bonus of removing the `ProcessMessageExecution` interface and fitting the messege execution task under the `ProcessMessages` interface.

Follow on: https://github.com/filecoin-project/lily/pull/871

### DIFF
--- a/tasks/messageexecutions/message.go
+++ b/tasks/messageexecutions/message.go
@@ -3,31 +3,30 @@ package messageexecutions
 import (
 	"context"
 
-	"github.com/filecoin-project/lily/chain/actors/adt"
-	"github.com/filecoin-project/lily/lens"
-	"github.com/filecoin-project/lily/lens/util"
-	"github.com/filecoin-project/lily/model"
-	messagemodel "github.com/filecoin-project/lily/model/messages"
-	visormodel "github.com/filecoin-project/lily/model/visor"
-	"github.com/filecoin-project/lily/tasks/messages"
 	"github.com/filecoin-project/lotus/chain/types"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/lily/lens/util"
+	"github.com/filecoin-project/lily/model"
+	messagemodel "github.com/filecoin-project/lily/model/messages"
+	visormodel "github.com/filecoin-project/lily/model/visor"
+	"github.com/filecoin-project/lily/tasks"
+	"github.com/filecoin-project/lily/tasks/messages"
 )
 
-func NewTask() *Task {
-	return &Task{}
-}
-
 type Task struct {
+	node tasks.DataSource
 }
 
-func (p *Task) Close() error {
-	return nil
+func NewTask(node tasks.DataSource) *Task {
+	return &Task{
+		node: node,
+	}
 }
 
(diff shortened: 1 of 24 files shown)


tasks/messageexecutions/message.go
```diff
@@ -39,6 +38,12 @@ func (p *Task) ProcessMessageExecutions(ctx context.Context, store adt.Store, ts
 		StateRoot: pts.ParentState().String(),
 	}
 
+	mex, err := p.node.GetMessageExecutionsForTipSet(ctx, ts, pts)
+	if err != nil {
+		report.ErrorsDetected = xerrors.Errorf("getting messages executions for tipset: %w", err)
+		return nil, report, nil
+	}
+
```
frrist, line 46: Returning the error here would be fatal meaning the corresponding walk, watch, or fill job would abort completely -- that would be inconsistent with current behavior. So, we list this as a detected error in the report which will mark the task as `ERROR` in the corresponding processing report, and return immediately with no data extracted.

tasks/messages/message.go
```diff
@@ -41,6 +47,14 @@ func (p *Task) ProcessMessages(ctx context.Context, ts *types.TipSet, pts *types
 		StateRoot: pts.ParentState().String(),
 	}
 
+	tsMsgs, err := p.node.GetExecutedAndBlockMessagesForTipset(ctx, ts, pts)
+	if err != nil {
+		report.ErrorsDetected = xerrors.Errorf("getting executed and block messages: %w", err)
+		return nil, report, nil
+	}
```
frrist, line 54: (ditto) Returning the error here would be fatal meaning the corresponding walk, watch, or fill job would abort completely -- that would be inconsistent with current behavior. So, we list this as a detected error in the report which will mark the task as `ERROR` in the corresponding processing report, and return immediately with no data extracted.

tasks/msapprovals/msapprovals.go
```diff
@@ -47,6 +46,13 @@ func (p *Task) ProcessMessages(ctx context.Context, ts *types.TipSet, pts *types
 		StateRoot: pts.ParentState().String(),
 	}
 
+	tsMsgs, err := p.node.GetExecutedAndBlockMessagesForTipset(ctx, ts, pts)
+	if err != nil {
+		report.ErrorsDetected = xerrors.Errorf("getting executed and block messages: %w", err)
+		return nil, report, nil
+	}
+	emsgs := tsMsgs.Executed
+
```
frrist, line 55: (ditto) Returning the error here would be fatal meaning the corresponding walk, watch, or fill job would abort completely -- that would be inconsistent with current behavior. So, we list this as a detected error in the report which will mark the task as `ERROR` in the corresponding processing report, and return immediately with no data extracted.

iand (approved):
LGTM. One step closer to simplifying the indexer

PR #855, opened 2022-02-11
fix: error marshalling in processing reports

### LINKED ISSUE #597: Visor Processing Reports Empty Errors Detected field for message and msapproval tasks
### Describe the bug:
The Errors Detected filed for message and msapproval tasks is "empty" (`{}`) when their status is `ERROR`:
```
912740,bafy2bzacebppacou5wgmp3simdxoth6yuackewmz3sseeon3sryin25v6gk3a,watch_1625688475,msapprovals,2021-07-07 21:00:00.008374,2021-07-07 21:00:00.008470,ERROR,,{}
912740,bafy2bzacebppacou5wgmp3simdxoth6yuackewmz3sseeon3sryin25v6gk3a,watch_1625688475,messages,2021-07-07 21:00:00.008374,2021-07-07 21:00:00.008469,ERROR,,{}
```
I believe the cause could be related to marshaling an array of errors to json when writing to the table.

### Steps to Reproduce:
Run visor in daemon mode against mainnet

frrist:
- fixes #597

### DIFF
--- a/chain/indexer.go
+++ b/chain/indexer.go
@@ -10,6 +10,16 @@ import (
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-hamt-ipld/v3"
+	"github.com/filecoin-project/lotus/chain/state"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log/v2"
+	"go.opencensus.io/stats"
+	"go.opencensus.io/tag"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"golang.org/x/xerrors"
+
 	"github.com/filecoin-project/lily/chain/actors/adt"
 	init_ "github.com/filecoin-project/lily/chain/actors/builtin/init"
 	"github.com/filecoin-project/lily/chain/actors/builtin/market"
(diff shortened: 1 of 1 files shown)


chain/indexer.go
```diff
@@ -29,15 +39,6 @@ import (
...
 	"github.com/filecoin-project/lily/tasks/msapprovals"
-	"github.com/filecoin-project/lotus/chain/state"
-	"github.com/filecoin-project/lotus/chain/types"
-	"github.com/ipfs/go-cid"
-	logging "github.com/ipfs/go-log/v2"
-	"go.opencensus.io/stats"
-	"go.opencensus.io/tag"
-	"go.opentelemetry.io/otel"
-	"go.opentelemetry.io/otel/attribute"
-	"golang.org/x/xerrors"
```
frrist, line 40: Nothing changing here, just updated my editor.

chain/indexer.go
```diff
@@ -434,7 +435,15 @@ func (t *TipSetIndexer) TipSet(ctx context.Context, ts *types.TipSet) error {
...
-			if res.Report[idx].ErrorsDetected != nil {
+			if err := res.Report[idx].ErrorsDetected; err != nil {
+				// because error is just an interface it may hold a value of any concrete type that implements it, and if
+				// said type has unexported fields json marshaling will fail when persisting.
+				e, ok := err.(error)
+				if ok {
+					res.Report[idx].ErrorsDetected = &struct {
+						Error string
+					}{Error: e.Error()}
+				}
```
placer14, line 446: Is it worth putting a placeholder for Error if it can't be extracted? Something like "invalid/empty error detail" or similar?

placer14 (approved):
GTM

iand (approved):

chain/indexer.go
frrist: Humm we could probably do that, but it would mean enforcing everything that writes to the struct filed being an error since it can contain types that do not implement the error interface:
https://github.com/filecoin-project/lily/blob/fe9a838028cb447ae54e8af0b5349d57c7912427/tasks/actorstate/task.go#L164-L170

I'm not sure what the best way to enforce that would be, I'm not sure if the `ErrorsDetected` filed can be type `error` due to go-pg. What do you think?

frrist: I'm going to get this merged, but feel free to continue here if anything comes to mind.

PR #868, opened 2022-02-16
feat(ci): Continuous deployment for tags and deploy branches

placer14:
This PR enables continuous delivery triggers into our infrastructure repos. There are a few changes included:
- separate docker image publishing into parallel jobs
- fix calibnet and butterflynet docker builds
- docker image builds trigger on
  - tags containing `dev`
  - branch names containing `deploy`
- include job which triggers the pipeline in the infra repo. Deployments trigger on
  - tags containing `dev`
  - branch names containing `deploy`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,50 +56,56 @@ commands:
       - run:
           name: Run lily within the docker image
           command: docker run --rm filecoin/lily:latest --version
-  publish-docker-from-branch-production:
+
+jobs:
+  publish-production-image-from-branch-for-mainnet:
+    executor: dockerizer
     steps:
+      - checkout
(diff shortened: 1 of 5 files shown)


frrist (approved):
Awesome stuff!
Approving on the assumption this has been manually verified to work. 

>docker image builds trigger on

Is this a change from the original behavior? Prior to this change didn't we always build docker containers?

placer14:
> > docker image builds trigger on
> Is this a change from the original behavior? Prior to this change didn't we always build docker containers?

The prior behavior has not changed. This behavior I outlined is new and is required to support the special branch/tag deployment cases (which also need docker images to deploy other than the usual builds).

Issue #867, opened 2022-02-15
Re-architect Lily for horizontal: diff extraction, diff storage, diff processing.


frrist:
Why: we need to be able to scale processing horizontally or we will end up being unable to process things.

This epic is meant to explore a new approach for processing and extracting data from the Filecoin chain. Currently, A Lily daemon syncs the chain, diffs the states, and extracts from the diffs.

We would rather split the functionality in several components:

* A diffing component's only job is to sync the chain and produce state-diffs between consecutive epochs. These diffs are agnostic to any processing and they are to be stored in a permanent medium. We should be able to have a diff for every two consecutives epochs since genesis.

* The diffs are made available to a second component, an processing component, that is able to fech diffs over the network, and extract chain information just like Lily does. The diffs are supposed to be relatively small, immutable and can be made available to many readers at the same time.

By separating processing from diffing, we can now fully parallelize processing.

The diffing component can be optimized separately to be as fast as possible, does not have to do any processing/extraction, thus is not affected by any problems from that side. Diffs can potentially be generated per tasks or actor types and diffing work can happen in parallel.

This requires no small amount of effort and planning:

* Determining what is the best solution for permanent storage of diffs
* Running diffing components in production and determining what the scaling strategy for them will be.
* Running processing components that are reliably notified about the existence of new diffs and are able to perform processing of them quickly enough.
* Ensuring that any errors in the process are tracked, and that data gaps can be identified and addressed.

As  a bonus side, reprocessing the whole chain when diffs are already ready, should be something that can happen very quickly.

Issue #864, opened 2022-02-15
Lily's watch should tolerate re-orgs


frrist:
- A re-org does not cause errors and is considered business as usual during walks.
- Processing on the forked branch continues until the re-org happens.
- Processing on the heaviest branch after re-org continues from the place where the fork branched off (no missing data for heaviest branch).

hsanjuan:
Blocked waiting indexer refactor, horizontal scalability.

frrist:
Closed via: https://github.com/filecoin-project/lily/pull/869. The watcher is now stateless and tolerant of reorgs.

PR #869, opened 2022-02-16
Indexer Refactor feature branch

frrist:
See me run here: https://protocollabs.grafana.net/goto/dfhCT9Pnk?orgId=1

# Related Issues
- https://github.com/filecoin-project/lily/issues/867, while not fully addressed, this PR modifies lily so it can address this issue with follow on work
- https://github.com/filecoin-project/lily/issues/864, This PR together with completion of #867 will fully address this. The key piece of work done here was making the indexer stateless. State-full indexing caused errors when reorgs occur, now that indexing is stateless reorgs can be processed without error. A single instance of lily is not performant enough to process all tasks during a reog, but if ran with a lighter task load, this implementation of lily addresses this.
 - https://github.com/filecoin-project/lily/issues/861 the work done in #871 closes this.

# Description
This PR is a feature branch as was compiled from the below PR:

## https://github.com/filecoin-project/lily/pull/849
This PR implements the `taskAPI` which is passed to and used by the indexer and its tasks.
- The `taskAPI` presents a smaller interface for easier testing of tasks, it's simpler to mock than than the entire LilyAPI.
- The `StateChangedActors` method has been removed from the indexer and now lives in the `taskAPI`. I think this is a lot cleaner.
- The `GetExecutedAndBlockMessagesForTipset` and `GetMessageExecutionsForTipSet` method have been memoized, this allows them to be called within the Processors instead of called in the indexer with results passed to the processors. The goal here is to allow Processors to call these methods separatly to increase readability and extensabiliy of the code without sacrificing performance.
## https://github.com/filecoin-project/lily/pull/850
This PR aims to keep indexer behavior consistent with the current while landing some refactor work to improve code extensibility.
- embed `GetExecutedAndBlockMessagesForTipset` in processors, preventing unnecessary (and expensive!) calls to the method when its output isn't used. And example of this would be running all actor tasks excluding the miner actor. Prior changes [here](https://github.com/filecoin-project/lily/pull/849/files#diff-88c7f968b3518b2c701d948daf6e103fdc0b57c3525a8e24f19645ae5c1d74c6R96-R115) ensure multiple calls to this method are only executed once, returning a cached value for subsequent calls. This change also makes the indexer's logic more concise.
- embed `GetMessageExecutions` in processors, taking advantage of the memoized method pattern used in previous commit. This has the added bonus of removing the `ProcessMessageExecution` interface and fitting the messege execution task under the `ProcessMessages` interface.

## https://github.com/filecoin-project/lily/pull/871
This PR performs a (large) refactor of the indexer used by lily to index and process that chain. I have done my best to make each commit reviewable on its own and encourage reviewers to go through them sequentially. Here are some high-level details of what's changed:

#### Miner diffing
The DiffSectors and DiffPreCommits methods moved to the DataSource API. The API memoizes calls to these methods, allowing separate miner extractors to call them as needed.

#### Actor task breakup
All actor extractors have are separated such that each extractor produces a single model. For example, there is no longer a single miner state extractor rather there are several: `DeadlineInfoExtractor`, `InfoExtractor`, `FeeDebtExtractor`, `LockedFundsExtractor`, `PoStExtractor`, `PreCommitInfoExtractor`, `SectorInfoExtractor`, `V7SectorInfoExtractor`, etc. Each extract a single model from a miner state, utilizing the memoized methods (mentioned above) when needed. Similarly true for the market, init, power, reward, verifiedreg, etc. actors.

#### Message task breakup
The Messages processors are separate in a similar way - there exists a processor for `GasOutputs`, `DerivedGasOutputs`, `Messages`, `ParsedMessages`, `InternalMessages`, `InternalParsedMessages`, `Receipts`, `BlockMessages`, etc. These processors, too, use the memoized methods available on the DataSource API when appropriate.

#### The `Indexer` package
Here is where the new chain indexing logic lives. Its made up of 3 parts (all of which have code comments that should be referred to for more detail):
- `StateProcessor`: accepts a set of processor interfaces and executes them all in parallel, returning their execution results via a channel.
- `TipSetIndexer`: accepts a list of tasks, instantiates the appropriate processors, and passes them to the `StateProcessor`. The `TipSetIndexer` consumes results from the `StateProcessor` and produces results containing task names, extracted models, and reports, returning them on a channel. If the `TipSetIndexer` is instructed to halt before receiving all expected results from the `StateProcessors` it pushes the appropriate `SKIP` reports into its output channel as well.
- `IndexManager`: consumes results from the `TipSetIndexer` and persists them to storage. The `IndexManager` will signal if any reports it receives from the `TipSetIndexer` have the status `SKIP` or `ERROR`. The `IndexManager` may also be given a worker pool to run multiple `TipSetIndexer` in parallel.

#### Simiplifcation of Processors
There are now only 3 processors used to extract state from the chain:
- `TipSetProcessor`: processes a singel tipset
- `TipSetsProcessor`: processes a current and an executed tipset
- `ActorProcessor`: processes a current and executed tipset and actors that changed during the executed tipset appliction.
Said processors are consumed and executed by the `StateProcessor`

#### `ReportProcessor`
This processor only returns a `visormodel.ProcessingReport` and writes null rounds to the `visormodel.ProcessingReport` table. It replaces logic that once lived in the consensus task.

#### Models as tasks
Lily now accepts a list of model names to extract rather than a list of tasks corresponding to sets of models. For example, the task `actorstatesminer` has been replaced with: `miner_current_deadline_infos`, `miner_fee_debts`, `miner_infos`, `miner_locked_funds`,`miner_pre_commit_infos`,`miner_sector_deals`, `miner_sector_events`, `miner_sector_infos`, `miner_sector_posts`
similar for all other `actorstates*` tasks. This gives more flexibility to the user for extracting data. And, together with the "Actor task breakup" change, allows for certaine extractors to be configured to run over specific actor versions (greatly simplifying the extraction of the MinerSectorInfos after the v7 actor upgrade)

# TODO
- [ ] Implement `--task-exclude` flag mentioned here: https://github.com/filecoin-project/lily/pull/871#issuecomment-1056908941
- [x] Address https://github.com/filecoin-project/lily/issues/897
- [ ]  document new APIs (index method) and task arguments


chain/watcher.go
```diff
@@ -51,22 +60,42 @@ type Watcher struct {
...
-		c.pool.Stop()
 		// ensure we clear the fatal error after shut down, this allows the watcher to be restarted without reinitializing its state.
 		c.setFatalError(nil)
+		// ensure we shut down the pool when the watcher stops.
+		c.pool.Stop()
+		// ensure we reset the tipset cache to avoid process stale state if watcher is restarted.
+		c.cache.Reset()
+		// unregister the observer
+		// TODO https://github.com/filecoin-project/lotus/pull/8441
+		//c.api.Unregister(notifier)
```
placer14, line 88: This PR is merged now. Does this need to change?

chain/watcher.go
```diff
@@ -208,3 +236,131 @@ const (
...
+
+	log.Debugw("head notifier revert", "to", to.Key().String(), "from", from.Key().String())
+	select {
+	case ev <- &HeadEvent{
+		Type:   HeadEventRevert,
+		TipSet: from,
+	}:
+	default:
+		log.Errorw("head notifier event channel blocked dropping revert event", "to", to.Key().String(), "from", from.Key().String())
+	}
```
placer14, line 364: I see both Apply and Revert drop the event if the buffer is full (different behavior from before the change, too). How will this affect the work needed on those tipsets? If so, what is the intended recovery mechanism?

placer14: If it's helpful, consider adding this information to the newly-exposed arg.

placer14:
Some thoughts about task names. I don't mind a full set of task options but I would consider creating a set of helper aliases. As Ian mentioned, `all` could be good for explicitness (I would expect default of `--tasks` to be all anyway). I would also include alias for the all of the prior versions tasks mapped onto their appropriate tasks. Ex: `actorstatesminer` would enable all models which were processed under that task previously. We can remove them later in a major version release.

placer14 (approved):
Sticky approval. (Sorry for slow review turnaround on this.) Congrats on getting this shipped! 💪 🎉

chain/watcher.go
frrist: Not yet, it will change when we upgrade to the lotus release with the change, so 1.15.2 or something. (Currently, this exits in lotus master, but we don't use master as our dep)

chain/watcher.go
frrist: We can either drop the event or [block the observer module](https://github.com/filecoin-project/lotus/blob/20bf46f3098b3c433231480dda98cbf66940a201/chain/events/observer.go#L196-L242): we don't want to block by waiting to send on the event channel since it will prevent other registered observers from receiving apply and revert events (method hangs), which is all-around bad.

Blocking _should_ never happen during normal operations since the watcher will queue work it can't currently perform: https://github.com/filecoin-project/lily/blob/indexer-refactor/chain/watcher.go#L176. But if a watch job blows up/fails/gets restarted, its observer will still be registered and receiving events https://github.com/filecoin-project/lily/blob/indexer-refactor/chain/watcher.go#L69. And may continue to receive events even after its unregistered: https://github.com/filecoin-project/lotus/blob/20bf46f3098b3c433231480dda98cbf66940a201/chain/events/observer.go#L259. So this change prevents "zombie observers" from blocking on Apply/Revert until they can be unregistered.

placer14: If I'm hearing correctly that this will prevent that work from being completed, it would probably be helpful to indicate a brief warning in that error that that param could be adjusted. And a good idea to include in the help text as well.

PR #871, opened 2022-02-16
Full indexer refactor

frrist:
### Wat
This PR performs a (large) refactor of the indexer used by lily to index and process that chain. I have done my best to make each commit reviewable on its own and encourage reviewers to go through them sequentially. Here are some high-level details of what's changed:

#### Miner diffing
The DiffSectors and DiffPreCommits methods moved to the DataSource API. The API memoizes calls to these methods, allowing separate miner extractors to call them as needed.

#### Actor task breakup
All actor extractors have are separated such that each extractor produces a single model. For example, there is no longer a single miner state extractor rather there are several: `DeadlineInfoExtractor`, `InfoExtractor`, `FeeDebtExtractor`, `LockedFundsExtractor`, `PoStExtractor`, `PreCommitInfoExtractor`, `SectorInfoExtractor`, `V7SectorInfoExtractor`, etc. Each extract a single model from a miner state, utilizing the memoized methods (mentioned above) when needed. Similarly true for the market, init, power, reward, verifiedreg, etc. actors.

#### Message task breakup
The Messages processors are separate in a similar way - there exists a processor for `GasOutputs`, `DerivedGasOutputs`, `Messages`, `ParsedMessages`, `InternalMessages`, `InternalParsedMessages`, `Receipts`, `BlockMessages`, etc. These processors, too, use the memoized methods available on the DataSource API when appropriate.

#### The `Indexer` package
Here is where the new chain indexing logic lives. Its made up of 3 parts (all of which have code comments that should be referred to for more detail):
- `StateProcessor`: accepts a set of processor interfaces and executes them all in parallel, returning their execution results via a channel.
- `TipSetIndexer`: accepts a list of tasks, instantiates the appropriate processors, and passes them to the `StateProcessor`. The `TipSetIndexer` consumes results from the `StateProcessor` and produces results containing task names, extracted models, and reports, returning them on a channel. If the `TipSetIndexer` is instructed to halt before receiving all expected results from the `StateProcessors` it pushes the appropriate `SKIP` reports into its output channel as well.
- `IndexManager`: consumes results from the `TipSetIndexer` and persists them to storage. The `IndexManager` will signal if any reports it receives from the `TipSetIndexer` have the status `SKIP` or `ERROR`. The `IndexManager` may also be given a worker pool to run multiple `TipSetIndexer` in parallel.

#### Simiplifcation of Processors
There are now only 3 processors used to extract state from the chain:
- `TipSetProcessor`: processes a singel tipset
- `TipSetsProcessor`: processes a current and an executed tipset
- `ActorProcessor`: processes a current and executed tipset and actors that changed during the executed tipset appliction.
Said processors are consumed and executed by the `StateProcessor`

#### `ReportProcessor`
This processor only returns a `visormodel.ProcessingReport` and writes null rounds to the `visormodel.ProcessingReport` table. It replaces logic that once lived in the consensus task.

#### Models as tasks
Lily now accepts a list of model names to extract rather than a list of tasks corresponding to sets of models. For example, the task `actorstatesminer` has been replaced with: `miner_current_deadline_infos`, `miner_fee_debts`, `miner_infos`, `miner_locked_funds`,`miner_pre_commit_infos`,`miner_sector_deals`, `miner_sector_events`, `miner_sector_infos`, `miner_sector_posts`
similar for all other `actorstates*` tasks. This gives more flexibility to the user for extracting data. And, together with the "Actor task breakup" change, allows for certaine extractors to be configured to run over specific actor versions (greatly simplifying the extraction of the MinerSectorInfos after the v7 actor upgrade)


chain/indexer/manager.go
```diff
@@ -0,0 +1,204 @@
...
+
+	stats.Record(ctx, metrics.IndexManagerActiveWorkers.M(i.active))
+	stats.Record(ctx, metrics.IndexManagerWaitingWorkers.M(int64(i.pool.WaitingQueueSize())))
+	if i.pool.WaitingQueueSize() > i.pool.Size() {
+		log.Warnw("queuing worker in Manager pool", "waiting", i.pool.WaitingQueueSize())
+	}
+	log.Infow("submitting tipset for async indexing", "height", ts.Height(), "active", i.active)
+
+	ctx, span := otel.Tracer("").Start(ctx, "Manager.TipSetAsync")
+	i.pool.Submit(func() {
```
frrist, line 116: This pool used to live in the [watcher](https://github.com/filecoin-project/lily/blob/fe9a838028cb447ae54e8af0b5349d57c7912427/chain/watcher.go#L134), its now managed here instead.

chain/indexer/manager.go
```diff
@@ -0,0 +1,204 @@
...
+				} else {
+					log.Infow("task success", "task", res.Name, "status", report.Status, "duration", report.CompletedAt.Sub(report.StartedAt))
+				}
+			}
+			m := &ModelResult{
+				Name:  res.Name,
+				Model: model.PersistableList{res.Report, res.Data},
+			}
+
+			// synchronously export extracted data and its report.
```
frrist, line 198: persisting `synchronously` here should be fine since the `taskResults` channel is buffered. Alternatively this could be performed with an errorGroup that we wait on for completion. I think there is a lot of value in this method `TipSet` returning once it's processed and persisted all results, otherwise, it's callers must go inspect the database to determine success.

chain/indexer/processor/state.go
```diff
@@ -0,0 +1,326 @@
...
+	ctx, span := otel.Tracer("").Start(ctx, "StateProcessor.State")
+
+	num := len(sp.tipsetProcessors) + len(sp.actorProcessors) + len(sp.tipsetsProcessors) + len(sp.builtinProcessors)
+	results := make(chan *Result, num)
+	taskNames := make([]string, 0, num)
+
+	taskNames = append(taskNames, sp.startReport(ctx, current, results)...)
+	taskNames = append(taskNames, sp.startTipSet(ctx, current, results)...)
+	taskNames = append(taskNames, sp.startTipSets(ctx, current, executed, results)...)
+	taskNames = append(taskNames, sp.startActor(ctx, current, executed, results)...)
```
frrist, line 90: I'd like a generic method that can be used to start any type of processor, but I couldn't figure out how to do that here. Ideas?

chain/indexer/tipset.go
```diff
@@ -0,0 +1,408 @@
...
+
+	tipsetProcessors := map[string]processor.TipSetProcessor{}
+	tipsetsProcessors := map[string]processor.TipSetsProcessor{}
+	actorProcessors := map[string]processor.ActorProcessor{}
+	reportProcessors := map[string]processor.ReportProcessor{
+		"builtin": indexer.NewTask(node),
+	}
+
+	for _, t := range tasks {
+		switch t {
```
frrist, line 91: Seeking ideas on how to make this switch less verbose.

chain/walker_test.go
```diff
@@ -57,22 +57,18 @@ func TestWalker(t *testing.T) {
...
+	require.NoError(t, err, "NewManager")
 	t.Logf("initializing indexer")
-	idx := NewWalker(tsIndexer, nodeAPI, 0, int64(beforeHead.Height()))
+	idx := NewWalker(im, nodeAPI, 0, int64(head.Height()), false)
 
 	t.Logf("indexing chain")
 	err = idx.WalkChain(ctx, nodeAPI, head)
 	require.NoError(t, err, "WalkChain")
 
-	// TODO NewTipSetIndexer runs its processors in their own go routines (started when TipSet() is called)
```
frrist, line 71: `WalkChain returns onces it's processed and persisted eveything, allow this to be removed.

chain/watcher.go
```diff
@@ -77,7 +74,7 @@ func (c *Watcher) index(ctx context.Context, he *HeadEvent) error {
 
 		// If we have a zero confidence window then we need to notify every tipset we see
 		if c.confidence == 0 {
-			if err := c.obs.TipSet(ctx, he.TipSet); err != nil {
+			if err := c.maybeIndexTipSet(ctx, he.TipSet); err != nil {
```
frrist, line 77: We call `maybeIndexTipSet` here to ensure tipsets are processed asynchronously while watching with no confidence. Previously this call would block which is undesired while watching the chain.

chain/watcher_test.go
```diff
@@ -181,7 +187,7 @@ func collectBlockHeaders(n lens.API, ts *types.TipSet) (blockHeaderList, error)
 	blocks := ts.Blocks()
 
 	for _, bh := range ts.Blocks() {
-		if bh.Height == 0 {
+		if bh.Height < 2 {
```
frrist, line 190: This change ensures the list of blocks returned by `collectBlockHeaders` never returns the genesis block which we don't process.

commands/index.go
```diff
@@ -0,0 +1,202 @@
...
+		_, err = api.LilyIndex(ctx, cfg)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	},
+}
+
+var IndexCmd = &cli.Command{
```
frrist, line 138: New command can be called like:
```
lily index --storage=Database1 --name=testing --window=5m --tasks="blocks" height 1578553
```
or
```
lily index --storage=Database1 --name=testing --window=5m --tasks=blocks tipset bafy2bzacecaoicpyxikq3yvabri5yvscc42hvvikrtdk73kiavog45wi4gify
```

I find this easier than calling walk with the same from and too heights.

itests/vector_test.go
```diff
@@ -30,8 +32,8 @@ func TestCalibrationVector(t *testing.T) {
 		t.Run(filepath.Base(vf.File.Name()), func(t *testing.T) {
 			tvb := NewVectorWalkValidatorBuilder(vf).
 				WithDatabase(strg).
-				WithRange(vf.From, vf.To-1). // TODO file bug
```
frrist, line 33: we can now walk the entire to and from range with no weird off-by-one errors, yay.

tasks/actorstate/actorstate.go
```diff
@@ -45,32 +61,12 @@ type ActorStateAPI interface {
...
 }
-
-// All supported actor state extractors
-var (
-	extractorsMu sync.Mutex
-	extractors   = map[cid.Cid]ActorStateExtractor{}
-)
-
-// Register adds an actor state extractor
-func Register(code cid.Cid, e ActorStateExtractor) {
```
frrist, line 62: this registration now happens in the TipSetIndexer.

tasks/actorstate/actorstate.go
```diff
@@ -20,13 +21,28 @@ import (
...
-	TipSet          *types.TipSet
-	ParentTipSet    *types.TipSet
+	Actor      types.Actor
+	ChangeType tasks.ChangeType
+	Address    address.Address
+	Current    *types.TipSet
+	Executed   *types.TipSet
+}
+
+func (a ActorInfo) Attributes() []attribute.KeyValue {
```
frrist, line 31: used when tracing to add attributes to the span.

tasks/actorstate/actorstate.go
```diff
@@ -20,13 +21,28 @@ import (
...
+	return []attribute.KeyValue{
+		attribute.String("address", a.Address.String()),
+		attribute.String("code", a.Actor.Code.String()),
+		attribute.String("head", a.Actor.Head.String()),
+		attribute.String("type", builtin.ActorNameByCode(a.Actor.Code)),
+		attribute.String("change", a.ChangeType.String()),
+	}
+}
+
+func (a ActorInfo) MarshalLogObject(enc zapcore.ObjectEncoder) error {
```
frrist, line 41: implement the zap inline logging interface:
```
log.Debugw("extract", zap.String("extractor", "InitExtractor"), zap.Inline(ActorInfo))
```

tasks/actorstate/miner/sectorv7.go
```diff
@@ -0,0 +1,108 @@
...
+	"go.opentelemetry.io/otel"
+	"go.uber.org/zap"
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/lily/model"
+	minermodel "github.com/filecoin-project/lily/model/actors/miner"
+	"github.com/filecoin-project/lily/tasks/actorstate"
+)
+
+type V7SectorInfoExtractor struct{}
```
frrist, line 15: `minermodel.MinerSectorInfoV7` gets its own extractor that we register with the appropriate actor code (v7 actors)

iand:
A question about the new way of specifying tasks to be run. Do we expect the user to be aware of table name changes. For example do they need to specify `miner_sector_infos` for pre 15 networks and `miner_sector_infos_v7` afterwards? Should they specify both? 

Since there are a lot of tasks, it would be useful to be able to specify an exclusion list so we could run all tasks except x and y. Maybe something like `lily walk --tasks= --exclude-tasks=miner_sector_events,miner_sector_posts`

chain/indexer/manager.go
```diff
@@ -0,0 +1,204 @@
...
+	api      tasks.DataSource
+	storage  model.Storage
+	indexer  Indexer
+	exporter Exporter
+	window   time.Duration
+
+	// used for async tipset indexing
+	pool   *workerpool.WorkerPool
+	fatal  error
+	active int64
```
iand, line 41:
```suggestion
	active int64 // must be accessed using atomic operations
```

chain/indexer/manager.go
```diff
@@ -0,0 +1,204 @@
...
+		defer func() {
+			atomic.AddInt64(&i.active, -1)
+			span.End()
+		}()
+
+		ts := ts
+		success, err := i.TipSet(ctx, ts)
+		if err != nil {
+			log.Errorw("index manager suffered fatal error", "error", err, "height", ts.Height(), "tipset", ts.Key().String())
+			i.fatal = err
```
iand, line 127: Take lock here

chain/indexer/manager.go
```diff
@@ -0,0 +1,204 @@
...
+		return false, err
+	}
+
+	// there are no results, bail.
+	if taskResults == nil {
+		return true, nil
+	}
+
+	success := true
+	ctx, exportSpan := otel.Tracer("").Start(ctx, "Manager.Export")
```
iand, line 170:
```suggestion
	ctx, exportSpan := otel.Tracer("").Start(ctx, "Manager.TipSet")
```

chain/indexer/manager.go
iand: Agree and empirically we know that persistence is only a small fraction of the time spent actually indexing

chain/indexer/processor/state.go
iand: Maybe you need to pass these functions a type with the following interface:

```Go
type TaskData interface {
    Current() *types.TipSet
    Executed() *types.TipSet
    Actors()  tasks.ActorStateChangeDiff
}
```
Each processor could then have the same signature:

```Go
Process(ctx context.Context, td TaskData) (model.Persistable, *visormodel.ProcessingReport, error)
```

chain/indexer/processor/state.go
```diff
@@ -0,0 +1,326 @@
...
+	"github.com/filecoin-project/lily/metrics"
+	"github.com/filecoin-project/lily/model"
+	visormodel "github.com/filecoin-project/lily/model/visor"
+	"github.com/filecoin-project/lily/tasks"
+)
+
+type TipSetProcessor interface {
+	// ProcessTipSet processes a tipset. If error is non-nil then the processor encountered a fatal error.
+	// Any data returned must be accompanied by a processing report.
+	// Implementation of this interface are responsible for abort when their context is canceled.
```
iand, line 23:
```suggestion
	// Implementations of this interface must abort processing when their context is canceled.
```

chain/indexer/processor/state.go
```diff
@@ -0,0 +1,326 @@
...
+
+		sp.pwg.Add(1)
+		go func() {
+			ctx, _ = tag.New(ctx, tag.Upsert(metrics.TaskType, name))
+			stats.Record(ctx, metrics.TipsetHeight.M(int64(current.Height())))
+			stop := metrics.Timer(ctx, metrics.ProcessingDuration)
+			defer stop()
+
+			pl := log.With("name", name, "height", current.Height())
+			pl.Infow("start processor")
```
iand, line 118:
```suggestion
			pl.Infow("processor started")
```

iand: For consistencey with other messages

lens/util/repo.go
```diff
@@ -41,44 +41,44 @@ var log = logging.Logger("lily/lens")
 // GetMessagesForTipset returns a list of messages sent as part of pts (parent) with receipts found in ts (child).
 // No attempt at deduplication of messages is made. A list of blocks with their corresponding messages is also returned - it contains all messages
 // in the block regardless if they were applied during the state change.
-func GetExecutedAndBlockMessagesForTipset(ctx context.Context, cs *store.ChainStore, sm *stmgr.StateManager, next, current *types.TipSet) (*lens.TipSetMessages, error) {
+func GetExecutedAndBlockMessagesForTipset(ctx context.Context, cs *store.ChainStore, sm *stmgr.StateManager, current, executed *types.TipSet) (*lens.TipSetMessages, error) {
 	ctx, span := otel.Tracer("").Start(ctx, "GetExecutedAndBlockMessagesForTipSet")
 	defer span.End()
-	if !types.CidArrsEqual(next.Parents().Cids(), current.Cids()) {
-		return nil, xerrors.Errorf("child tipset (%s) is not on the same chain as parent (%s)", next.Key(), current.Key())
+	if !types.CidArrsEqual(current.Parents().Cids(), executed.Cids()) {
```
iand, line 47: This is enforcing a previous/next relationship between the tipsets. I don't really find the name `executed` to be useful. Why can't it be `parent` since it must be the tipset whose blocks are in the parent collection of the current tipset. It helps with the mental model that these tipsets are closely related: one follows another.

iand: If not, the text in the next error needs adjusting

tasks/actorstate/miner/miner.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/filecoin-project/lily/model"
+)
+
+// was services/processor/tasks/miner/miner.go
+
+// StorageMinerExtractor extracts miner actor state
+type StorageMinerExtractor struct{}
+
+func (m StorageMinerExtractor) Extract(ctx context.Context, a actorstate.ActorInfo, node actorstate.ActorStateAPI) (model.Persistable, error) {
```
iand, line 22: Is this used? I thought each extractor was run indepedently.

iand (commented):
Not finished my review but submitting so you have sight of comments sooner

iand:
As an aside. Since this allows us to restrict which tables are being written to, we could limit the migration check for those tables in the database rather than expecting every table. That would allow lily to run against any database that contains the set of tables that it expects to write to.

chain/datasource/datasource.go
```diff
@@ -0,0 +1,392 @@
...
+	diffSectorsGroup singleflight.Group
+
+	diffPreCommitCache *lru.Cache
+	diffPreCommitGroup singleflight.Group
+}
+
+func NewDataSource(node lens.API) (*DataSource, error) {
+	t := &DataSource{
+		node:                node,
+		executedBlkMsgGroup: singleflight.Group{},
```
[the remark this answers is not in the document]
frrist: Yup, removing.

chain/datasource/datasource.go
```diff
@@ -0,0 +1,392 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	t.executedTsCache, err = lru.New(4)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO these cache sizes will need to increase depending on the number of miner actors at each epoch
```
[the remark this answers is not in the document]
frrist: Good point, adding metrics and env vars for configuring values

chain/datasource/miner.go
```diff
@@ -0,0 +1,83 @@
...
+
+	curA, err := t.Actor(ctx, addr, ts.Key())
+	if err != nil {
+		return nil, err
+	}
+	preA, err := t.Actor(ctx, addr, pts.Key())
+	if err != nil {
+		return nil, err
+	}
+	key := curA.Head.String() + preA.Head.String()
```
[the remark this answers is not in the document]
frrist: Yeah, good idea, adding that in.

chain/indexer/manager.go
```diff
@@ -0,0 +1,204 @@
...
+	for _, opt := range opts {
+		opt(im)
+	}
+	return im, nil
+}
+
+// TipSetAsync enqueues `ts` into the Manager's worker pool for processing. An error is returned if any Indexer in
+// the pool encounters a fatal error, else the method returns immediately.
+func (i *Manager) TipSetAsync(ctx context.Context, ts *types.TipSet) error {
+	if i.fatal != nil {
```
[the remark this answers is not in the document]
frrist: whoops, addressing.

chain/indexer/manager.go
frrist: actually just going to remove this, was residual from previous impl and the method already has a span.

chain/indexer/processor/state.go
frrist: oh I like this. Then the `Actors()` method implementation would memoize the result of ActorStateChanges and let all processors to call the method internally, similar to what we do now for ExectuedMessage and such. 

This will simplify the StateProcessor down to two processor interfaces:
```go
type Data interface {
	Current() *types.TipSet
	Executed() *types.TipSet
	Actors(ctx context.Context) (tasks.ActorStateChangeDiff, error)
}

type ModelProcessor interface {
	Process(ctx context.Context, d Data) (model.Persistable, *visormodel.ProcessingReport, error)
}

type ReportProcessor interface {
	Process(ctx context.Context, current *types.TipSet) (visormodel.ProcessingReportList, error)
}
```

Given the amount of code churn implementing this now will cause I'd prefer to do this as a follow on once we settle on design here, thoughts?

lens/util/repo.go
frrist: I felt `executed` was more descriptive as (I think) it aligns well with expected consensus: 
- it reminds me all messages it contains have been applied(executed) on the chain (to produce `current`)
- reminds me that `current` is an un-executed tipset. (almost call `current` `unexecuted` instead..)

Child and Parent just makes me think of their relation, one follows another, and doesn't describe the nature of EC as well.

tasks/actorstate/miner/miner.go
frrist: Its not, I'll delete it

chain/indexer/tipset.go
placer14: Not to throw interfaces at this again, but if all initialization could be done behind an interface w something like `Init(<all deps over all tasks>...)` they could also be done in a forloop. Each Init doesn't need to use all of the args.

placer14 (commented):
I could only get through half of it this afternoon. Will put more into this in the evening. Looks great so far!

chain/find.go
```diff
@@ -26,7 +28,7 @@ var FullTaskSet mapset.Set
 
 func init() {
 	FullTaskSet = mapset.NewSet()
-	for _, t := range AllTasks {
+	for _, t := range indexer.AllTasks {
```
placer14, line 31: Not obvious where `indexer` comes from and should be more obvious where that is. I'd also like assurance (verbal or explicitly in code) that this is the only mutation and won't change accidentally. (Concerned about future changes potentially triggering this footgun.)

chain/find.go
```diff
@@ -261,7 +264,7 @@ select height, task
 order by height desc
 ;
 `,
-		len(AllTasks), // arg 0
+		len(indexer.AllTasks),                      // arg 0
```
placer14, line 267: I don't believe this change is going to work in the following scenario:

User runs walks using tasks `messages`, `chaineconomics` and `blocks`. Let's say epochs 4 and 7 produce SKIPs for one of the tasks (say `blocks`) which need to be gapfilled. Then the user `gap finds` over that range for tasks `messages` and `blocks` and not `chaineconomics`. The query would consider the gaps from `blocks` (in epochs 4 and 7) filled because `len(indexer.AllTasks)` is 2 and is the number of records found for heights 4 and 7 is the same count (despite the gaps). (See diagram for vis)

```
 Task State
          ┌────┬────┬────┬────┬────┬────┬────┬────
          │ 2  │ 3  │ 4  │ 5  │ 6  │ 7  │ 8  │ 9
  Blocks  │ ok │ ok │ SK │ ok │ ok │ SK │ ok │ ok
  Messages│ ok │ ok │ ok │ ok │ ok │ ok │ ok │ ok
  Economics ok │ ok │ ok │ ok │ ok │ ok │ ok │ ok
          │    │    │    │    │    │    │    │
          └────┴────┴────┴────┴────┴────┴────┴────
```

This query will probably need to be explicit about which tasks need to match here. I can come back around to help here if needed.

commands/index.go
```diff
@@ -0,0 +1,202 @@
...
+		if indexFlags.name != "" {
+			indexName = indexFlags.name
+		}
+
+		var tsStr string
+		if tsStr = cctx.Args().First(); tsStr == "" {
+			return xerrors.Errorf("tipset argument required")
+		}
+
+		tsk, err := parseTipSetKey(strings.Split(tsStr, ","))
```
placer14, line 48: Should the split be inside `parseTipSetKey`?

commands/index.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil
+	},
+}
+
+var IndexHeightCmd = &cli.Command{
+	Name:  "height",
+	Usage: "Index the state of a tipset from the filecoin blockchain by height",
+	Action: func(cctx *cli.Context) error {
+		ctx := lotuscli.ReqContext(cctx)
+
```
placer14, line 86: Couldn't hurt to include an `Args()` count check here? (Unsure how missing args are handled when accessed.)

tasks/messageexecutions/internal_parsed_message/task.go
```diff
@@ -0,0 +1,90 @@
...
+	node tasks.DataSource
+}
+
+func NewTask(node tasks.DataSource) *Task {
+	return &Task{
+		node: node,
+	}
+}
+
+func (p *Task) ProcessTipSets(ctx context.Context, current *types.TipSet, executed *types.TipSet) (model.Persistable, *visormodel.ProcessingReport, error) {
```
placer14, line 29: Note to self: Didn't see this task added to the list of tasks.

tasks/messages/gas_economy/task.go
```diff
@@ -0,0 +1,109 @@
...
+	node tasks.DataSource
+}
+
+func NewTask(node tasks.DataSource) *Task {
+	return &Task{
+		node: node,
+	}
+}
+
+func (t *Task) ProcessTipSets(ctx context.Context, current *types.TipSet, executed *types.TipSet) (model.Persistable, *visormodel.ProcessingReport, error) {
```
placer14, line 33: Note to self: Didn't see this task added to the list of tasks.

tasks/actorstate/miner/sectorv7.go
```diff
@@ -35,6 +35,10 @@ func (V7SectorInfoExtractor) Extract(ctx context.Context, a actorstate.ActorInfo
 	// transform sector changes to a model
 	sectorModel := minermodel.MinerSectorInfoV7List{}
 	for _, added := range sectorChanges.Added {
+		sectorKeyCID := ""
+		if added.SectorKeyCID != nil {
+			sectorKeyCID = added.SectorKeyCID.String()
+		}
```
placer14, line 41: Nit: This changeset is not quite what the commit message says.

placer14 (commented):
More comments. Almost done.

chain/indexer/registry.go
```diff
@@ -0,0 +1,101 @@
+package indexer
+
+const (
+	// blocks
+	BlocksTask = "blocks"
+
+	// consensus
+	ChainConsensusTask = "chain_consensus"
```
placer14, line 8: Why do we use both `snake_case` and `nocase` for task name representations. I assume (hope) the snake_case variant is never exposed publicly?

placer14 (approved):
This is a great improvement. Review completed. Approving in advance w expectation of fixing the find bug.

tasks/actorstate/market/market.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// SanitizeLabel ensures that s is a valid utf8 string by replacing any ill formed bytes with a replacement character.
+func SanitizeLabel(s string) string {
+	if utf8.ValidString(s) {
+		return s
+	}
+
+	tr := runes.ReplaceIllFormed()
+	return tr.String(s)
```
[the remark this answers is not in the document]
frrist: This isn't code isn't modified in the PR, just moved. Will revisit this as a follow on (I suspect there isn't a lot to gain by removing the var)

chain/find.go
frrist: indexer is a package, `AllTasks` is a variable from the package. This is a mechanical change since the indexer was move into a new package.

chain/find.go
frrist: This is a mechanical change due to the creation of the indexer package. No logic has changed here.

commands/index.go
frrist: the check is below on line 93

chain/indexer/registry.go
frrist: We only use `snake_case`, these strings, eg. `chain_consensus` are provided by the used on the command line. Would you prefer `nocase`?

chain/watcher_test.go
frrist: Due to the below line (188) adding bh.Parents to the blocks list, parent of 1 is 0 and we don't want 0 (genesis) in the list since we can't process it.

commands/index.go
frrist: Yea good idea

iand (approved):
I'm done on the rebiew but would appreciate your thoughts on my comment about table names that change over time

chain/indexer/tipset.go
```diff
@@ -0,0 +1,408 @@
...
+			))
+		case MinerSectorEventsTask:
+			actorProcessors[t] = actorstate.NewTask(node, actorstate.NewTypedActorExtractorMap(
+				miner.AllCodes(), miner2.SectorEventsExtractor{},
+			))
+		case MinerSectorPoStTask:
+			actorProcessors[t] = actorstate.NewTask(node, actorstate.NewTypedActorExtractorMap(
+				miner.AllCodes(), miner2.PoStExtractor{},
+			))
+		case MinerSectorInfoTask:
```
frrist, line 127: @iand Here is where I've combined tables produced by the `miner_sector_info` task arg, passing this task name will produce models for both the `miner_sector_info` and `miner_sector_info_7v` table.

frrist:
> A question about the new way of specifying tasks to be run. Do we expect the user to be aware of table name changes. 

Right now no - I added a comment [here ](https://github.com/filecoin-project/lily/pull/871/files#r821225011)for context. This does hide a bit of complexity, unsure what the pattern should be for handling cases like this. What do you think?

chain/indexer/registry.go
placer14: The inconsistency in the interface from the prior version rubs me the wrong way (interface changes are always painful for the user). While there's nothing wrong with changing it, I would personally err on the side of leaving things as consistent as possible from one version to the next unless there's a good reason.

On the other hand, with the interface changing so drastically here I can be less bothered by the new arg pattern. And the division of existing tasks does make its sub-tables/task (which are now tasks themselves) somewhat awkward to name without changing all of them.

Arguing for/against doesn't leave me with strong feels one way or the other, so please decide what you think is best here and I'll support downstream. #bikeshedding 

If you leave it and in case you haven't decided/considered this, I would at least want next release that includes this refactor to have a minor version bump rather than a revision bump. 🙏

placer14:
I imagine whatever is going to consume these models and persist them to postgres to also provide a view over them for the user to interface with as well. As in, a single "table" to query which hides joining on multiple versions of schema for the same model from the user. (Assuming, also, that we think  this model/deserializing/persistence thing should also have the concern of creating opinionated views like this.)

Downside is that CSVs are unable to express these domain-specific schema changes. Something that we could probably only express in documentation to help the user consume this reasonably. This may not be a major issue as anyone handling CSVs probably has some amount of skill to handle this... Especially if the documentation includes some sample views that they can crib from.

chain/indexer/registry.go
frrist: The change here is that we no longer have tasks that produce many models, rather there is (mostly) a one-to-one mapping between the task name and the model it produces. There isn't really a new arg pattern here, rather there a different args altogether.

frrist:
@placer14 this PR doesn't change how models are persisted to storage, rather it gives users more control over what models are persisted.

frrist:
I've rebased this PR's target (`indexer-refactor`) onto master and this PR onto `indexer-refactor`. I'm going to merge this once CI gives me the green light. We can continue the conversation regarding implementation here: https://github.com/filecoin-project/lily/pull/869

PR #880, opened 2022-02-25
Test out parallel ADT diffing in Parallel

frrist:
cc @iand


iand:
Rebased this on master to try out

frrist:
revisiting once some deps land in AMT and HAMT packages

PR #884, opened 2022-02-26
fix: get missing tasks for each height.

kasteph:


chain/find.go
```diff
@@ -222,44 +222,42 @@ func (g *GapIndexer) findTaskEpochGaps(ctx context.Context) (visor.GapReportList
...
+	group by height, task, status
+	having count(*) != ?0
+),
+
+-- dense_ranked_tasks: where the actual "ranking" happens
+-- rank by status_rank
+-- rank 1 means the lowest value status rank found of the tuple (height, task)
+dense_ranked_tasks as (
+	select *,
+	dense_rank() over (partition by height, task order by status_rank) as ranked_status
```
kasteph, line 249: A sample of  how the dense ranking happens:

```
height,task,status,ranked_status
890000,actorstatesinit,OK,1
890000,actorstatesinit,SKIP,2
890000,actorstatesmarket,OK,1
890000,actorstatesmarket,SKIP,2
890000,actorstatesminer,ERROR,2
890000,actorstatesminer,OK,1
```

By ranking on `status_rank` with partitions over `(height, task)`, we are ensuring that we are selecting in the set the lowest ranked task and any tasks that have had an `OK` (or whatever given) status.

chain/find.go
```diff
@@ -222,44 +222,42 @@ func (g *GapIndexer) findTaskEpochGaps(ctx context.Context) (visor.GapReportList
...
-		where status = ?4
-		group by incomplete_task_counts.height, incomplete_task_counts.status
-		having count(incomplete_task_counts.height) != ?0
-
-	) incomplete_heights
+with
+
+-- ranked_tasks: filters by height and prepares ranking status of tasks by int
+-- only considers heights with incomplete task count
+ranked_tasks as (
```
placer14, line 229: I would like names to be descriptive of their domain-specific properties rather than functional properties. This will make it easier to reason about the purpose as well as the mechanics in complex queries like this instead of just the latter.

Ex: `ranked_tasks` -> `status_weighted_tasks_from_incomplete_epochs`

Aside: I have no problem with (and would prefer) verbosity in my names if they add value over terse names. Particularly if they are separated by many lines. But please use your judgement on names. #namingishard

chain/find.go
placer14: > we are ensuring that we are selecting in the set the _lowest_ ranked task

Where is this assertion happening? I only see a condition for `rt.ranked_status = 1` (which doesn't even need to be an int since matching on strings work just as well... and it's not clear if the transform+equivalency check on an int is going to be cheaper than just a text equivalency check on text.)

placer14 (commented):
Apologies, but I'm not clear on what the motivation for this change is or what this intends to fix/improve. Is there an issue or could we include a description in the PR? 

Most glaring here is the value of the dense rank isn't clear to me (yet). (Also, very sorry for the late review. 🙏)

frrist:
This PR came about from a one-off conversation between @kasteph and myself. There isn't an issue tracking this, as it's mostly explorative work. The tentative next steps are refactoring the gap-find query into a single SQL function. @kasteph I've converted this to a draft to signal this, hope that's alright with you.

kasteph:
@frrist sounds good, thanks!

chain/find.go
kasteph: @placer14 thanks! I'll try to be more descriptive  (and more correct) with names.

chain/find.go
kasteph: It's on this [line](https://github.com/filecoin-project/lily/pull/884/files#diff-d49953a87209cc88edee91a3c0173f069297d6c4835cd1e17340094e4fbfd94aR258), but upon taking a look at it again, you're right. A simple text check would work just fine. I initially worked on this query with the assumption that we would be finding tasks do not have a single `OK` on it, hence the ranking.

placer14: Yeah, I think using a weighted filter (via dense_rank) might be a little overkill here. If the absence of the 'OK' is the case you're attempting to solve for here, it would be more obvious as a LEFT JOIN selecting for NULL.

Here's my whack at this some time back which we decided we weren't ready for. Maybe there are some patterns there that might help you. (https://github.com/filecoin-project/lily/blob/48437e66bcf1c7b5f990de3f69d633e38b1353bf/chain/find.go#L88-L193)

schemas/v1/6_gap_find.go
```diff
@@ -0,0 +1,72 @@
+package v1
+
+// Schema version 1 adds gap find function
```
frrist, line 3: nit: version 1.6

schemas/v1/6_gap_find.go
```diff
@@ -0,0 +1,72 @@
...
+		AND v.task IN (SELECT * FROM interesting_tasks)
+		GROUP BY 1, 2, x.status
+	),
+
+	null_round AS (
+		SELECT pr.height, t.task
+		FROM heights_in_processing_report pr
+		CROSS JOIN interesting_tasks t
+		WHERE (
+			pr.status_information = null_status OR pr.status_information IS NOT null
```
frrist, line 56: This should only assert on `status_information` == `null_status` since `status_information` can have values in it not relating to a null round. This also needs a test case if you don't mind extending the new test suite I added.

schemas/v1/6_gap_find.go
```diff
@@ -0,0 +1,72 @@
+package v1
+
+// Schema version 1 adds gap find function
+
+func init() {
+	patches.Register(
+		6,
+		`
+CREATE FUNCTION {{ .SchemaName | default "public"}}.gap_find(
```
frrist, line 9: can postgres functions be commented? If so it would be helpful to add a short summary describing the arguments this method expects and the results it returns.

frrist (approved):
Goood stuff! Excited to have this. Please address comments before merging.

schemas/v1/6_gap_find.go
kasteph: Thanks @frrist!

If there is a task with a `status_information` of `NULL_ROUND` but a `status` of `INFO`, is that considered a complete task or a gap?

frrist: So far there is not - and this makes me wonder if it would be better to make `NULL_ROUND` an actual status (which we can't do since this needs to be backward compatible with previous logic).

I supposed the more correct assertion for determining if a row corresponds to a null round would be to assert that `status = INFO and status_information = NULL_ROUND` since those are the values written by the gap find and consensus tasks.

PR #885, opened 2022-03-02
fix: state cache falls through to slow path on error

iand:
Prevents cache failures causing extraction to fail


Dockerfile
```diff
@@ -1,54 +0,0 @@
-# This file was generated with `make docker-files` and should not
```
frrist, line 1: You're probably aware - undo this change.

Dockerfile.dev
```diff
@@ -1,42 +0,0 @@
-# This file was generated with `make docker-files` and should not
```
frrist, line 1: ditto

lens/util/store.go
```diff
@@ -307,6 +302,20 @@ func (cas *CachingStateStore) Get(ctx context.Context, c cid.Cid, out interface{
...
 }
 
+func (cas *CachingStateStore) tryAssign(value interface{}, out interface{}) error {
+	o := reflect.ValueOf(out).Elem()
+	if !o.CanSet() {
+		return xerrors.Errorf("out parameter cannot be set")
+	}
+
+	if !value.(reflect.Value).Type().AssignableTo(o.Type()) {
+		return xerrors.Errorf("out parameter cannot be assigned cached value")
```
frrist, line 312: include the types in the error message

lens/util/store.go
```diff
@@ -307,6 +302,20 @@ func (cas *CachingStateStore) Get(ctx context.Context, c cid.Cid, out interface{
 	return nil
 }
 
+func (cas *CachingStateStore) tryAssign(value interface{}, out interface{}) error {
+	o := reflect.ValueOf(out).Elem()
+	if !o.CanSet() {
+		return xerrors.Errorf("out parameter cannot be set")
```
frrist, line 308: include type in error message

frrist (approved):
Nice! This is ready to merge once comments are addressed.

Dockerfile
iand: Hmm. make clean deletes this file so what is the expected pattern for building? Do I need to regerenate the dockerfiles and recommit every time?

lens/util/store.go
iand: done

lens/util/store.go
iand: done

iand:
@frrist restored dockerfiles

frrist:
Looks good, my approval is sticky.

PR #907, opened 2022-03-09
fix(schema): Rollback v15 ProveCommitAggregate param schema to match 

### LINKED ISSUE #904: Aggregate Proof params schema changed in v0.8.6
### Describe the bug:

> Hi team, we’ve been counting sectors in aggregation type of messages, and we’ve noticed a recent parameter change in parsed_messages params (attached screenshots). Can we safely assume that elemcount and Count are the same parameters in two different naming fashions?

image
image

### Steps to Reproduce:

Walk epochs after v15 network upgrade (`1594679`) and observe `parsed_messages` for `method = 'ProveCommitAggregate'` and observe new schema. These messages before the afore mentioned epoch will show the prior schema.

### Lily Version: v0.8.6

[Reported in Filecoin slack by Edward Shi.](https://filecoinproject.slack.com/archives/C01V2SPTURL/p1646641678644719) Thanks Ed!

placer14:
fixes #904

### DIFF
--- a/lens/util/repo.go
+++ b/lens/util/repo.go
@@ -460,10 +460,15 @@ var bitfieldCountMarshaller = func(v interface{}) ([]byte, error) {
 		return nil, err
 	}
 
+	// this struct matches the param schema used in network v14
+	// see https://github.com/filecoin-project/lily/pull/821/files#r821851219
 	var ret = struct {
-		Count uint64
-		RLE   []uint64
-	}{}
+		Count uint64   `json:"elemcount"`
+		RLE   []uint64 `json:"rle"`
+		Type  string   `json:"_type"`
+	}{
+		Type: "bitfield",
+	}
 	if r.HasNext() {
 		first, err := r.NextRun()
 		if err != nil {
--- a/tasks/messages/message_test.go
+++ b/tasks/messages/message_test.go
@@ -203,7 +203,7 @@ func TestParseMessageParams(t *testing.T) {
 				ChainCommitEpoch: 1287345,
 			}),
 			wantMethod:  "SubmitWindowedPoSt",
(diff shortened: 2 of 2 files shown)


schemas/v1/6_restore_prove_aggregate_param_schema.go
```diff
@@ -0,0 +1,36 @@
...
+
+		-- and do not add if the key is missing
+		false
+
+	)
+	FROM {{ .SchemaName | default "public" }}.parsed_messages
+
+	-- network v15 network upgrade epoch
+	WHERE height > 1594679
+	AND method = 'ProveCommitAggregate';
```
placer14, line 34: This limits the update to only the messages since the upgrade epoch and for this specifically affect message type. The cost will grow the longer that v0.8.6-v0.8.7 is run before updating to v0.8.8.

schemas/v1/6_restore_prove_aggregate_param_schema.go
```diff
@@ -0,0 +1,36 @@
...
+
+		-- modify the value at key "SectorNumbers"
+		'{SectorNumbers}'::text[],
+
+		-- to use this interpolated json literal
+		concat(
+			'{"rle":', params->'SectorNumbers'->>'RLE',
+			',"elemcount":', params->'SectorNumbers'->>'Count',
+			',"_type":"bitfield"}'
+		)::jsonb,
```
placer14, line 24: This creates a JSON literal which is cast to a `jsonb`. It interpolates the existing values into the schema with the original keys.

schemas/v1/6_restore_prove_aggregate_param_schema.go
frrist: This is slick!
One concern I have: we will need to apply this migration to all messages with params containing bitfields, I think this include more than just the `ProveCommitAggregate` params. Perhaps there is a query that could find all these types?

placer14: Yes, we can definitely do a manual search to see which types need adjusting and write a migration for each. 👍

placer14:
After discussion aside w @frrist we aren't sure whether we want to include a migration with the fix as it would allow uncaught invariants to continue existing when we can be certain of accuracy by deleting and re-walking over the affected epochs (which would be communicated via docs/notices).

So for live node operators, the application of this fix looks something like:

1. Upgrade to lily v0.8.8
2. Manually execute SQL to remove incorrect data
3. Run `lily walk` over affected epoch ranges on affected tasks.

This is being parked until our in-sync discussion tomorrow. Comments welcome.

hsanjuan:
> This is being parked until our in-sync discussion tomorrow. Comments welcome.

I too think we can live without a migration and fix manually.

The most important thing are the dumps, which cannot be fixed by migration anyways.

iand:
Can you create an issue with the `backfill` label noting what needs to be backfilled, including height ranges

frrist (approved):
LGTM, please merge after a 🟢 CI

placer14:
We agreed to remove the migration so it could be applied as required by the use case along w documentation describing the bug (which will be limited to live data). [Our data archive will correct this bug](https://github.com/filecoin-project/lily/issues/911) with an appropriate announcement of it's completion.

PR #922, opened 2022-04-01
fix: prevent deadlock when persisting models with same height and different stateroots

frrist:


kasteph (approved):

chain/indexer/exporter.go
```diff
@@ -16,28 +20,50 @@ import (
...
+// ExportResult persists []ModelResult `results` to model.Storage `strg`. An error is returned if persisting the
+// model fails. This method will block if models at `height` are being persisted allowing the following constraints to be met:
+// - if data with height N and SR1 is being persisted and a request to persist data with the same values is made, allow it
+// - if data with height N and SR2 is being persisted and a request to persist data with height N and SR1 is made, block
+func (me *ModelExporter) ExportResult(ctx context.Context, strg model.Storage, height int64, results []*ModelResult) error {
+	// lock exporting based on height only allowing a single height to be persisted simultaneously
+	heightKey := strconv.FormatInt(height, 10)
+	me.heightKeyMu.LockKey(heightKey)
+	if err := me.heightKeyMu.UnlockKey(heightKey); err != nil {
+		//NB: this could be a panic or ignored since it would indicate some fundamentally impossible error, the lock will always exist given the prior lock call.
```
placer14, line 46: I'd prefer error to panic so as to not interrupt the daemon and other jobs.

placer14 (approved):

chain/indexer/exporter.go
frrist: Sounds good, will leave this method as is. The comment is meant to inform the reader that this error _should_ be impossible should they encounter it - and if they do something is very wrong with the implementation.

PR #927, opened 2022-04-06
Frrist/watcher state cleanup

frrist:
This change allows the watcher to be stopped and started without keeping stale state around: state includes workerpool, tipset cache, and notifier. This change will have a follow on once https://github.com/filecoin-project/lotus/pull/8441 lands in a release.

### DIFF
--- a/chain/watcher.go
+++ b/chain/watcher.go
@@ -61,21 +61,26 @@ func (c *Watcher) Run(ctx context.Context) error {
 	// init the done channel for each run since jobs may be started and stopped.
 	c.done = make(chan struct{})
 
+	// create a worker pool with workers to index tipsets as they become avaiable.
+	c.pool = workerpool.New(c.poolSize)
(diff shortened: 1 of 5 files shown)


chain/watcher.go
```diff
@@ -208,3 +236,131 @@ const (
...
+	}
+
+	log.Debugw("head notifier revert", "to", to.Key().String(), "from", from.Key().String())
+	select {
+	case ev <- &HeadEvent{
+		Type:   HeadEventRevert,
+		TipSet: from,
+	}:
+	default:
+		log.Errorw("head notifier event channel blocked dropping revert event", "to", to.Key().String(), "from", from.Key().String())
```
frrist, line 363: Most of this 🟢 was this code moving files, the default case to avoid blocking on the channel is new logic

iand (approved):
LGTM

PR #929, opened 2022-04-14
Lily distributed worker pattern

frrist:
Implementation of https://www.notion.so/pl-strflt/Distributed-Lily-Worker-Pattern-bbf101e392144e4184fae1ef7074ee02


lens/lily/impl.go
```diff
@@ -113,23 +163,44 @@ func (m *LilyNodeAPI) LilyWatch(_ context.Context, cfg *LilyWatchConfig) (*sched
...
+		Events:         m.Events,
+		ChainModuleAPI: m.ChainModuleAPI,
 	}
 
-	taskAPI, err := datasource.NewDataSource(m)
-	if err != nil {
-		return nil, err
-	}
+	var idx indexer.Indexer
+	if cfg.Queue == "" {
```
placer14, line 172: IMO, this condition is very mysterious at first glance and isn't clear why this should trigger the storage setup vs not. If possible, I would make it more clear through better variable/function names or comments.

Something like `cfg.IsNotifier() bool` or `configureForNotifier(*LilyWatchConfig) bool`? Or whatever functional behavior this is actually checking for. (ex: IsNotifier describes behavior vs IsIntegrated which describes the domain object. For the casual reader, the behavioral name is more valuable of the two.)

lens/lily/impl.go
```diff
@@ -161,21 +229,30 @@ func (m *LilyNodeAPI) LilyWalk(_ context.Context, cfg *LilyWalkConfig) (*schedul
 		JobName: cfg.Name,
 	}
 
-	// create a database connection for this watch, ensure its pingable, and run migrations if needed/configured to.
-	strg, err := m.StorageCatalog.Connect(ctx, cfg.Storage, md)
-	if err != nil {
-		return nil, err
-	}
+	var idx indexer.Indexer
+	if cfg.Queue == "" {
```
placer14, line 233: Same.

lens/lily/impl.go
placer14: Or BETTER, make this a concern inside of `NewWatcher`. I don't see any reason for setup to live in this abstraction when so many of these are args to NewWatcher anyway. WDYT?

testutil/db.go
```diff
@@ -90,3 +90,17 @@ func tryTestDatabaseLock(ctx context.Context, db *pg.DB) func(context.Context) (
...
+
+// TruncateBlockTables ensures the indexing tables are empty
+func TruncateBlockTables(tb testing.TB, db *pg.DB) error {
+	_, err := db.Exec(`TRUNCATE TABLE block_headers`)
+	require.NoError(tb, err, "block_headers")
+
+	_, err = db.Exec(`TRUNCATE TABLE block_parents`)
+	require.NoError(tb, err, "block_parents")
+
+	_, err = db.Exec(`TRUNCATE TABLE drand_block_entries`)
```
placer14, line 102: If the truncation is important to test success, we should require.NoError here too.

placer14 (approved):
This was a pleasure to read. Minor comments and suggestions. Sticky approval.

lens/lily/impl.go
frrist: Yeah great point, I wasn't sure how to wire this all up. 

How would you feel about new methods that are specific to the distributed indexer? Something like `WatchNotify`, `WalkNotify`, `FillNotify`.

testutil/db.go
frrist: I think we do on the line below?

lens/lily/impl.go
placer14: This seems like a good idea. 👍

chain/watch/observer.go
```diff
@@ -0,0 +1,155 @@
...
+	ev := h.eventsCh()
+	h.mu.Unlock()
+
+	// This is imprecise since it's inherently racy but good enough to emit
+	// a warning that the event may block the sender
+	if len(ev) == cap(ev) {
+		log.Warnw("head notifier buffer at capacity", "queued", len(ev))
+	}
+
+	log.Debugw("head notifier setting head", "tipset", ts.Key().String())
```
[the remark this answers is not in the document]
frrist: This was a mechanical change so I will leave the behavior as is. We can revisit in a follow on if this prove to be confusing.

chain/watch/watcher.go
```diff
@@ -0,0 +1,270 @@
...
+// NewWatcher creates a new Watcher. confidence sets the number of tipsets that will be held
+// in a cache awaiting possible reversion. Tipsets will be written to the database when they are evicted from
+// the cache due to incoming later tipsets.
+func NewWatcher(api WatcherAPI, indexer indexer.Indexer, name string, opts ...WatcherOpt) *Watcher {
+	w := &Watcher{
+		api:     api,
+		name:    name,
+		indexer: indexer,
+
+		bufferSize: WatcherDefaultBufferSize,
```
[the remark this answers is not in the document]
frrist: Going to leave this as is so users can be specific with the values they set. Scaling the value based on the worker and buffer-size flag would unexpected from a UX perspective.

commands/worker.go
```diff
@@ -0,0 +1,98 @@
...
+			return err
+		}
+
+		if err := printNewJob(os.Stdout, res); err != nil {
+			return err
+		}
+
+		return nil
+	},
+}
```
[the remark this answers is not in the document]
frrist: Generally I am not a fan of documentation existing in two places (lilium.sh and `help`), it adds overhead for the maintainer and introduces the possibility for conflicting instructions if both are not kept up to date. When we are more certain of this implementation we can add documentation in one place (lilium.sh I suspect)

PR #953, opened 2022-05-06
deps: update go-amt-ipld w/ parallel diffing

frrist:


chain/actors/adt/diff/array.go
```diff
@@ -2,14 +2,32 @@ package diff
...
 
+var AmtParallelWorkerLimit int64
+
+const AmtParallelWorkerEnv = "LILY_AMT_PARALLEL_WORKER_LIMIT"
+
+func init() {
+	AmtParallelWorkerLimit = int64(runtime.NumCPU())
+	workerStr := os.Getenv(AmtParallelWorkerEnv)
+	if workerStr != "" {
+		AmtParallelWorkerLimit, err := strconv.ParseInt(workerStr, 10, 64)
```
placer14, line 24: nit: Put this error check in the `if` above to flatten.

placer14 (approved):

chain/actors/adt/diff/array.go
placer14: Actually, that would ruin the `AmtParallelWorkerLimit` on error. nm. GTM

PR #975, opened 2022-06-03
chore: reject invalid tasks on client

### LINKED ISSUE #969: Starting a notifier with an invalid task name should fail on client

frrist:
closes #969

Ideally, this check would be performed once in a `job run` `Before` [option](https://github.com/filecoin-project/lily/blob/master/commands/job/job.go#L27), but that would break the `survey` command since it uses different task names. This is making me rethink if the `survey` job should go somewhere else. Thoughts?

### DIFF
--- a/commands/job/fill.go
+++ b/commands/job/fill.go
@@ -1,11 +1,13 @@
 package job
 
 import (
+	"fmt"
 	"os"
 
 	lotuscli "github.com/filecoin-project/lotus/cli"
 	"github.com/urfave/cli/v2"
 
+	"github.com/filecoin-project/lily/chain/indexer/tasktype"
 	"github.com/filecoin-project/lily/commands"
 	"github.com/filecoin-project/lily/lens/lily"
 )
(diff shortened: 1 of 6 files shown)


placer14:
What was the motivation for separating `peeragents` again? IIRC, it was primarily a behavorial difference in that it was indexing external/ephemeral data and not local data. Which means it probably doesn't need the majority of the machinery that all of the other commands use. So maybe there is a separate binary here which just consumes a node API? ...`lotus-survey` effectively becomes just `lotus` with some API<->DB interactions on top.

commands/job/fill.go
```diff
@@ -35,6 +37,14 @@ Constraints:
 		GapFillNotifyCmd,
 	},
 	Before: func(cctx *cli.Context) error {
+		tasks := RunFlags.Tasks.Value()
+		for _, taskName := range tasks {
+			if _, found := tasktype.TaskLookup[taskName]; found {
+			} else if _, found := tasktype.TableLookup[taskName]; found {
```
placer14, line 43: Why twice? No objection but it's unusual.

placer14 (approved):

frrist:
> What was the motivation for separating peeragents again? IIRC, it was primarily a behavorial difference in that it was indexing external/ephemeral data and not local data

Yeah, the `Survey`er collects information on peers lily is connected to, meaning it doesn't inspect the chain - unlike the rest of the tasks.

> Which means it probably doesn't need the majority of the machinery that all of the other commands use. So maybe there is a separate binary here which just consumes a node API? ...lotus-survey effectively becomes just lotus with some API<->DB interactions on top.


Yeah, it needs a DB and an API. It may be the case that drone already does some of this, do you know?

PR #1000, opened 2022-06-21
chore: walk test uses in memory storage

frrist:
- reduce the requirements for running test

I have noticed this test [flakes sometimes](https://app.circleci.com/pipelines/github/filecoin-project/lily/3536/workflows/6ef9122e-0372-4bd6-8041-0c6505b958d3/jobs/15593). On the off chance this is a race condition within Postgres I have removed it as a dependency of the test opting to use an in-memory storage type instead.

### DIFF
--- a/chain/walk/walker_test.go
+++ b/chain/walk/walker_test.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	itestkit "github.com/filecoin-project/lotus/itests/kit"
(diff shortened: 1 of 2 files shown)


kasteph:
@frrist shouldn't one test for Postgres race conditions anyway? This test might actually be surfacing such an issue (which I think you've also brought up before).

PR #1001, opened 2022-06-22
feat: implement market deal proposal label migration

### LINKED ISSUE #987: Actor version 8 DealProposal type change
### Description
The [DealLabel is a kinded](https://github.com/filecoin-project/go-state-types/blob/v0.1.9/builtin/v8/market/deal.go#L34-L40) union of string or byte slice and serializes to a CBOR string or CBOR byte string depending on which form it takes. Lilys model containing DealProposal's expects a string: https://github.com/filecoin-project/lily/blob/v0.10.0/model/actors/market/dealproposal.go#L50. We need to decide if we should provide a migration to store the byte representation of this value, or drop the value if it's not a string

### Acceptance criteria
- [ ] Decide if it's worth writing a migration to store byte representation of DealProposal Label value.

frrist:
- fixes #987


hsanjuan:
Why not migrating the current table to have one extra column (IsString), and start using that column from now on?

When reprocessing the chain then we will get the field correctly populated (we don't have to do it now).

Having two tables is a pain. We always say that we are going to add views and what not, but we don't, it complicates the exporting etc.

Given that we have to do a full chain reprocessing soon (where we can go to having 1 table only again), it would be nice to not have 2 tables in the meantime.

frrist:
Agreed, having two tables is complicated. Previously when adding a [migration similar to this](https://github.com/filecoin-project/lily/blob/master/schemas/v1/5_miner_sector_snaps.go) we did so to prevent breaking the archive process. But since we are planning to do a full chain reprocess soon I'd be open to making this an addition to the existing table schema. cc @kasteph

frrist:
@hsanjuan and @kasteph could you share more of your thoughts here? Right now this PR has created a new model/table. But if were wanting to go a different route we'll need to know before the week is over.

hsanjuan:
> @hsanjuan and @kasteph could you share more of your thoughts here? Right now this PR has created a new model/table. But if were wanting to go a different route we'll need to know before the week is over.

Iirc, single table should be fine. Add one extra column but don't populate it for the things in the past, and start using the new format on the existing label column too (no need to base64 all existing labels in all existing rows).

frrist:
closing in favor of https://github.com/filecoin-project/lily/pull/1015

PR #1008, opened 2022-06-24
feat: sync wait --until-epoch N

placer14:


commands/sync.go
```diff
@@ -78,28 +79,50 @@ var SyncStatusCmd = &cli.Command{
...
 	},
 	Action: func(cctx *cli.Context) error {
+		if opts.untilEpoch > 0 && opts.watch {
+			return fmt.Errorf("until-epoch and watch flags cannot be used together")
+		}
+		// ensure default doesn't cause an internal exit by making the value
+		// impossibly large
+		if opts.untilEpoch == 0 {
+			opts.untilEpoch = abi.ChainEpoch(math.MaxInt64)
+		}
```
placer14, line 112: I didn't want to set MaxInt64 as the default in the arg because it would be misleading. This is janky, but as a quick hack it seems acceptable.

PR #1010, opened 2022-06-26
chore: add output flag to sync status command

kasteph:
## Background
I frequently have to stop a Lily daemon manually when I do backfills or partial repository archiving work. Until filecoin-project/lotus#8641 is ready, I need to be able to have a way to determine what the latest height is.

## Changes

~~This PR accepts a string flag to the `stop` command that determines at the most recent height the given stage stopped. This allows me to use the height in `stdout` to be used as an input for next steps in the data pipeline.~~

It accepts a bool flag to `sync` called `output` that would enable me to continuously poll the sync command with simple bash one-liners and check at what stage and what height it currently is at.

e.g.

```
❯ ./lily sync status --output="text"
2022-06-27T13:11:35.035+0200    WARN    cliutil util/apiinfo.go:81      API Token not set and requested, capabilities might be limited.
header sync 1072527
``` 

```
❯ ./lily sync status --output="json"
2022-06-27T16:16:06.003+0200    WARN    cliutil util/apiinfo.go:81      API Token not set and requested, capabilities might be limited.
{"Stage":1,"Height":1075951}

```

### DIFF
--- a/commands/sync.go
+++ b/commands/sync.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
@@ -16,6 +17,11 @@ import (
 	"github.com/filecoin-project/lily/lens/lily"
 )
 
+type SyncStatus struct {
+	Stage  api.SyncStateStage
+	Height abi.ChainEpoch
+}
+
 var SyncCmd = &cli.Command{
 	Name:  "sync",
 	Usage: "Inspect or interact with the chain syncer",
@@ -28,6 +34,14 @@ var SyncCmd = &cli.Command{
 var SyncStatusCmd = &cli.Command{
 	Name:  "status",
 	Usage: "Report sync status of a running lily daemon",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:     "output",
+			Usage:    "Print only the current sync stage at the latest height. One of [text, json]",
+			Aliases:  []string{"o"},
+			Required: false,
+		},
+	},
 	Action: func(cctx *cli.Context) error {
 		ctx := lotuscli.ReqContext(cctx)
 		lapi, closer, err := GetAPI(ctx)
@@ -41,9 +55,16 @@ var SyncStatusCmd = &cli.Command{
 			return err
 		}
 
-		fmt.Println("sync status:")
+		output := cctx.String("output")
+
+		var max abi.ChainEpoch = -1
+		maxStateSync := api.StageIdle
 		for _, ss := range state.ActiveSyncs {
-			fmt.Printf("worker %d:\n", ss.WorkerID)
+			if max < ss.Height && maxStateSync <= ss.Stage {
+				max = ss.Height
+				maxStateSync = ss.Stage
+			}
+
 			var base, target []cid.Cid
 			var heightDiff int64
 			var theight abi.ChainEpoch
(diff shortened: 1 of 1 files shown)


commands/stop.go
```diff
@@ -1,13 +1,23 @@
...
 )
 
 var StopCmd = &cli.Command{
 	Name:  "stop",
 	Usage: "Stop a running lily daemon",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:   "height",
+			Usage:  "Get the last chain height during the given sync stage. One of [header, message]",
+			Hidden: false,
```
kasteph, line 18: Adding this explicitly because I want to put this up for discussion. We could potentially keep it hidden so that Lily users are not necessarily encouraged to use this.

commands/sync.go
```diff
@@ -41,9 +49,24 @@ var SyncStatusCmd = &cli.Command{
 			return err
 		}
 
-		fmt.Println("sync status:")
+		short := cctx.Bool("short")
```
iand, line 52: Wouldn't it be cleaner to add an `output` flag which can be `text` or `json`. The default is text, but you could add json output and then query that using `jq` or similar?

commands/sync.go
```diff
@@ -58,21 +81,27 @@ var SyncStatusCmd = &cli.Command{
...
+				fmt.Printf("\tHeight: %d\n", ss.Height)
+				if ss.End.IsZero() {
+					if !ss.Start.IsZero() {
+						fmt.Printf("\tElapsed: %s\n", time.Since(ss.Start))
+					}
+				} else {
+					fmt.Printf("\tElapsed: %s\n", ss.End.Sub(ss.Start))
+				}
+				if ss.Stage == api.StageSyncErrored {
+					fmt.Printf("\tError: %s\n", ss.Message)
```
iand, line 99: I think the short output should indicate whether there is an error or not. Not relevant if you choose to output json instead

commands/stop.go
```diff
@@ -16,6 +26,37 @@ var StopCmd = &cli.Command{
 		}
 		defer closer()
 
+		if cctx.String("height") != "" {
```
iand, line 29: Is this necessary. Can't you always do `lily sync status` just before calling `lily stop` to get the same information?

kasteph: You're right, I'll remove this.

placer14 (changes requested):

commands/sync.go
```diff
@@ -58,21 +80,41 @@ var SyncStatusCmd = &cli.Command{
...
+						return err
+					}
+					fmt.Printf(string(j) + "\n")
+				case "text":
+					fallthrough
+				default:
+					fmt.Printf("%s %d\n", maxStateSync, max)
+				}
+
 			}
```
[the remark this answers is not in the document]
kasteph: Makes sense! Thank you!

commands/sync.go
```diff
@@ -28,6 +34,15 @@ var SyncCmd = &cli.Command{
 var SyncStatusCmd = &cli.Command{
 	Name:  "status",
 	Usage: "Report sync status of a running lily daemon",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:     "output",
+			Usage:    "Print only the current sync stage at the latest height. One of [text, json]",
+			Aliases:  []string{"o"},
+			Required: false,
+			Value:    "text",
```
frrist, line 43: When the flag isn't provided its default value will be `"text"`, is this intended? In order for this [case](https://github.com/filecoin-project/lily/pull/1010/files#diff-ab57c1124da90ef9a1f72210e299ca8a5facfb7a36b704dcbeb803957c6b9525R91) to evaluate the user will need to explicitly set the flags value to `""`.

kasteph: good catch, thanks

commands/sync.go
```diff
@@ -58,21 +79,38 @@ var SyncStatusCmd = &cli.Command{
...
+			switch output {
+			case "json":
+				j, err := json.Marshal(SyncStatus{Stage: maxStateSync, Height: max})
+				if err != nil {
+					return err
 				}
-			} else {
-				fmt.Printf("\tElapsed: %s\n", ss.End.Sub(ss.Start))
+				fmt.Printf(string(j) + "\n")
+			case "":
```
frrist, line 90: could call this "pretty" or "verbose" and set the default value of the `--output` flag as such, but not a big deal

frrist (approved):
LGTM :shipit:

PR #1018, opened 2022-07-05
fix: many incorrect pointer equality comparisons

### LINKED ISSUE #1017: Missing data from `miner_fee_debt` at epoch `1957803`
### Describe the bug:

Expecting data from SP `f07830` when running task `miner_fee_debt` at epoch `1957803`. When running this task focused on this epoch, the output does not include this miner's record.

We can confirm the extractor is executing where lily detects the state diff for this SP.

We expect data at this height for this SP due to data presence from Lily v0.10.1 in production.

### Steps to Reproduce:

1. Run lily v0.11.0 w state that includes epoch `1957803`.
2. Execute `lily job run --tasks="miner_fee_debt" --storage="File1" index height --height=1957803` with storage configured for writing CSVs:

```
        [Storage.File.File1]
          Format = "CSV"
          Path = "/var/lib/lily/"
          OmitHeader = false
          FilePattern = "{table}.csv"
```
3. Observe missing records with `cat /var/lib/lily/miner_fee_debt.csv | grep f07830`.

frrist:
fixes #1017 

- the hardest of facepalms

### DIFF
--- a/tasks/actorstate/miner/context.go
+++ b/tasks/actorstate/miner/context.go
@@ -21,41 +21,35 @@ func NewMinerStateExtractionContext(ctx context.Context, a actorstate.ActorInfo,
 		return nil, fmt.Errorf("loading current miner state: %w", err)
 	}
 
-	prevTipset := a.Current
-	var prevState miner.State
-	if a.Current.Height() != 1 {
-		prevTipset = a.Executed
-
-		prevActor, err := node.Actor(ctx, a.Address, a.Executed.Key())
-		if err != nil {
-			// if the actor exists in the current state and not in the parent state then the
-			// actor was created in the current state.
-			if err == types.ErrActorNotFound {
-				return &MinerStateExtractionContext{
-					PrevState:     prevState,
-					PrevTs:        prevTipset,
-					CurrActor:     &a.Actor,
-					CurrState:     curState,
-					CurrTs:        a.Current,
-					PreviousState: false,
-				}, nil
-			}
-			return nil, fmt.Errorf("loading previous miner %s at tipset %s epoch %d: %w", a.Address, a.Executed.Key(), a.Current.Height(), err)
+	prevActor, err := node.Actor(ctx, a.Address, a.Executed.Key())
+	if err != nil {
+		// actor doesn't exist yet, may have just been created.
+		if err == types.ErrActorNotFound {
+			return &MinerStateExtractionContext{
+				PrevTs:               a.Executed,
+				CurrActor:            &a.Actor,
+				CurrState:            curState,
+				CurrTs:               a.Current,
+				PrevState:            nil,
+				PreviousStatePresent: false,
+			}, nil
 		}
+		return nil, fmt.Errorf("loading previous miner %s at tipset %s epoch %d: %w", a.Address, a.Executed.Key(), a.Current.Height(), err)
+	}
 
-		prevState, err = node.MinerLoad(node.Store(), prevActor)
-		if err != nil {
-			return nil, fmt.Errorf("loading previous miner actor state: %w", err)
-		}
+	// actor exists in previous state, load it.
+	prevState, err := node.MinerLoad(node.Store(), prevActor)
+	if err != nil {
+		return nil, fmt.Errorf("loading previous miner actor state: %w", err)
 	}
 
 	return &MinerStateExtractionContext{
-		PrevState:     prevState,
-		PrevTs:        prevTipset,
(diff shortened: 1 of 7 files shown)


tasks/actorstate/miner/context.go
```diff
@@ -49,23 +50,25 @@ func NewMinerStateExtractionContext(ctx context.Context, a actorstate.ActorInfo,
...
 	PrevState miner.State
 	PrevTs    *types.TipSet
 
-	CurrActor *types.Actor
-	CurrState miner.State
-	CurrTs    *types.TipSet
+	CurrActor     *types.Actor
+	CurrState     miner.State
+	CurrTs        *types.TipSet
+	PreviousState bool
```
placer14, line 69: nit: A slightly more expressive name:
```suggestion
	PreviousStateExists bool
```

tasks/actorstate/miner/deadline_info.go
```diff
@@ -36,7 +36,9 @@ func (DeadlineInfoExtractor) Extract(ctx context.Context, a actorstate.ActorInfo
 		if err != nil {
 			return nil, err
 		}
-		if prevDeadlineInfo == currDeadlineInfo {
+		// TODO implement equality function
+		// dereference pointers to check equality
+		if *prevDeadlineInfo == *currDeadlineInfo {
```
placer14, line 41: This is brittle. We have to be sure that `err == nil` guarantees we're protected from a nil-deref panic here and forever into the future. How is that being guaranteed? I would prefer this check is explicit here instead of expecting it inside the caller. If we have a local equality function doing that, that could work just a well.

placer14, line 41: This is brittle. We have to be sure that `err == nil` guarantees we're protected from a nil-deref panic here and forever into the future. How is that being guaranteed? I would prefer this check is explicit here instead of expecting it inside the caller. If we have a local equality function doing that, that could work just a well.

```suggestion
		if prevDeadlineInfo != nil &&
		   currDeadlineInfo != nil &&
		   *prevDeadlineInfo == *currDeadlineInfo {
```

tasks/actorstate/miner/locked_funds.go
```diff
@@ -36,7 +36,11 @@ func (LockedFundsExtractor) Extract(ctx context.Context, a actorstate.ActorInfo,
 		if err != nil {
 			return nil, fmt.Errorf("loading previous miner locked funds: %w", err)
 		}
-		if prevLocked == currLocked {
+
+		// if all values are equal there is no change.
+		if prevLocked.VestingFunds.Equals(currLocked.VestingFunds) &&
+			prevLocked.PreCommitDeposits.Equals(currLocked.PreCommitDeposits) &&
+			prevLocked.InitialPledgeRequirement.Equals(currLocked.InitialPledgeRequirement) {
```
placer14, line 43: Protect against nil-deref panic here?

tasks/actorstate/multisig/multisig.go
```diff
@@ -119,11 +119,12 @@ type MsigExtractionContext struct {
 	CurrState multisig.State
 	CurrTs    *types.TipSet
 
-	Store adt.Store
+	Store         adt.Store
+	PreviousState bool
```
placer14, line 123: nit:
```suggestion
	PreviousStateExists bool
```

tasks/actorstate/verifreg/verifreg.go
```diff
@@ -20,11 +20,12 @@ type VerifiedRegistryExtractionContext struct {
 	PrevState, CurrState verifreg.State
 	PrevTs, CurrTs       *types.TipSet
 
-	Store adt.Store
+	Store         adt.Store
+	PreviousState bool
```
placer14, line 24: nit:
```suggestion
	HasPreviousState bool
```

placer14 (approved):
Some nits for safety and clarity. Otherwise, LGTM

tasks/actorstate/miner/locked_funds.go
frrist: the `LockedFunds` type isn't a pointer, nor are its fields so we can safely use them here.

tasks/actorstate/miner/deadline_info.go
frrist: Since we know the miner has state we can safely assume it has non-nil deadline information. (the method that creates deadline info will never return null: https://github.com/filecoin-project/specs-actors/blob/v8.0.1/actors/builtin/miner/deadlines.go#L15

PR #1022, opened 2022-07-08
Fix a typo or a bug for command help text: job run find

cxytz01:
the help text's example is wrong, the task name: block_headers will not match the` tasktype.TableLookup.`


cxytz01:
https://lilium.sh/data/dumps/

kasteph (approved):
thanks!

PR #1032, opened 2022-08-13
fix: extract miner sector state changes

### LINKED ISSUE #1031: Fault/recover event still happen after a termination event
### Describe the bug:
image
Some sectors seem to fault/recover even after a termination event.

frrist:
This PR fixes #1031 

The previous implementation diffed previous and current partitions directly. This was incorrect since miners
may compact their partitions resulting in sectors moving to different partitions, which the previous code proceeded to record as invalid events. Instead, we now collect all sectors from all miner partitions and diff the sets directly.


tasks/actorstate/miner/sector_events.go
```diff
@@ -161,117 +246,136 @@ func ExtractMinerPreCommitEvents(extState extraction.State, preCommits *miner.Pr
...
-
-	if dlDiff == nil {
-		return nil, nil
-	}
-
-	removed := bitfield.New()
-	faulted := bitfield.New()
-	recovered := bitfield.New()
-	recovering := bitfield.New()
+// SectorStates contains a set of bitfields for active, live, fault, and recovering sectors.
```
placer14, line 249: Would be good to mention how this bitfield represents each type of sector represented here. (a packed binary list of sector IDs?)

tasks/actorstate/miner/sector_events.go
```diff
@@ -161,117 +246,136 @@ func ExtractMinerPreCommitEvents(extState extraction.State, preCommits *miner.Pr
...
-		return nil, fmt.Errorf("walking miners recovered sectors: %w", err)
+	// current faulty sectors minus previous faulty sectors are sectors faulted this epoch.
+	faulted, err := bitfield.SubtractBitField(current.Faulty, previous.Faulty)
+	if err != nil {
+		return nil, fmt.Errorf("comparing current faulty sectors to previous faulty sectors %w", err)
 	}
-	return out, nil
+
+	// previous faulty sectors minus current active sectors are sectors recovered this epoch.
+	recovered, err := bitfield.IntersectBitField(previous.Faulty, current.Active)
```
placer14, line 369: The comment and code don't match. I assume we only want to see which previous faults overlap with current actives... this makes sense as being "Recovered". Let's update this comment.

`// previous faulty sectors which match (intersect) active sectors are sectors recovered this epoch.`

placer14 (approved):
This seems good to me.

A thought: It would probably be easy to write a query which proves the invariant isn't in our exports which we could periodically run over our exported data.

Can you outline/include how this data bug was detected originally? I can see how we can run tests like these in CI. (Ideas @kasteph?)

tasks/actorstate/miner/sector_events.go
frrist: Whoops, bad copy paste

frrist:
> A thought: It would probably be easy to write a query which proves the invariant isn't in our exports which we could periodically run over our exported data.

@davidgasquez is working on this as a part of the dbt workflow.

> Edit: Would the check be as simple as taking all Termination events and making sure that these sectors never have additional events after the Termination occurs?

That would be part of the check, but there are more cases, for example - sectors cannot recover before becoming faulted, sectors cannot terminate before being added, faulted sectors cannot become faulty, etc.

A sector lifecycle is a state machine, so we'll want to validate all sector events correspond to valid states.

placer14:
My approval is sticky. Good catch on accidental context sharing. 🙇

PR #1040, opened 2022-08-18
Replace ExecutedAndBlockMessages with individual methods

frrist:
### What
This PR removes the `ExecutedAndBlockMessages` method and replaces it smaller more specific methods: `TipSetMessageReceipts`, `TipSetBlockMessages`, and `ComputeGasOutputs`. 

While implementing, I noticed a few bugs in the now replaced `ExecutedAndBlockMessages` method, which are tracked in #1045 and #1043 and fixed in this PR. Furthermore, I took the opportunity to address #1041 while implementing `TipSetMessageReceipts`.

### DIFF
--- a/lens/interface.go
+++ b/lens/interface.go
@@ -91,23 +91,16 @@ type BlockMessages struct {
 	SecpMessages []*types.SignedMessage // SECP messages in block `Block`
 }
 
+// BlockMessageReceipts contains a block its messages and their corresponding receipts.
+// The Receipts are one-to-one with Messages index.
 type BlockMessageReceipts struct {
-	Block                 *types.BlockHeader
-	Message               []types.ChainMsg
-	Receipt               []*types.MessageReceipt
+	Block *types.BlockHeader
+	// Messages contained in Block.
+	Messages []types.ChainMsg
+	// Receipts contained in Block.
+	Receipts []*types.MessageReceipt
+	// MessageExectionIndex contains a mapping of Messages to their execution order in the tipset they were included.
 	MessageExecutionIndex map[types.ChainMsg]int
-}
-
-func (bmr *BlockMessageReceipts) Iterator() (*MessageReceiptIterator, error) {
-	if len(bmr.Message) != len(bmr.Receipt) {
-		return nil, fmt.Errorf("invalid construction, expected equal number receipts (%d) and messages (%d)", len(bmr.Receipt), len(bmr.Message))
-	}
-	return &MessageReceiptIterator{
-		idx:      0,
-		msgs:     bmr.Message,
-		receipts: bmr.Receipt,
-		exeIdx:   bmr.MessageExecutionIndex,
-	}, nil
 }
 
 type MessageReceiptIterator struct {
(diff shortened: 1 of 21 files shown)


lens/interface.go
```diff
@@ -78,21 +85,56 @@ type MessageExecution struct {
...
+}
+
+func (mri *MessageReceiptIterator) HasNext() bool {
+	if mri.idx < len(mri.msgs) {
+		return true
+	}
+	return false
+}
+
+func (mri *MessageReceiptIterator) Next() (types.ChainMsg, int, *types.MessageReceipt) {
```
kasteph, line 127:
```suggestion
func (mri *MessageReceiptIterator) Next() (types.ChainMsg, *types.MessageReceipt) {
```
is the index necessary as part of the return value? i did a quick scan of where this is being called and only saw it once [here](https://github.com/filecoin-project/lily/pull/1040/files#diff-9134a88da118334f295ecafd8ba9e83fb372a7922b753bb32f7f4b3099cf2c1dR72).

could the index also be accessed as:
```go
msg, rec := itr.Next()
idx := itr.exeIdx[msg]
```

lens/lily/impl.go
```diff
@@ -544,6 +538,179 @@ func (m *LilyNodeAPI) GetMessageExecutionsForTipSet(ctx context.Context, next *t
...
+		out = append(out, &lens.BlockMessages{
+			Block:        blk,
+			BlsMessages:  blkMsgs.BlsMessages,
+			SecpMessages: blkMsgs.SecpkMessages,
+		})
+	}
+	return out, nil
+}
+
+// TipSetMessageReceipts returns the blocks and messages in `pts` and their corresponding receipts from `ts` matching block order in tipset (`pts`).
```
kasteph, line 564: What does `pts` mean? parent tipset?

kasteph (approved):
I didn't see any glaring logical errors in any of the tasks nor in the `TipSetMessageReceipts()` function. My previous comments are not blockers.

lens/lily/impl.go
```diff
@@ -544,6 +538,179 @@ func (m *LilyNodeAPI) GetMessageExecutionsForTipSet(ctx context.Context, next *t
...
+			// loading after computing state should succeed as tipset computation produces message receipts
+			rs, err = adt.AsArray(m.Store(), ts.Blocks()[0].ParentMessageReceipts)
+			if err != nil {
+				return nil, fmt.Errorf("load message receipts after tipset execution (something if very wrong contact a developer): %w", err)
+			}
+		} else {
+			return nil, fmt.Errorf("loading message receipts %w", err)
+		}
+	}
+	// so we only load the receipt array one
```
placer14, line 600:
```suggestion
	// so we only load the receipt array once
```

lens/lily/impl.go
```diff
@@ -544,6 +538,179 @@ func (m *LilyNodeAPI) GetMessageExecutionsForTipSet(ctx context.Context, next *t
...
+		vmi, err := vm.NewVM(ctx, &vm.VMOpts{
+			StateBase:      ts.ParentState(),
+			Epoch:          ts.Height(),
+			Bstore:         m.ChainAPI.Chain.StateBlockstore(),
+			Actors:         filcns.NewActorRegistry(),
+			Syscalls:       m.StateManager.Syscalls,
+			CircSupplyCalc: m.StateManager.GetVMCirculatingSupply,
+			NetworkVersion: util.DefaultNetwork.Version(ctx, ts.Height()),
+			BaseFee:        ts.Blocks()[0].ParentBaseFee,
+		})
```
placer14, line 684: Is the VM very "heavy" in size? Should `vmi` be memoized within `LilyNodeAPI` and have the appropriate statetree applied at the time of request within the wrapper?

placer14 (approved):

lens/lily/impl.go
frrist: Great question. Currently, this method is only called by the [GasOuptut](https://github.com/filecoin-project/lily/pull/1040/files#diff-da0b890b207570c25f37d6773fb522f78ad0718f242c5be13f4dbfa04a053f58R77) task, so until another task needs to call this method we can leave out caching.

PR #1042, opened 2022-08-23
chore: Some UX around the `chain state-*` commands

placer14:
- state list outputs JSON
- all commands under `state-*` prefix
- fix bug in `for` loop breaking on null rounds

### DIFF
--- a/lens/lily/impl.go
+++ b/lens/lily/impl.go
@@ -652,7 +652,7 @@ func (m *LilyNodeAPI) FindOldestState(ctx context.Context, limit int64) ([]*Stat
 			HasReceipts: maybeFullTS.HasReceipts,
 			HasState:    maybeFullTS.HasState,
 		})
-		if (head.Height()-abi.ChainEpoch(i)) <= oldestEpochLimit {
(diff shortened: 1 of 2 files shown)


lens/lily/impl.go
```diff
@@ -651,7 +652,7 @@ func (m *LilyNodeAPI) FindOldestState(ctx context.Context, limit int64) ([]*Stat
 			HasReceipts: maybeFullTS.HasReceipts,
 			HasState:    maybeFullTS.HasState,
 		})
-		if !maybeFullTS.HasState || int64(head.Height()-abi.ChainEpoch(i)) == limit {
+		if (head.Height()-abi.ChainEpoch(i)) <= oldestEpochLimit {
```
placer14, line 655: Here's the null round fix. (remove `maybeFullTS.HasState`) I assume we don't want to break if we hit a null round?

frrist:
Looks great, thanks will incorporate with [frrist/find-oldest-statetree](https://github.com/filecoin-project/lily/tree/frrist/find-oldest-statetree)

PR #1044, opened 2022-08-23
feat(docker): Update docker-compose to use lily worker pattern

### LINKED ISSUE #1034: Create functional docker-compose for running lily with distributed indexing
### Description
To make local development easier a docker-compose file should be provided that allows users to run a set of lily nodes that perform indexing using the distributed indexer pattern.

### Acceptance criteria
- [ ] create Redis instance
- [ ] create a configurable number of lily nodes

placer14:
Closes #1034 

- remove `daemon` and replace w `notifier` and `worker` containers
- `notifier` and `worker` each have it's own envvars and config
- redis instance is brought up first and both `notifier` and `worker` successfully register their queues
- prometheus should be target scraping both though I haven't tested it (or jaeger/tasks as I don't have a repo handy)

### DIFF
--- a/build/lily/docker_init.sh
+++ b/build/lily/docker_init.sh
@@ -4,14 +4,17 @@ set -exo pipefail
 
 mkdir -p ${LILY_REPO}/keystore
 
-if [[ ! -z "${_LILY_DOCKER_INIT_IMPORT_MAINNET_SNAPSHOT}" ]]; then
-  # import snapshot when _LILY_DOCKER_INIT_IMPORT_MAINNET_SNAPSHOT is set
+
+if [[ ! -z "${LILY_DOCKER_INIT_IMPORT_MAINNET_SNAPSHOT}" ]]; then
+  # set default snapshot path if not already defined
+  snapshot="${LILY_DOCKER_INIT_IMPORT_SNAPSHOT_PATH:-https://snapshots.mainnet.filops.net/minimal/latest}"
+
+  # import snapshot when LILY_DOCKER_INIT_IMPORT_MAINNET_SNAPSHOT is set
   if [[ -f "${LILY_REPO}/datastore/_imported" ]]; then
     echo "Skipping import, found ${LILY_REPO}/datastore/_imported file."
   else
-    echo "Importing snapshot from https://fil-chain-snapshots-fallback.s3.amazonaws.com/mainnet/minimal_finality_stateroots_latest.car..."
-    lily init --import-snapshot="https://fil-chain-snapshots-fallback.s3.amazonaws.com/mainnet/minimal_finality_stateroots_latest.car"
-
+    echo "Importing snapshot from ${snapshot}"
+    lily init --import-snapshot=${snapshot}
     status=$?
     if [ $status -eq 0 ]; then
       touch "/var/lib/lily/datastore/_imported"
--- a/build/lily/docker_init.sh
+++ b/build/lily/docker_init.sh
@@ -22,6 +22,6 @@ else
   lily init
 fi
 
-chmod 0600 ${LILY_REPO}/keystore/*
+chmod -R 0600 ${LILY_REPO}/keystore
 
 lily $@
--- a/build/lily/notifier.env
+++ b/build/lily/notifier.env
@@ -5,19 +5,20 @@ LILY_REPO=/var/lib/lily
 #LILY_CONFIG=/var/lib/lily/config.toml
 
 # Postgres URL may be overridden
-LILY_STORAGE_POSTGRESQL_DB_URL="postgres://postgres@timescaledb:5432/postgres"
+LILY_STORAGE_POSTGRESQL_DB_URL=postgres://postgres@timescaledb:5432/postgres?sslmode=disable
 
 # Enable IMPORT_SNAPSHOT below to use snapshot on lily startup
-#_LILY_DOCKER_INIT_IMPORT_MAINNET_SNAPSHOT=true
+#LILY_DOCKER_INIT_IMPORT_MAINNET_SNAPSHOT=true
+#LILY_DOCKER_INIT_IMPORT_SNAPSHOT_PATH=
 
 # Debugging options
 #LILY_TRACING=true
-LILY_PROMETHEUS_PORT="localhost:9090"
+LILY_PROMETHEUS_PORT=0.0.0.0:9090
 
 # Logging options
 GOLOG_LOG_FMT=json
 GOLOG_LOG_LEVEL=info
-LILY_LOG_LEVEL_NAMED="vm:error,badgerbs:error"
+LILY_LOG_LEVEL_NAMED=vm:error,badgerbs:error
 
 # Queue configuration
-LILY_REDIS_ADDR="redis-queue:6379"
+LILY_REDIS_ADDR=redis-queue:6379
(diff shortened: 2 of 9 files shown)


placer14:
@frrist This has the basis for what you asked, but we can't use (docker-compose) replicas for the worker. (We can't assign their ports (API and prom) uniquely and only the first binding succeeds.) The next best thing we can do is make additional boilerplate for `worker-2` and `worker-3` that is commented out but can be enabled and works. What do you think?

Could you also test this out with a repo and see if anything here doesn't make sense?

build/prometheus/prometheus.yml
```diff
@@ -4,4 +4,4 @@ scrape_configs:
     scrape_interval: 5s
 
     static_configs:
-      - targets: ['localhost:9991']
+      - targets: ['notifier:9090', 'worker:9091']
```
placer14, line 7: This is using `notifier` and `worker` instead of `localhost` because I had to disable `network_mode: host`. If you want host networking back, these will have to return to `localhost`.

docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
       - '--web.console.templates=/usr/share/prometheus/consoles'
     ports:
       - 9090:9090
-    network_mode: "host"
+    #network_mode: "host"
     restart: always
 
   grafana:
```
placer14, line 43: nit: We should pick a version number for this and pin it. Latest current major version okay with you?

docker-compose.yml
```diff
@@ -47,38 +47,77 @@ services:
...
+      # holds lily datastore repo
+      - lily_worker_data:/var/lib/lily
+      # persist params through restarts
+      - lily_worker_tmp:/var/tmp/filecoin-proof-parameters
+      # notifier-specific config
+      - ./build/lily/worker_config.toml:/var/lib/lily/config.toml
+      - ./build/lily/docker_init.sh:/usr/bin/docker_init.sh
+    entrypoint: /usr/bin/docker_init.sh
+    command:
+      - "daemon --bootstrap=false"
```
placer14, line 114: Note: `--bootstrap=false` here for testing containers. May need to remove this.

docker-compose.yml
```diff
@@ -47,38 +47,77 @@ services:
...
+      # holds lily datastore repo
+      - lily_notifier_data:/var/lib/lily
+      # persist params through restarts
+      - lily_notifier_tmp:/var/tmp/filecoin-proof-parameters
+      # notifier-specific config
+      - ./build/lily/notifier_config.toml:/var/lib/lily/config.toml
+      - ./build/lily/docker_init.sh:/usr/bin/docker_init.sh
+    entrypoint: /usr/bin/docker_init.sh
+    command:
+      - "daemon --bootstrap=false"
```
placer14, line 89: Note: `--bootstrap=false` here for testing containers. May need to remove this.

docker-compose.yml
```diff
@@ -47,38 +47,77 @@ services:
...
 
 volumes:
   timescaledb: {}
   prometheus_data: {}
   grafana_data: {}
-  lily_data: {}
+  lily_notifier_data: {}
+  lily_worker_data: {}
+  lily_notifier_tmp: {}
+  lily_worker_tmp: {}
```
placer14, line 123: The `lily_*_tmp` volumes will persist the params so you don't have to keep downloading them. <3

docker-compose.yml
frrist: I think leaving this as is (bootstrap=false) makes sense for now 👍

docker-compose.yml
frrist: Yup sounds good.

build/prometheus/prometheus.yml
frrist: Also seems fine, I generally only poke at Grafana anyways.

build/lily/notifier.env
```diff
@@ -5,16 +5,19 @@ LILY_REPO=/var/lib/lily
 #LILY_CONFIG=/var/lib/lily/config.toml
 
 # Postgres URL may be overridden
-#LILY_STORAGE_POSTGRESQL_DB_URL=postgres://user:pass@host:port/database?options
```
frrist, line 8: I think we'll want to remove the quotes from here and worker.env, else the container see:`"\"postgres://postgres@timescaledb:5432/postgres\""`

build/lily/worker.env
```diff
@@ -0,0 +1,23 @@
+# Path for lily state should be on a persisted volume
```
frrist, line 1: ditto comment about quoting

build/lily/docker_init.sh
```diff
@@ -22,6 +22,6 @@ else
   lily init
```
frrist, line 22: I can't comment on the line, but above on line 11, could we add an `elif` with something like:
```bash
elif [[ ! -z "${LILY_DOCKER_INIT_IMPORT_SNAPSHOT_PATH}" ]]; then
    echo "Importing snapshot from ${LILY_DOCKER_INIT_IMPORT_SNAPSHOT_PATH}"
    lily init --import-snapshot=${LILY_DOCKER_INIT_IMPORT_SNAPSHOT_PATH}
```
then we could optionally specify locally mounted snapshots in the compose file: https://github.com/filecoin-project/lily/pull/1044/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R86

docker-compose.yml
frrist: If it's not already, could the daemon's repo also be persisted? (so we don't redownload the snapshots)

build/lily/notifier.env
```diff
@@ -5,16 +5,19 @@ LILY_REPO=/var/lib/lily
...
+LILY_STORAGE_POSTGRESQL_DB_URL="postgres://postgres@timescaledb:5432/postgres"
```
frrist, line 8: append `?sslmode=disable` to this and the worker.env

build/lily/notifier.env
```diff
@@ -5,16 +5,19 @@ LILY_REPO=/var/lib/lily
...
 
 # Enable IMPORT_SNAPSHOT below to use snapshot on lily startup
 #_LILY_DOCKER_INIT_IMPORT_MAINNET_SNAPSHOT=true
 
 # Debugging options
 #LILY_TRACING=true
-LILY_PROMETHEUS_PORT="prometheus:9090"
+LILY_PROMETHEUS_PORT="localhost:9090"
```
frrist, line 15: im not sure this works, I wasn't able to scrap any metrics from the nodes.

docker-compose.yml
placer14: Both `worker` and `notifier` have their repo's persisted.

build/lily/notifier.env
placer14: I pushed a change if you want to try this again?

build/lily/docker_init.sh
placer14: I made the envvar get a default assignment if found to be empty.

placer14:
I confirmed getting metrics from worker and notifier into prom and seeing them in grafana. 🤝

build/lily/notifier_config.toml
```diff
@@ -0,0 +1,14 @@
+[API]
```
frrist, line 1: needs a storage config

frrist (approved):
LOVE IT!! Works great, good to merge after last few comments addressed

Issue #1046, opened 2022-08-24
Build Production Docker Image failing in CI


frrist:
### Describe the bug:
https://app.circleci.com/pipelines/github/filecoin-project/lily/3721/workflows/0ad7e358-b2a9-4a5b-ac72-3f5c03598476/jobs/17116 -> https://app.circleci.com/pipelines/github/filecoin-project/lily/3721/workflows/0ad7e358-b2a9-4a5b-ac72-3f5c03598476/jobs/17116/parallel-runs/0/steps/0-103

CI is failing to build docker images

### Steps to Reproduce:
Create a branch and push, let CI run, see failure

PR #1047, opened 2022-08-24
fix: upgrade docker engine

kasteph:
closes: #1046 

issue looks like glibc being incompatible with docker engine versions <= 20.10.9


frrist (approved):
love it! thanks

PR #1048, opened 2022-08-30
feat: implement miner protocol survey task

### LINKED ISSUE #989: Track Miner Deal Making protocol via survey job
### Description
Track the deal-making protocol version used by miners: https://boost.filecoin.io/configuration#dealmaking

### Acceptance criteria
- [ ] create new table for data
- [ ] implement new job on survey command

frrist:
- closes #989


tasks/survey/minerprotocols/minerprotocols.go
```diff
@@ -0,0 +1,177 @@
...
+			pool.Stop()
+			return nil, ctx.Err()
+		default:
+		}
+
+		miner := miner
+
+		// find the miner, if DNE abort as this indicates an error in the API as a miner was returned from StateListMiners that DNE in state tree.
+		// passing EmptyTSK causes API to use current chain head.
+		minerInfo, err := t.api.StateMinerInfo(ctx, miner, types.EmptyTSK)
```
frrist, line 91: TODO: only connect to active miners

tasks/survey/minerprotocols/minerprotocols.go
```diff
@@ -0,0 +1,177 @@
...
+
+var (
+	resultBufferEnv  = "LILY_SURVEY_MINER_PROTOCOL_BUFFER"
+	resultBufferSize = 50
+
+	workerPoolSizeEnv = "LILY_SURVEY_MINER_PROTOCOL_WORKERS"
+	workerPoolSize    = 50
+
+	fetchTimeoutEnv = "LILY_SURVEY_MINER_PROTOCOL_TIMEOUT_SECONDS"
+	fetchTimeout    = 30
```
placer14, line 30: Can you include docs in this PR (or link to an issue) for these envvars?

tasks/survey/minerprotocols/minerprotocols.go
```diff
@@ -0,0 +1,177 @@
...
+		case <-ctx.Done():
+			pool.Stop()
+			return nil, ctx.Err()
+		default:
+		}
+
+		miner := miner
+
+		// find the miner, if DNE abort as this indicates an error in the API as a miner was returned from StateListMiners that DNE in state tree.
+		// passing EmptyTSK causes API to use current chain head.
```
placer14, line 90: Wouldn't it be possible for a miner to fall out of `StateListMiners` between tipsets? If we're always querying HEAD while walking through miners, this failure case may happen more often than we would expect. Would it be better to always query using the same tipset key?

And even if you disagree, I don't think it's appropriate to bail on the rest of the miner list, is it? Maybe just continue instead?

placer14 (approved):

tasks/survey/minerprotocols/minerprotocols.go
frrist: Good catch, will fetch the current chain head at the start of the method and use it for all calls.

PR #1087, opened 2022-11-09
fix: remove actor code mapping

birdychang:
Without actor code mapping, MakeGetActorCodeFunc & all GetActorCode calls only take a few ms to execute.
Time to construct actor code mapping with 5M state store caching takes ~10s.
Time to construct actor code mapping without caching takes ~40s.

### DIFF
--- a/lens/util/repo.go
+++ b/lens/util/repo.go
@@ -25,7 +25,6 @@ import (
 	"github.com/filecoin-project/lotus/chain/vm"
 	"github.com/filecoin-project/specs-actors/actors/util/adt"
 
-	builtininit "github.com/filecoin-project/lily/chain/actors/builtin/init"
 	"github.com/filecoin-project/lily/lens"
 )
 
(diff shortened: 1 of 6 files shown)


lens/util/repo.go
```diff
@@ -201,10 +192,10 @@ func MakeGetActorCodeFunc(ctx context.Context, store adt.Store, next, current *t
 		// TODO accept a context, don't take the function context.
 		_, innerSpan := otel.Tracer("").Start(ctx, "GetActorCode")
 		defer innerSpan.End()
-		// Shortcut lookup before resolving
-		c, ok := actorCodes[a]
-		if ok {
-			return c, true
+
+		act, err := nextStateTree.GetActor(a)
+		if act != nil {
```
kasteph, line 197:
```suggestion
		if err != nil {
```

lens/util/repo.go
```diff
@@ -213,9 +204,9 @@ func MakeGetActorCodeFunc(ctx context.Context, store adt.Store, next, current *t
 			return cid.Undef, false
 		}
 
-		c, ok = actorCodes[ra]
-		if ok {
-			return c, true
+		act, err = nextStateTree.GetActor(ra)
+		if act != nil {
```
kasteph, line 208:
```suggestion
		if err != nil {
```

kasteph (commented):
I think these should get rid of the linting errors.

lens/util/repo.go
birdychang: If we don't get an object, the code should continue with fallback logic.

lens/util/repo.go
birdychang: Same. Only returns when we have an actor object. Otherwise, use the previous state tree.

PR #1089, opened 2022-11-15
refactor: polish GetActorCodeFunc

frrist:


birdychang (approved):
Thanks for the refactor. I noticed in [stateTree.GetActor()](https://github.com/filecoin-project/lotus/blob/04a8239df4edc678030f5b2964798b4b0ebb1c1a/chain/state/statetree.go#L358), it's doing similar lookup logic. Do we still need the extra ResolveAddress here?

frrist:
oooOOo good catch, will remove that call, merge this into your branch, and then move onto the master.

Issue #1083, opened 2022-10-24
TipSetCache's bug


cxytz01:
### Describe the bug:
I think there is a bug here.
In Watcher.init, warm function inserts the current tipset’s parents to ring buffer except for current tipset. https://github.com/filecoin-project/lily/blob/master/chain/watch/watcher.go#L112
And in watcher.index, function SetCurrent just replaces the tail of the buffer’s tipset to current tipset,  the tipset that has been replaced is deleted. The tipset has been deleted has no opportunity to be indexing.
https://github.com/filecoin-project/lily/blob/85faa2322268e69f7fa65c3a2dc9a3d582dc98ab/chain/cache/tipset.go#L116 

There are the tss variable data when warm function has been executed:
 tss height: 1409742
 tss height: 1409741
 tss height: 1409740
 tss height: 1409739
 tss height: 1409738
 tss height: 1409737
 tss height: 1409734
 tss height: 1409733
 tss height: 1409732
 tss height: 1409731
and the current height is: 1409744
when SetCurrent function has been executed, the buffer is:
 i: 0, height: 1409731
 i: 1, height: 1409732
 i: 2, height: 1409733
 i: 3, height: 1409734
 i: 4, height: 1409737
 i: 5, height: 1409738
 i: 6, height: 1409739
 i: 7, height: 1409740
 i: 8, height: 1409741
 i: 9, height: 1409744
height: 1409742 was deleted.

frrist:
@cxytz01 I suspect you are correct. Would love a PR that fixes this if you're up for it.

PR #1098, opened 2022-12-01
fix: TipSetCache inconsistent data  (filecoin-project/lily#1083)

chihmingyang:
Refactor the setCurrent method.


frrist:
Thanks for the fix @chihmingyang will review shortly

PR #1101, opened 2022-12-14
chore(docker): Cache proof params in production docker image

placer14:
closes https://github.com/filecoin-project/sentinel-infra/issues/408

Note the tradeoff here is a larger image (going from ~300MB to ~700MB) in exchange for shorter startup time and avoidance of potential future issues related to transient proof downloads.

(This also means that proof changes will require a new build/release.)


placer14:
The integration test appears to be a failure of IPFS and not the test itself. I will be unavailable to merge this over the break. Please merge at will.

frrist (approved):

Issue #1106, opened 2022-12-23
Duplicated rows in vm_messages task when using CSV as storage


davidgasquez:
### Describe the bug:
When running lily job run --storage CSV --task vm_messages walk --from X --to Y, the vm_messages.csv that is produced contains some duplicated rows.

### Steps to Reproduce: 

1. Initialize Lily from any snapshot and run a walk covering 200 epochs storing the results in CSVs: `lily job run --storage CSV --task vm_messages walk --from X --to Y`. I tried with `--from 100100 --to 100300` using `snapshot_97920_100802_1666950561.car.zst` available at `s3://sentinel-backfill/historical-exports/snapshot_97920_100802_1666950561.car.zst`.
2. See duplicated lines with `cat /tmp/vm_messages.csv | sort | uniq -d`.

This is the output of `2`. Every row is a duplicated found in the original CSV. 
```
height,state_root,cid,source,from,to,value,method,actor_code,exit_code,gas_used,params,returns
100299,bafy2bzacedrmbruuyltyesap455e53pkgn7ogatt2l5w5c7xirs4z7tjci3de,bafy2bzacedylon6kapiwufnddau6simggg7roabbdljzkx66rc3ccizixqi7o,bafy2bzaceamicuy2h55d2pwip5srxydiquj5rnginulosb7xzfch3bkqjwzke,f01235,f04,0,9,bafkqaetgnfwc6mjpon2g64tbm5sxa33xmvza,0,481415,,"{""PledgeCollateral"":{""Int"":8305121611547023530889461},""QualityAdjPower"":{""Int"":385958680480186368},""QualityAdjPowerSmoothed"":{""PositionEstimate"":{""Int"":131031567524352342715596315439041873899679366810938465141},""VelocityEstimate"":{""Int"":1338356856550395441892507665153955554806865091651389}},""RawBytePower"":{""Int"":385721185968586752}}"
[...]

PR #1107, opened 2023-01-03
chore: add unique constraint on vm_messages.params

kasteph:
Stacks on: #1105 (please review #1105 first before this one) 
Resolves: #1106

To prevent duplicate `vm_messages`, we can add a constraint on the `params`.

### DIFF
--- a/schemas/v1/14_vm_messages_uniq_params.go
+++ b/schemas/v1/14_vm_messages_uniq_params.go
@@ -1,0 +1,11 @@
+package v1
+
+func init() {
+	patches.Register(
+		14,
+		`
+ALTER TABLE {{ .SchemaName | default "public"}}.vm_messages
+  ADD CONSTRAINT vm_messages_uniq_params UNIQUE (height, state_root, cid, source, params);
+`,
+	)
+}


placer14:
Two thoughts:

Functionally: Could this constraint be reduced to fewer columns? The likelihood of a `(state_root, cid)` collision is practically non-existent (since the hash for the `cid` depends on the `state_root` hash). With this in hand, could we safely remove the `source` from the constraint to make insertion a little cheaper?

Overall: I'm also not sure how these duplicates are being introduced? Do the duplicate internal messages  (with the same params) actually occur at the same `(height, state_root, cid, source)` combinations? (The issue is vague on the cause of this problem.) Do we know why this is? Are they important to capture?

The best I can think is that the VM runs certain state changes over and over by virtue of the source message being included in multiple blocks within a single tipset. If this is the case, then the `param` unique constraint doesn't add anything as the CID would match (because params change the message CID, correct?).

kasteph:
> Functionally: Could this constraint be reduced to fewer columns? The likelihood of a (state_root, cid) collision is practically non-existent (since the hash for the cid depends on the state_root hash). With this in hand, could we safely remove the source from the constraint to make insertion a little cheaper?

That sounds good.

> Overall: I'm also not sure how these duplicates are being introduced? Do the duplicate internal messages (with the same params) actually occur at the same (height, state_root, cid, source) combinations? (The issue is vague on the cause of this problem.) Do we know why this is? Are they important to capture?

Forrest commented on Slack:

_Duplicates here are again an artifact of the table schema. There is no requirement that these messages are unique since they do not appear on chain. It is likely that case that these parameters are different for each of the duplicate messages._

frrist:
We need to hold on this change until more details are added to https://github.com/filecoin-project/lily/issues/1106. I don't have a clear picture of what problem we are trying to fix here - I actually don't see a problem.

frrist:
closing in favor of https://github.com/filecoin-project/lily/pull/1127

PR #1117, opened 2023-01-27
Some opinions on the new models

kasteph:
- add EventsRoot to Receipt
- add Signature to Message
- embed Message type inside VMMessage since it's basically that type [here](https://github.com/filecoin-project/lily/blob/frrist/etl-pkg/pkg/extract/chain/vm.go#L14)

### DIFF
--- a/pkg/transform/gorm/models/chainmessage.go
+++ b/pkg/transform/gorm/models/chainmessage.go
@@ -16,6 +16,7 @@ type Message struct {
 	GasPremium types.DbToken
 	Method     uint64
 	Params     []byte
+	Signature  string `gorm:"jsonb"`
 }
 
 type ParsedMessageParams struct {
--- a/pkg/transform/gorm/models/receipt.go
+++ b/pkg/transform/gorm/models/receipt.go
@@ -19,8 +19,9 @@ type MessageReceipt struct {
 }
 
 type Receipt struct {
-	Index    int64 `gorm:"primaryKey"`
-	ExitCode int64
-	GasUsed  int64
-	Return   []byte
+	Index      int64 `gorm:"primaryKey"`
+	ExitCode   int64
+	GasUsed    int64
+	Return     []byte
+	EventsRoot string
 }
--- a/pkg/transform/gorm/models/vmmessages.go
+++ b/pkg/transform/gorm/models/vmmessages.go
@@ -5,11 +5,8 @@ import "github.com/filecoin-project/lily/pkg/transform/gorm/types"
 type VmMessage struct {
 	Source  types.DbCID `gorm:"primaryKey"`
 	Cid     types.DbCID `gorm:"primaryKey"`
-	To      types.DbAddr
-	From    types.DbAddr
-	Value   types.DbToken
-	Method  uint64
-	Params  []byte
-	Receipt Receipt `gorm:"embedded"`
+	Message Message     `gorm:"embedded"`
+	Receipt Receipt     `gorm:"embedded"`
 	Error   string
+	Index   uint64
 }


pkg/transform/gorm/models/chainmessage.go
```diff
@@ -16,6 +16,7 @@ type Message struct {
 	GasPremium types.DbToken
 	Method     uint64
 	Params     []byte
+	Signature  string `gorm:"jsonb"`
```
frrist, line 19: Perhaps this should be raw bytes instead of json. The message signature type can be marshalled to CBOR before persisting to this field: https://github.com/filecoin-project/go-state-types/blob/master/crypto/signature.go#L51

pkg/transform/gorm/models/vmmessages.go
```diff
@@ -5,11 +5,8 @@ import "github.com/filecoin-project/lily/pkg/transform/gorm/types"
...
 	Source  types.DbCID `gorm:"primaryKey"`
 	Cid     types.DbCID `gorm:"primaryKey"`
-	To      types.DbAddr
-	From    types.DbAddr
-	Value   types.DbToken
-	Method  uint64
-	Params  []byte
-	Receipt Receipt `gorm:"embedded"`
+	Message Message     `gorm:"embedded"`
+	Receipt Receipt     `gorm:"embedded"`
```
frrist, line 9: While it is true that VM messages are basically messages, many of the fields such as `GasLimit`, `GasFee`, and `GasPremium` are all `0`. Similarly, VMMessages do not contain a `Signature`, nor do their receipts contain an `EventRoot` (less sure on this, but a hunch). I think it may be more efficient to avoid embedding the `Message` and `Receipt` structures in `VmMessgae` and instead take a subset of the fields that are populated. wdyt?

pkg/transform/gorm/models/receipt.go
```diff
@@ -19,8 +19,9 @@ type MessageReceipt struct {
...
 type Receipt struct {
-	Index    int64 `gorm:"primaryKey"`
-	ExitCode int64
-	GasUsed  int64
-	Return   []byte
+	Index      int64 `gorm:"primaryKey"`
+	ExitCode   int64
+	GasUsed    int64
+	Return     []byte
+	EventsRoot string
```
frrist, line 26: nit: use `types.DbCid` here as done in the `Message` structure.

kasteph: thanks!

pkg/transform/gorm/models/chainmessage.go
kasteph: Oh that's a good idea. Transformations on the basic Filecoin types can always be done later.

pkg/transform/gorm/models/vmmessages.go
kasteph: Looking at gorm [Embedded Structs](https://gorm.io/docs/models.html#Embedded-Struct) has me thinking that it's better to have common non-null types between `Message` and `VmMessage` wrapped like:

```
type BaseMessage struct {
  To
  From
  Value
  Method
  Params
}
```

which can then be respectively added as an embedded struct for both `Message` and `VmMessage`. This way it'll be easier to see what they do have in common and what they don't. But this is also not a ⛰️ I'm gonna die on lol.

pkg/transform/gorm/models/receipt.go
```diff
@@ -19,8 +19,9 @@ type MessageReceipt struct {
...
 type Receipt struct {
-	Index    int64 `gorm:"primaryKey"`
-	ExitCode int64
-	GasUsed  int64
-	Return   []byte
+	Index      int64 `gorm:"primaryKey"`
+	ExitCode   int64
+	GasUsed    int64
+	Return     []byte
+	EventsRoot types.DbCid
```
kasteph, line 26: typo:

`types.DbCID` and not `types.DbCid`

PR #1114, opened 2023-01-17
Ntwk/hyperspacenet

frrist:
build this code with `make hyperspacenet`
Will merge when lotus cuts v1.20 release and we pull said deps in.

PR #1121, opened 2023-01-30
chore: Add minimal support for v10 and v11 evm actors

kasteph:
Lily is missing this builtin actor which, I think, is the reason why the walks from my comment regarding not being able to decode [`InvokeContract fil/10/evm`](https://github.com/filecoin-project/lily/pull/1114#issuecomment-1408107588).

This is a copy-pasta from Lotus.


frrist:
I don't think this is needed, see : https://github.com/filecoin-project/lily/pull/1114#issuecomment-1409431128

PR #1123, opened 2023-02-01
Add EventsRoot to VM Messages and Receipts

kasteph:
From [FIP-0049](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0049.md#new-vmemit_event-syscall), a new field, `EventsRoot` is added to receipts.

### DIFF
--- a/tasks/messageexecutions/vm/task.go
+++ b/tasks/messageexecutions/vm/task.go
@@ -135,7 +135,6 @@ func (t *Task) ProcessTipSets(ctx context.Context, current *types.TipSet, execut
 				Method:    uint64(child.Message.Method),
 				// Params will be filled below if exit code is non-zero
 				// Return will be filled below if exit code is non-zero
-				EventsRoot: child.Receipt.EventsRoot.String(),
 			}
 
 			// only parse params and return of successful messages since unsuccessful messages don't return a parseable value.
--- a/tasks/messageexecutions/vm/task.go
+++ b/tasks/messageexecutions/vm/task.go
@@ -135,6 +135,7 @@ func (t *Task) ProcessTipSets(ctx context.Context, current *types.TipSet, execut
 				Method:    uint64(child.Message.Method),
 				// Params will be filled below if exit code is non-zero
 				// Return will be filled below if exit code is non-zero
+				EventsRoot: child.Receipt.EventsRoot.String(),
 			}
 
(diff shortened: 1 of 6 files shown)


tasks/messages/receipt/task.go
```diff
@@ -81,10 +81,11 @@ func (t *Task) ProcessTipSets(ctx context.Context, current *types.TipSet, execut
...
 
-				Message:  msg.Cid().String(),
-				Idx:      index,
-				ExitCode: int64(rec.ExitCode),
-				GasUsed:  rec.GasUsed,
+				Message:    msg.Cid().String(),
+				Idx:        index,
+				ExitCode:   int64(rec.ExitCode),
+				GasUsed:    rec.GasUsed,
+				EventsRoot: rec.EventsRoot.String(),
```
frrist, line 88: I don't think we need to record the individual events for each VM message receipt, nor am I sure that VMMessage receipts even contain events. Let's drop this change since the on-chain receipt will already contain all of this information.

frrist (commented):
Thanks for making this, I'll address the comments I left.

frrist:
As a follow on to this work we should consider how we would represent the events contained in a message receipt event root as a table in the database. Knowing the CID of an Events Root is only useful if the content itself can be inspected.

frrist (approved):
Give me changes a look, and if they look okay please merge

tasks/messageexecutions/vm/task.go
```diff
@@ -135,6 +135,7 @@ func (t *Task) ProcessTipSets(ctx context.Context, current *types.TipSet, execut
 				Method:    uint64(child.Message.Method),
 				// Params will be filled below if exit code is non-zero
 				// Return will be filled below if exit code is non-zero
+				EventsRoot: child.Receipt.EventsRoot.String(),
```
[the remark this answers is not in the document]
kasteph: ah thanks, that makes sense

tasks/messages/receipt/task.go
kasteph: I asked on #fil-lotus-dev and the consensus seems to be that the VM Message receipt will have the event root https://filecoinproject.slack.com/archives/CP50PPW2X/p1675129261393459 (or are they talking about FVM?). prob stupid question but how does a regular message receipt contain info about a vm message receipt?

PR #1126, opened 2023-02-02
fix: Fix tipset cache bug with SetCurrent

FlattestWhite:
Addresses the bug in: https://github.com/filecoin-project/lily/issues/1083.


FlattestWhite:
cc @frrist

frrist:
I think the bug is with the `Warm` method rather than the `SetCurrent` implementation. Thanks for getting this started.

birdychang:
@frrist isn't this [PR](https://github.com/filecoin-project/lily/pull/1098/files) meant to fix the bug?

PR #1131, opened 2023-02-11
[command] add currentheight to job list

Terryhung:
### Description
The json returned by lily job list should include current height for walk, watch and gapfill jobs.
### Approach
Add the `currentHeight` in `Reporter`, then pass the `Reporter` to Watcher, Walker and Filler. Therefore, they can update the `currentHeight` during job execution. Finally, we could see the current height for walk, watch and gapfill jobs.

### DIFF
--- a/schedule/scheduler.go
+++ b/schedule/scheduler.go
@@ -84,10 +84,10 @@ type JobConfig struct {
 
 type Reporter struct {
 	// Current Height is the current height of the job
-	CurrentHeight int
+	CurrentHeight int64
 }
 
-func (r *Reporter) UpdateCurrentHeight(height int) {
+func (r *Reporter) UpdateCurrentHeight(height int64) {
 	r.CurrentHeight = height
 }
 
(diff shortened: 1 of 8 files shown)


schedule/scheduler.go
```diff
@@ -79,6 +82,19 @@ type JobConfig struct {
...
+type Reporter struct {
+	// Current Height is the current height of the job
+	CurrentHeight int64
+}
+
+func (r *Reporter) UpdateCurrentHeight(height int64) {
+	r.CurrentHeight = height
+}
+
+func (r *Reporter) GetStatus() Reporter {
```
frrist, line 94: Remove this method, I don't see the point in having it, do you?

schedule/scheduler.go
```diff
@@ -376,7 +394,11 @@ func (s *Scheduler) Jobs() []JobListResult {
 			Params:              j.Params,
 			StartedAt:           j.StartedAt,
 			EndedAt:             j.EndedAt,
-		})
+		}
+		if j.Reporter != nil {
+			result.Report = j.Reporter.GetStatus()
```
frrist, line 399: with `GetStatus` removed, we don't need this check.

frrist (approved):
Looks good once comments are addressed.

PR #1135, opened 2023-02-16
fix: move the models-* subcmd to a different cmd

### LINKED ISSUE #1134: `help` subcommand not working as expected
### Describe the bug:

When running `lily help daemon`:
```
No help topic for 'daemon'
```

Instead of:

```
NAME:
   lily daemon - Start a lily daemon process.

USAGE:
   lily daemon [command options] [arguments...]

DESCRIPTION:
   Starts lily in daemon mode with its own blockstore.
   In daemon mode lily synchronizes with the filecoin network and runs jobs such
   as walk and watch against its local blockstore. This gives better performance
   than operating against an external blockstore but requires more disk space and
   memory.
   
   Before starting the daemon for the first time the blockstore must be initialized
   and synchronized. Lily can use a copy of an already synchronized Lotus node
   repository or can initialize its own empty one and import a snapshot to catch
   up to the chain.
   
.....
```

This also happens for all other commands.

### Steps to Reproduce:

1. Build lily
2. `lily help daemon`

kasteph:
fixes #1134


frrist (approved):

commands/help.go
```diff
@@ -1,63 +1,12 @@
...
-		for field, comment := range modelFields {
-			t.AppendRow(table.Row{field, comment})
-			t.AppendSeparator()
-		}
-		fmt.Println(t.Render())
-		return nil
-	},
-}
-
 var HelpCmd = &cli.Command{
```
frrist, line 7: Just delete this, since it has no sub commands

kasteph: done! thanks!

PR #1137, opened 2023-02-17
fix: use builtin lotus method to get events

frrist:
- it handles the case of legacy events


lens/interface.go
```diff
@@ -48,6 +48,9 @@ type ChainAPI interface {
 
 	MessagesForTipSetBlocks(ctx context.Context, ts *types.TipSet) ([]*BlockMessages, error)
 	TipSetMessageReceipts(ctx context.Context, ts, pts *types.TipSet) ([]*BlockMessageReceipts, error)
+
+	// added during hyperspace
+	ChainGetEvents(ctx context.Context, root cid.Cid) ([]types.Event, error)
```
frrist, line 53: Provided by lotus here: https://github.com/filecoin-project/lotus/blob/v1.20.0-hyperspace-nv20/node/impl/full/chain.go#L662

kasteph: cool!

kasteph (approved):

PR #1141, opened 2023-02-22
feat: track verifreg actor claim HAMT changes

frrist:
- request made here: https://github.com/protocol/pldw/issues/177#issuecomment-1372920524

### DIFF
--- a/schemas/v1/17_verifreg_claims.go
+++ b/schemas/v1/17_verifreg_claims.go
@@ -7,8 +7,8 @@ func init() {
 	CREATE TABLE IF NOT EXISTS {{ .SchemaName | default "public"}}.verified_registry_claims  (
 	    height BIGINT NOT NULL,
 	    state_root TEXT NOT NULL,
-		provider TEXT NOT NULL,
 		claim_id BIGINT NOT NULL,
+		provider TEXT NOT NULL,
 		client TEXT NOT NULL,
 		data TEXT NOT NULL,
 		size BIGINT NOT NULL,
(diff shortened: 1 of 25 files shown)


schemas/v1/17_verifreg_claims.go
```diff
@@ -0,0 +1,25 @@
...
+		client TEXT NOT NULL,
+		data TEXT NOT NULL,
+		size BIGINT NOT NULL,
+		term_min BIGINT NOT NULL,
+		term_max BIGINT NOT NULL,
+		term_start BIGINT NOT NULL,
+		sector BIGINT NOT NULL,
+		event 		{{ .SchemaName | default "public"}}.verified_registry_event_type NOT NULL,
+		
+		PRIMARY KEY(height, state_root, provider, claim_id)
```
kasteph, line 21: Could a `claim_id` be the same for different `provider`s? If not, perhaps the `PRIMARY KEY` could be simplified to the following:

```
PRIMARY KEY(height, state_root, claim_id
```

to make it faster.

frrist: @anorth or @ZenGround0 as authors of [FIP-45](https://github.com/filecoin-project/FIPs/blob/3cd2e0f8157be2edf9061a1eda1f39c1b88ca903/FIPS/fip-0045.md) could you either of you comment on the uniqueness of ClaimIDs? Are they globally unique (like DealIDs) or unique per provider (like sectorIDs to miners)? The FIP doesn't seem to specify this, or I can't find where it does, at least.

frrist: @ZenGround0 and I looked at the in the office together, and the answer is: ClaimID is a globally unique ID meaning I should (and will) change the PK definition not to include the provider in the constraint.

PR #1143, opened 2023-02-23
fix: reduce time bucket size

birdychang:


metrics/metrics.go
```diff
@@ -203,7 +203,7 @@ var DefaultViews = []*view.View{
 	{
 		Measure:     PersistDuration,
 		Aggregation: defaultMillisecondsDistribution,
-		TagKeys:     []tag.Key{TaskType, Table, ActorCode},
+		TagKeys:     []tag.Key{TaskType, Table},
```
kasteph, line 206: Oh, why is ActorCode removed?

frrist: High cardinality, maybe?

metrics/metrics.go
```diff
@@ -9,7 +9,7 @@ import (
 	"go.opencensus.io/tag"
 )
 
-var defaultMillisecondsDistribution = view.Distribution(0.01, 0.05, 0.1, 0.3, 0.6, 0.8, 1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1000, 2000, 5000, 10000, 20000, 30000, 50000, 100000, 200000, 500000, 1000000, 2000000, 5000000, 10000000, 10000000)
```
frrist, line 12: @birdychang I assume this change is the primary motivation of the PR?

tasks/actorstate/task.go
```diff
@@ -151,8 +151,8 @@ func (t *Task) startActorStateExtraction(ctx context.Context, current, executed
 							Head:    ac.Actor.Head,
 							Address: addr,
 						}
-						stop := metrics.Timer(ctx, metrics.StateExtractionDuration)
-						defer stop()
+						// stop := metrics.Timer(ctx, metrics.StateExtractionDuration)
+						// defer stop()
```
frrist, line 155: What is the motivation for this the removal of this? If we _do_ mean to remove it, let's opt to delete the lines instead of commenting it out. And if this is the only location this metric is instrumented, let's delete the metric too.

metrics/metrics.go
birdychang: yes

metrics/metrics.go
birdychang: When I check the metric and code, `ActorCode` was never included. So I'm removing it from TagKeys.

tasks/actorstate/task.go
birdychang: I think we figure out a better way is to drop unwanted metrics in prometheus. I'll revert this.

metrics/metrics.go
frrist: Ohh its probably because we forget to include it when logging the metric...

frrist (approved):

metrics/metrics.go
birdychang: I see. I think the cardinality is already high enough :P

PR #1149, opened 2023-03-01
Migrate vm_messages to hypertable

birdychang:


birdychang:
Tested in staging db. Took 4 hr to migrate a 50GB table.

frrist:
Are stagings resource significantly less than productions?

birdychang:
> 

16 cores v.s. 24 cores, and 122GB v.s. 192GB.

schemas/v1/13_vm_messages_index.go
```diff
@@ -5,8 +5,9 @@ func init() {
 		13,
 		`
 ALTER TABLE {{ .SchemaName | default "public"}}.vm_messages
-    ADD COLUMN "index" BIGINT NOT NULL,
-    ADD CONSTRAINT vm_messages_uniq_index UNIQUE (height, state_root, cid, source, index);
+    ADD COLUMN "index" BIGINT NOT NULL DEFAULT -1;
+
+ALTER TABLE {{ .SchemaName | default "public"}}.vm_messages ALTER COLUMN "index" DROP DEFAULT;
```
kasteph, line 10: Maybe add a comment in the migration that here you are dropping the default value so that it can be dropped right away for existing rows.

birdychang: will do.

schemas/v1/14_vm_messages_migration.go
```diff
@@ -0,0 +1,21 @@
+package v1
+
+func init() {
+	patches.Register(
+		14,
+		`
+{{- if and .SchemaName (ne .SchemaName "public") }}
```
frrist, line 7: I am surprised this is required.

schemas/v1/14_vm_messages_migration.go
```diff
@@ -0,0 +1,21 @@
...
+SET search_path TO {{ .SchemaName }},public;
+{{- end }}
+
+SELECT create_hypertable(
+	'vm_messages',
+	'height',
+	chunk_time_interval => 2880,
```
frrist, line 14: What is the significance of 2880? The number of epochs in a day?

birdychang: Yes, number of epochs in a day. So with the retention policy, we can drop a chunk everyday.

schemas/v1/14_vm_messages_migration.go
frrist: ahh right, since we are using the `create_hypertable` method I think this is required.

schemas/v1/14_vm_messages_migration.go
frrist: can you add a comment stating this please?

frrist (approved):
Looks good once the last review comment is addressed. Please target this change at the `frrist/nv18-rc1` branch

PR #1161, opened 2023-03-15
fix: Add function for cleaning the Unicode escape sequence

Terryhung:
It will cause the error in PostgreSQL: `unsupported Unicode escape sequence`

Therefore, we have to fix it.


Terryhung:
Maybe we can just store the empty string without storing the byte code which is unusable/unreadable?

PR #1162, opened 2023-03-15
fix: prevent unsuccessful task from retrying

frrist:


birdychang (approved):

frrist:
For now, I think this is an acceptable fix as a backstop, but we'll want to address https://github.com/filecoin-project/lily/issues/1163 soon

PR #1169, opened 2023-03-21
fix: log unknown actor type without returning error

Terryhung:
Don't error on unknown method numbers for FEVM.

Due to the [previous fix](https://github.com/filecoin-project/lily/pull/1165/files), it will return the `Unknown` method, then cause the error in `parsed_message`.  Therefore, we can use the same logic as `vm_message`: logging the error without returning error.

### DIFF
--- a/lens/util/repo.go
+++ b/lens/util/repo.go
@@ -185,7 +185,8 @@ func MethodAndParamsForMessage(m *types.Message, destCode cid.Cid) (string, stri
 
 	params, method, err := ParseParams(m.Params, m.Method, destCode)
 	if method == "Unknown" {
-		return "", "", fmt.Errorf("unknown method for actor type %s: %d", destCode.String(), int64(m.Method))
+		log.Errorf("unknown method for actor type %s: %d", destCode.String(), int64(m.Method))
+		return "", "", nil
 	}
 	if err != nil {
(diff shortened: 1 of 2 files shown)


lens/util/repo.go
```diff
@@ -185,7 +185,8 @@ func MethodAndParamsForMessage(m *types.Message, destCode cid.Cid) (string, stri
 
 	params, method, err := ParseParams(m.Params, m.Method, destCode)
 	if method == "Unknown" {
-		return "", "", fmt.Errorf("unknown method for actor type %s: %d", destCode.String(), int64(m.Method))
```
frrist, line 188: I'm not sure this is the correct behavior if a method is unknown. I don't think we can write empty strings to models that expect a method name. One option would be to remove returning the name of the method being called here, and instead post process tables needing a method name string. cc @birdychang

birdychang: Why not just put ‘unknown’ as the method name?

birdychang: We can always post process later. Also, don’t we wanna deprecate parsed messages? I don’t really care about the model. Let’s make it work first.

birdychang: Or just store the method code as the method name here. There’s should be separate columns in case it fails to lookup the name.

Terryhung: Store the method in string format.

birdychang (approved):

PR #1174, opened 2023-04-06
feat: Add the state and code_cid for actors table

Terryhung:
## Description

Based on the information provided, it appears that the `actor_states` table is being deprecated because the data it contains is mostly covered by the `actors` table. However, there are two columns: `state` and` code_cid`, which are present in `actor_states` but missing in `actors`. To address this, these two columns will be added to the `actors` table, thereby making the `actor_states` table redundant. Once this is done, the `actor_states` table can be safely deprecated.

### DIFF
--- a/schemas/v1/20_actor_add_state.go
+++ b/schemas/v1/20_actor_add_state.go
@@ -7,7 +7,11 @@ func init() {
 ALTER TABLE {{ .SchemaName | default "public"}}.actors
     ADD COLUMN IF NOT EXISTS "state" jsonb;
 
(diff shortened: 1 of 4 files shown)


birdychang (approved):

schemas/v1/20_actor_add_state.go
```diff
@@ -0,0 +1,17 @@
+package v1
+
+func init() {
+	patches.Register(
+		20,
```
kasteph, line 5:
```suggestion
		21,
```

Don't forget to rename the file as well from `20_actor_add_state.go` to `21_actor_add_state.go`

PR #1192, opened 2023-05-03
fix: batch insert actor & actor_state data

### LINKED ISSUE #1190: Support Batch Insert in Lily Tasks
### Description

After reviewing the performance of Lily in Jaeger, we have discovered that the persistence of the actor and actor state is taking between 14 to 16 seconds.

image

The reason for this delay is that only a few records are being inserted for each transaction. Therefore, we could improve the performance by attempting batch inserts within a single transaction.

### Acceptance criteria

Implementing measures to speed up the persistence performance in actor and actor_state.

### Where to begin

https://github.com/filecoin-project/lily/blob/master/model/interface.go#L23
https://github.com/filecoin-project/lily/blob/master/model/actors/common/actors.go#L52

birdychang:
Currently, actor state changes are stored as `model.PersistableList`, which doesn't support batch persistence. This PR adds  a transformer to actorstate task to convert `model.PersistableList` to corresponding model type that supports batch persistence.
This resolves #1190.


kasteph (approved):
looks very nice, thanks!

Issue #1191, opened 2023-04-29
After each upgrade, there are always one or two heights where the CPU is full and the task cannot run


LINJINTIANDE:
### Describe the bug:
After each upgrade, there is always one or two heights where the cpu is full and the task cannot run, for example, the height is 2809803
### Steps to Reproduce:
lily job run  --storage=lily --tasks=block_header,block_parent,drand_block_entrie,id_addresses,actor,actor_state,internal_messages,internal_parsed_messages,message,block_message,receipt,message_gas_economy,derived_gas_outputs,parsed_message,multisig_approvals,multisig_transaction,miner_current_deadline_info,miner_fee_debt,miner_info,miner_locked_fund,miner_pre_commit_info,miner_sector_deal,miner_sector_event,miner_sector_infos_v7 walk --from=2809803  --to=2809803
![image](https://user-images.githubusercontent.com/42436479/235301841-b4aab1ca-4537-4b33-b787-a429aaab30b6.png)
### machine configuration
CPU Authentic AMD7302
Memory 1t
GPU2080
4t ssd

Terryhung:
Hi @LINJINTIANDE,

Based on our analysis, we have identified that the root cause of the issue is the large number of actor changes that need to be handled during this epoch. Specifically, as part of the lotus [migration](https://github.com/filecoin-project/go-state-types/blob/1e6cf0d47cdda75383ef036fc2725d1cf51dbde8/builtin/v11/migration/top.go#L19), a total of 2,146,633 actors were modified, leading to a high demand for CPU resources for tasks related to actor changes.

**We recommend considering the option to skip this epoch given the unusually high number of actor changes.** 

Typically, in a normal epoch, the number of actor changes does not exceed 1000. To address this issue, we propose adding a threshold for tasks in the watch job. Specifically, if the number of actor changes exceeds 10,000, we should skip the task to prevent it from consuming the entire CPU and allow it to fail gracefully. Additionally, we plan to support the adjustment of this threshold within the walk job.

Thanks!

LINJINTIANDE:
Thanks for your answer. 🙏

PR #1195, opened 2023-05-08
feat: add filter for actor changes

Terryhung:
This feature can help prevent this [issue](https://github.com/filecoin-project/lily/issues/1191) from occurring again.

By modifying the environment variable: `LILY_ACTOR_CHANGE_LIMITER`, it is able to adjust the threshold.


tasks/actorstate/task.go
```diff
@@ -41,6 +43,25 @@ func NewTaskWithTransformer(node tasks.DataSource, extractorMap ActorExtractorMa
 	}
 }
 
+const actorChangeLimiterEnv = "LILY_ACTOR_CHANGE_LIMITER"
+
+const defaultActorChangeLimit = 10000
+
+// Skip task processing when actor state changes abnormally.
+func (t *Task) exceedActorChangeLimit(changes map[address.Address]tasks.ActorStateChange) bool {
```
birdychang, line 51: nit: `exceedsActorChangeLimit`

birdychang (approved):

tasks/actorstate/task.go
```diff
@@ -41,6 +43,25 @@ func NewTaskWithTransformer(node tasks.DataSource, extractorMap ActorExtractorMa
...
+	// The 10,000 is a default threshold
+	limit := defaultActorChangeLimit
+
+	// Access the custom threshold set in the environment variable
+	s := os.Getenv(actorChangeLimiterEnv)
```
kasteph, line 56:
```suggestion
	s := os.Getenv(actorChangeLimiterEnv)
```

[os.LookupEnv](https://pkg.go.dev/os#LookupEnv) might be nicer.

.gitignore
```diff
@@ -21,3 +21,5 @@ visor
 build/.*
 support/tools/bin
 
+# Deployment
+deployment/.env
```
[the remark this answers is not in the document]
Terryhung: We will put the `.env` file in the deployment folder because of docker-compose usage, therefore we should add the `gitignore` here  😄 .

tasks/actorstate/task.go
Terryhung: nice

.gitignore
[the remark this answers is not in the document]
Terryhung: I change it to `*.env`.

tasks/actorstate/task.go
Terryhung: done

.gitignore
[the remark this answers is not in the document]
kasteph: Simply doing `.env` would ignore it in all paths instead of `*.env` which would only ignore in the root path. If we do hyperspecific paths like so, we could face the same problem again where the secrets are revealed.

.gitignore
```diff
@@ -21,3 +21,5 @@ visor
 build/.*
 support/tools/bin
 
+# Deployment
+*.env
```
kasteph, line 25: `*.env` will not match `.env` files in subfolders. `.env` would match all `.env` files in all paths in the git project.

See: https://www.atlassian.com/git/tutorials/saving-changes/gitignore

kasteph (changes requested):
`*.env` will not match paths that

.gitignore
Terryhung: I tested it in local, moreover, the the doc you provided, I saw it will match the file in subfolders.
截圖 2023-05-11 下午3 11 10

birdychang: guess we do want to filter out files like `abc.env` in the future.

birdychang (approved):

kasteph (approved):

PR #1205, opened 2023-05-16
fix: CodeCID should be at end of model

kasteph:
CSVs produced follow the order of the definition in the model. So CSVs of `actors` produced with this model definition are ordered like so:

```
height,id,state_root,code,code_cid,head,balance,nonce,state
```

This is problematic when inserting CSVs to other data sources as one then has to create these two columns before the change is introduced and fill these `null` values manually (really only a problem for `code_cid` which is in the middle of the table).

Note that Postgres doesn't have this issue (new columns are appended at the end).


birdychang (approved):
Thanks!

PR #1208, opened 2023-05-19
feat: add new task for fevm receipt

Terryhung:


tasks/fevm/receipt/tasks.go
```diff
@@ -0,0 +1,141 @@
...
+	return &Task{
+		node: node,
+	}
+}
+
+func (p *Task) isSentToEVMAddress(ctx context.Context, addr address.Address, tsk types.TipSetKey) bool {
+	act, err := p.node.Actor(ctx, addr, tsk)
+	if err != nil {
+		// If actor not found, then this is a placeholder address
+		if addr.String()[:5] == "f410f" {
```
birdychang, line 43: `addr.Protocol() == addr.Delegated` better?

tasks/fevm/receipt/tasks.go
```diff
@@ -0,0 +1,141 @@
...
+func NewTask(node tasks.DataSource) *Task {
+	return &Task{
+		node: node,
+	}
+}
+
+func (p *Task) isSentToEVMAddress(ctx context.Context, addr address.Address, tsk types.TipSetKey) bool {
+	act, err := p.node.Actor(ctx, addr, tsk)
+	if err != nil {
+		// If actor not found, then this is a placeholder address
```
birdychang, line 42: nit: If actor not found, check if it's a placeholder address.

tasks/fevm/receipt/tasks.go
```diff
@@ -0,0 +1,141 @@
...
+		if addr.String()[:5] == "f410f" {
+			log.Infof("Sent to Placeholder address: %v", addr)
```
birdychang, line 44: Debugf?

chain/datasource/datasource.go
```diff
@@ -40,51 +40,32 @@ var (
...
 	}
+	return defaultValue
+}
 
+func init() {
+	tipsetMessageReceiptCacheSize = getCacheSizeFromEnv(tipsetMessageReceiptSizeEnv, 4)
+	executedTsCacheSize = getCacheSizeFromEnv(executedTsCacheSizeEnv, 4)
+	diffPreCommitCacheSize = getCacheSizeFromEnv(diffPreCommitCacheSizeEnv, 500)
+	diffSectorCacheSize = getCacheSizeFromEnv(diffSectorCacheSizeEnv, 500)
+	actorCacheSize = getCacheSizeFromEnv(actorCacheSizeEnv, 1000)
```
birdychang, line 68: Looks much cleaner!

chain/datasource/datasource.go
```diff
@@ -192,13 +188,30 @@ func (t *DataSource) Store() adt.Store {
...
 	if span.IsRecording() {
 		span.SetAttributes(attribute.String("tipset", tsk.String()))
 		span.SetAttributes(attribute.String("address", addr.String()))
 	}
 	defer span.End()
-	return t.node.StateGetActor(ctx, addr, tsk)
+
+	key, err := asKey(addr, tsk)
+	if err != nil {
+		return nil, err
```
birdychang, line 201: not sure when this would happen, but shouldn't we still try to continue and call `t.node.StateGetActor` instead of returning error?
```
if err == nil {
  value, found := t.actorCache.Get(key)
  ...
}
```

tasks/fevm/receipt/tasks.go
Terryhung: nice 👍

tasks/fevm/receipt/tasks.go
Terryhung: changed

chain/datasource/datasource.go
Terryhung: fixed!

birdychang (approved):
Thanks!

PR #1213, opened 2023-05-26
feat: add new task: fevm_transaction

Terryhung:


schemas/v1/24_fevm_transaction.go
```diff
@@ -0,0 +1,31 @@
...
+		chain_id             BIGINT,
+		nonce                BIGINT,
+		type                 BIGINT,
+		"from"               TEXT,
+		"to"                 TEXT,
+		input                TEXT,
+		value                TEXT,
+		gas                  BIGINT,
+		max_fee_per_gas      TEXT,
+		max_priority_fee_per_gas TEXT,
```
birdychang, line 22: should we do `numeric`?

birdychang (approved):

schemas/v1/24_fevm_transaction.go
Terryhung: change to numeric

Issue #1172, opened 2023-03-23
new task: contracts


kasteph:
### Description

We need a new task for contracts because:

- it's useful for FEVM analysis
- a lot of questions/insights being asked regarding contracts since FEVM launch
- using the current schema as is (joining on actors and actor_states using the head column) is painfully long

### Acceptance criteria

- a new task that extracts contract data (BytecodeHash), the associated actor id, and bytecode size
- a new schema for this model

### Where to begin

- check actor state task and actors task to get bytecodehash (in state column) and id respectively
- [Forrest's comment](https://github.com/protocol/pldw/issues/214#issuecomment-1476783490) on obtaining bytecode size:
<details>
<summary> Forrest's comment </summary>
As an example, an actor state in the DB may resemble the following:
```
[
  {
    "height": 2701059,
    "head": "bafy2bzacedme3aewyducwobpy3lyru3wgtvwpzb6qkqhficehlcb3srczr5pe",
    "code": "bafk2bzaceahmzdxhqsm7cu2mexusjp6frm7r4kdesvti3etv5evfqboos2j4g",
    "state": {
      "Nonce": 2,
      "Bytecode": {
        "/": "bafk2bzacebfu7qcpzlvligpsiv5nyw4bvkkkllxvw3f6d3iqafzyz4zdhrnaw"
      },
      "Tombstone": null,
      "BytecodeHash": [
        149,
        202,
        166,
        1,
        240,
        103,
        238,
        235,
        180,
        119,
        219,
        207,
        246,
        253,
        252,
        218,
        135,
        201,
        252,
        83,
        139,
        175,
        103,
        184,
        247,
        232,
        128,
        204,
        177,
        204,
        240,
        219
      ],
      "ContractState": {
        "/": "bafy2bzacecz5gkgmgyfb6qeip32xn243qtfbpzi4jy2j7xxmc6j6s6xlqric4"
      }
    }
  }
]
```
I don't know of a way to obtain the bytecode size based on the hash alone. It is worth noting that in this specific example, the byte code is retrievable directly from IPFS: https://ipfs.io/ipfs/bafk2bzacebfu7qcpzlvligpsiv5nyw4bvkkkllxvw3f6d3iqafzyz4zdhrnaw (7.3KB)

I think if our goals are to obtain information about:
- EVM bytecode-size distribution.
- EVM state-size histogram (size of the EVM actor's state-tree minus the size of the bytecode is likely sufficient).

[...]

PR #1214, opened 2023-05-30
feat: add new task fevm contract

Terryhung:
This PR is able to resolve the https://github.com/filecoin-project/lily/issues/1172.

### DIFF
--- a/model/fevm/contract.go
+++ b/model/fevm/contract.go
@@ -1,0 +1,46 @@
+package fevm
+
+import (
+	"context"
+
+	"go.opencensus.io/tag"
+
+	"github.com/filecoin-project/lily/metrics"
+	"github.com/filecoin-project/lily/model"
+)
+
+type FEVMContract struct {
+	tableName struct{} `pg:"fevm_contract"` // nolint: structcheck
+
+	// Height message was executed at.
+	Height int64 `pg:",pk,notnull,use_zero"`
+
+	// Actor address.
+	ActorID string `pg:",notnull"`
+
+	// Actor Address in ETH
+	EthAddress string `pg:",notnull"`
+
+	ByteCode     string `pg:",notnull"`
+	ByteCodeHash string `pg:",notnull"`
+
+	Balance string `pg:"type:numeric,notnull"`
+	Nonce   uint64 `pg:",use_zero"`
+}
+
+func (f *FEVMContract) Persist(ctx context.Context, s model.StorageBatch, version model.Version) error {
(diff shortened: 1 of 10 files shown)


model/fevm/contract.go
```diff
@@ -0,0 +1,46 @@
...
+	"context"
+
+	"go.opencensus.io/tag"
+
+	"github.com/filecoin-project/lily/metrics"
+	"github.com/filecoin-project/lily/model"
+)
+
+type FEVMContract struct {
+	tableName struct{} `pg:"fevm_contract"` // nolint: structcheck
```
kasteph, line 13:
```suggestion
	tableName struct{} `pg:"fevm_contracts"` // nolint: structcheck
```

We should make this consistent as with other models in Lily. They're all plural and it makes sense as in a table you don't have just a single contract but multiple.

schemas/v1/25_fevm_contract.go
```diff
@@ -0,0 +1,19 @@
...
+	patches.Register(
+		25,
+		`
+	CREATE TABLE IF NOT EXISTS {{ .SchemaName | default "public"}}.fevm_contract  (
+	    height               BIGINT NOT NULL,
+		actor_id             TEXT,
+		eth_address          TEXT,
+		byte_code            TEXT,
+		byte_code_hash       TEXT,
+		balance              numeric,
```
kasteph, line 13: nit:

```suggestion
		balance              NUMERIC,
```

schemas/v1/25_fevm_contract.go
```diff
@@ -0,0 +1,19 @@
...
+		nonce                BIGINT,
+		PRIMARY KEY(height, actor_id)
```
kasteph, line 15: Did Starboard also have the PK on the `height` and `actor_id`? We might also want to add another column to that, either the `eth_address` or `nonce` as an actor can deploy many contracts.

model/fevm/contract.go
Terryhung: changed

tasks/fevm/contract/task.go
```diff
@@ -0,0 +1,118 @@
...
+
+	out := make(fevm.FEVMContractList, 0)
+	errs := []error{}
+	for _, change := range actorChanges {
+		actor := change.Actor
+		if actor.Address == nil {
+			continue
+		}
+
+		if !util.IsSentToEVMAddress(ctx, p.node, *actor.Address, executed.Key()) {
```
birdychang, line 67: should be called `IsEVMAddress`

birdychang (approved):

PR #1219, opened 2023-06-05
fix: keep running the whole tipset task without cancel

Terryhung:


chain/indexer/integrated/tipset/tipset.go
```diff
@@ -145,7 +145,6 @@ func (ti *TipSetIndexer) TipSet(ctx context.Context, ts *types.TipSet) (chan *Re
 				if res.Error != nil {
 					llt.Errorw("task returned with error", "error", res.Error.Error())
 					errCh <- res.Error
-					return
```
birdychang, line 148: if we continue to execute, will `ErrorsDetected` be populated so the task report will show the reason?

Terryhung: yes, it would be populated

birdychang (approved):

PR #1222, opened 2023-06-06
fix: run more backfill tipsetkey during import chain

Terryhung:


commands/init.go
```diff
@@ -45,6 +46,13 @@ var InitCmd = &cli.Command{
 			EnvVars:     []string{"LILY_SNAPSHOT"},
 			Destination: &initFlags.importSnapshot,
 		},
+		&cli.IntFlag{
+			Name:        "backfill-tipsetkey-range",
+			Usage:       "Determine the extent of backfilling from the head.",
+			EnvVars:     []string{"LILY_BACKFILL_TIPSETKEY_RANGE"},
+			Value:       1800,
```
birdychang, line 53: our snapshot range is 2890 epochs. Should we make the default something like 3000 or 3600 to ensure we always backfill more than a day?

commands/util/import.go
```diff
@@ -189,6 +190,11 @@ func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool)
 		return fmt.Errorf("importing chain failed: %w", err)
 	}
 
+	err = backfillMoreEpochsForTipsetKey(ctx, ts, cst, backfillTipsetkeyRange)
```
birdychang, line 193: Add comment to explain `cst.Import()` only backfills 1800 epochs.
Make the name shorter `backfillTipsetKey` .

birdychang (approved):

commands/init.go
Terryhung: changed

commands/util/import.go
Terryhung: added more comment!

commands/util/import.go
```diff
@@ -189,6 +190,14 @@ func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool)
 		return fmt.Errorf("importing chain failed: %w", err)
 	}
 
+	// The cst.Import function will only backfill 1800 epochs of tipsetkey,
+	// as determined by the value specified in TipsetkeyBackfillRange.
```
birdychang, line 194: this line might be confusing since we have `backfillTipsetKeyRange`. No need to mention the variable in `cst.import`.

birdychang (approved):

commands/util/import.go
Terryhung: removed

PR #1236, opened 2023-06-23
fix: var names for chain commands

### LINKED ISSUE #1224: These two variable names in the source code need to be swapped. ChainStateComputeRange and ChainStateCompute.
### Describe the bug:
I found a problem with source code variable names.
[file addr](https://github.com/filecoin-project/lily/blob/master/commands/chain.go)
```go
// line 324
var ChainStateComputeRange = &cli.Command{
	Name:  "state-compute",
	Usage: "Generates the state at epoch `N`",
        // ...
}

// line 357
var ChainStateCompute = &cli.Command{
	Name:  "state-compute-range",
	Usage: "Generates the state from epoch `FROM` to epoch `TO`",
        // ...
}
```
The two declared variables do not match what they were initialized to, the names of the two variables should be swapped.
### Steps to Reproduce:
```go
// line 324
var ChainStateComputeRange = &cli.Command{
	Name:  "state-compute",
	Usage: "Generates the state at epoch `N`",
        // ...
}

// line 357
var ChainStateCompute = &cli.Command{
	Name:  "state-compute-range",
	Usage: "Generates the state from epoch `FROM` to epoch `TO`",
        // ...
}
```
### Lily Version: vx.x.x
v0.16.0

kasteph:
closes #1224

### DIFF
--- a/commands/chain.go
+++ b/commands/chain.go
@@ -321,7 +321,7 @@ var ChainStateInspect = &cli.Command{
 	},
 }
 
-var ChainStateComputeRange = &cli.Command{
+var ChainStateCompute = &cli.Command{
 	Name:  "state-compute",
 	Usage: "Generates the state at epoch `N`",
 	Flags: []cli.Flag{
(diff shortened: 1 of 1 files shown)


birdychang (approved):
Thanks for the fix!

PR #1240, opened 2023-07-01
feat: add human readable actor codes in fevm_traces

kasteph:
This feature adds human-readable actor codes for sender and receiver to fulfill an offline request via Slack from @ychiaoli18.

This should make queries necessary for the FEVM analyses faster as it will no longer require merges with the actors table.

### DIFF
--- a/model/fevm/trace.go
+++ b/model/fevm/trace.go
@@ -57,9 +57,9 @@ type FEVMTrace struct {
 	// Returns codec.
 	ReturnsCodec uint64 `pg:",notnull,use_zero"`
 	// Human-readable identifier of receiver (To).
-	ToActorCode string `pg:",notnull"`
+	ToActorName string `pg:",notnull"`
 	// Human-readable identifier of sender (From).
-	FromActorCode string `pg:",notnull"`
+	FromActorName string `pg:",notnull"`
 }
 
 func (f *FEVMTrace) Persist(ctx context.Context, s model.StorageBatch, version model.Version) error {
--- a/model/fevm/trace.go
+++ b/model/fevm/trace.go
@@ -38,7 +38,7 @@ type FEVMTrace struct {
 	Method uint64 `pg:",notnull,use_zero"`
 	// Method in readable name.
 	ParsedMethod string `pg:",notnull"`
-	// ActorCode of To (receiver).
+	// ActorCode of To (receiver) as a CID.
 	ActorCode string `pg:",notnull"`
 	// ExitCode of message execution.
 	ExitCode int64 `pg:",notnull,use_zero"`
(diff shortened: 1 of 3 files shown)


Terryhung (approved):

model/fevm/trace.go
```diff
@@ -56,6 +56,10 @@ type FEVMTrace struct {
 	ParamsCodec uint64 `pg:",notnull,use_zero"`
 	// Returns codec.
 	ReturnsCodec uint64 `pg:",notnull,use_zero"`
+	// Human-readable identifier of receiver (To).
+	ToActorCode string `pg:",notnull"`
+	// Human-readable identifier of sender (From).
+	FromActorCode string `pg:",notnull"`
```
birdychang, line 62: We should call them `ToActorName` and `FromActorName`.

schemas/v1/28_add_actor_names_to_fevm_traces.go
```diff
@@ -0,0 +1,17 @@
...
+
+func init() {
+	patches.Register(
+		28,
+		`
+	ALTER TABLE {{ .SchemaName | default "public"}}.fevm_traces
+		ADD COLUMN IF NOT EXISTS "to_actor_code" text NOT NULL;
+	
+	ALTER TABLE {{ .SchemaName | default "public"}}.fevm_traces
+		ADD COLUMN IF NOT EXISTS "from_actor_code" text NOT NULL;
```
birdychang, line 11: update

tasks/fevm/trace/task.go
```diff
@@ -139,13 +152,15 @@ func (t *Task) ProcessTipSets(ctx context.Context, current *types.TipSet, execut
...
-				ActorCode:           toActorCode,
+				ActorCode:           actorCode,
 				Method:              uint64(child.Message.Method),
 				Index:               child.Index,
 				Params:              ethtypes.EthBytes(child.Message.Params).String(),
 				Returns:             ethtypes.EthBytes(child.Receipt.Return).String(),
 				ParamsCodec:         child.Message.ParamsCodec,
 				ReturnsCodec:        child.Receipt.ReturnCodec,
+				ToActorCode:         toActorCode,
+				FromActorCode:       fromActorCode,
```
birdychang, line 163: update

birdychang (changes requested):

model/fevm/trace.go
kasteph: sounds good, we should probably do it in other tables too then where `Code` is actually human readable name, [such as in this table in the schema](https://github.com/filecoin-project/lily/blob/master/schemas/v1/base.go#L118), to be consistent. How should we reconcile this? Should we rename the `actors` table to use `ActorName` too?

birdychang: Yes, we should. Btw, should be call it `actorName` or `actorType`?

kasteph: I think `ActorName` is good

birdychang (approved):
Thanks.

PR #1242, opened 2023-07-06
fix: add Transform implementation for PreCommitInfoExtractorV8

kasteph:
For earlier parts of the chain, i.e. when miner actors were v8 and below, we get the following error:

```
{
    "level": "warn",
    "ts": "2023-07-06T00:03:04.875Z",
    "logger": "lily/tasks",
    "caller": "actorstate/task.go:140",
    "msg": "failed to transform with error: ",
    "error": "expected MinerPreCommitInfoV9List type but got: miner.MinerPreCommitInfoList"
}
```

### DIFF
--- a/chain/indexer/integrated/processor/state.go
+++ b/chain/indexer/integrated/processor/state.go
@@ -449,19 +449,28 @@ func MakeProcessors(api tasks.DataSource, indexerTasks []string) (*IndexerProces
 				),
 				minertask.LockedFundsExtractor{},
 			)
+		case tasktype.MinerPreCommitInfoV1_8:
+			out.ActorProcessors[t] = actorstate.NewTaskWithTransformer(
+				api,
+				actorstate.NewCustomTypedActorExtractorMap(
+					map[cid.Cid][]actorstate.ActorStateExtractor{
+						mineractors.VersionCodes()[actorstypes.Version0]: {minertask.PreCommitInfoExtractorV8{}},
+						mineractors.VersionCodes()[actorstypes.Version2]: {minertask.PreCommitInfoExtractorV8{}},
+						mineractors.VersionCodes()[actorstypes.Version3]: {minertask.PreCommitInfoExtractorV8{}},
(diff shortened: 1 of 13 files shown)


Terryhung (approved):
nice!

birdychang:
Thanks for the fix!

storage/sql.go
```diff
@@ -352,10 +353,13 @@ func verifyModel(ctx context.Context, db *pg.DB, schemaName string, m *orm.Table
...
-		if datatype == "timestamp with time zone" {
+		switch datatype {
+		case "timestamp with time zone":
+			fallthrough
+		case "timestamp without time zone":
 			datatype = "timestamptz"
-		} else if datatype == "timestamp without time zone" {
-			datatype = "timestamp"
+		case "ARRAY":
+			datatype = "bigint[]"
```
birdychang, line 362: why is this needed?

chain/indexer/integrated/processor/state.go
```diff
@@ -449,19 +449,28 @@ func MakeProcessors(api tasks.DataSource, indexerTasks []string) (*IndexerProces
 				),
 				minertask.LockedFundsExtractor{},
 			)
+		case tasktype.MinerPreCommitInfoV1_8:
```
birdychang, line 452: Thank you! This is much better.

birdychang (approved):

storage/sql.go
kasteph: Because of this: https://github.com/filecoin-project/lily/blob/master/storage/sql_test.go#L49 -- it's a convenience function being used in such a test.

Moreover, the mapping of data types between Postgres and Golang is not one-to-one [that's why the original code has this condition as well](https://github.com/filecoin-project/lily/blob/master/storage/sql.go#L355-L359) because it tries to equate a Postgres datatype with a `fld.SQLType` [here](https://github.com/filecoin-project/lily/blob/master/storage/sql.go#L361-L363)

PR #1264, opened 2023-09-22
feat: add new actor dump: miner

Terryhung:
We have implemented a miner actor dump task, which allows us to comprehensively inspect the current status of miners and easily analyze the historical behavior of a specific miner.

Use the `StateLookupRobustAddress` to get the miner address.

### DIFF
--- a/lens/interface.go
+++ b/lens/interface.go
@@ -58,6 +58,7 @@ type ChainAPI interface {
 
 type StateAPI interface {
 	StateGetActor(ctx context.Context, addr address.Address, tsk types.TipSetKey) (*types.Actor, error)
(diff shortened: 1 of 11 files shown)


lens/interface.go
```diff
@@ -58,6 +58,7 @@ type ChainAPI interface {
 
 type StateAPI interface {
 	StateGetActor(ctx context.Context, addr address.Address, tsk types.TipSetKey) (*types.Actor, error)
+	StateLookupRobustAddress(ctx context.Context, addr address.Address, tsk types.TipSetKey) (address.Address, error)
```
Terryhung, line 61: This Lotus API will use the `ForEach` to iterate address map. Maybe we could implement a new map: key is actor_id, value is robust address.

birdychang (approved):

PR #1267, opened 2023-10-04
fix: use exponential backoff to attempt reconnect to tsdb

### LINKED ISSUE #1211: Reestablish the connection to PostgreSQL after encountering an error during connection
### Description
In Lily's job, if an error occurs in the PostgreSQL connection, the job will halt and stop running. We should enable Lily to have the capability to reestablish the connection in case of an error.

### Acceptance criteria

Reconnect to PostgreSQL after terminating the connection between Lily and PostgreSQL.

kasteph:
closes #1211


Terryhung (approved):
Nice

PR #1338, opened 2025-03-21
chore: update lotus to v1.32.0-rc3

Terryhung:
```
./lily migrate --db="postgres://postgres:postgres@localhost:5432/lily_test?sslmode=disable" --name migration --schema public --to 1.43
```


tasks/periodic_actor_dump/miner_actor/tasks.go
```diff
@@ -228,15 +228,15 @@ func calculateTerminateFeeForSector(currentEpoch abi.ChainEpoch, sector *miner.S
...
 	// relevant_replaced_age = min(PowerBaseEpoch - Activation, lifetimeCap - cappedSectorAge)
 	relevantReplacedAge := big.NewInt(int64(min(sector.PowerBaseEpoch-sector.Activation, lifetimeCap-cappedSectorAge)))
 	// Add the replaced sector's contribution.
-	expectedReward = big.Add(expectedReward, big.Mul(sector.ReplacedDayReward, relevantReplacedAge))
+	expectedReward = big.Add(expectedReward, big.Mul(*sector.ReplacedDayReward, relevantReplacedAge))
 	// penalized_reward = expected_reward / TERMINATION_REWARD_FACTOR_DENOM
 	penalizedReward := big.Div(expectedReward, big.NewInt(int64(TERMINATION_REWARD_FACTOR_DENOM)))
 	// Termination fee = ExpectedStoragePledge + (penalized_reward / EPOCHS_IN_DAY)
-	return big.Add(sector.ExpectedStoragePledge, big.Div(penalizedReward, big.NewInt(int64(EPOCHS_IN_DAY))))
+	return big.Add(*sector.ExpectedStoragePledge, big.Div(penalizedReward, big.NewInt(int64(EPOCHS_IN_DAY))))
```
rvagg, line 239: beware that these are going to start being `nil` from nv25 so you should be careful to guard for them

But also, maybe this should be replaced with a go-state-types variation of this since termination fees have been "simplified" and have a different calculation now. See https://github.com/filecoin-project/go-state-types/pull/376 for some exposed methods that help with this; https://github.com/filecoin-project/lotus/pull/12968 has use of them but also exposes them through the lotus go-state-types abstractions which might be more helpful. See the itest in there for how it can be used (a couple of variations).

